PR #7, opened 2024-03-07
Implement X-LoRA quantized model

EricLBuehler:


EricLBuehler:
Note that highly quantized models (eg., 4bit) do not perform well.

PR #10, opened 2024-03-10
Initial optimizations

EricLBuehler:


EricLBuehler:
On an H100, the model takes 30ms per completion token while llama.cpp takes 10ms per completion token. Thus the model speed is the primary obstacle.

EricLBuehler:
This will use my Candle backend which will be optimized.

hugoabonizio:
Hi @EricLBuehler! Great job with the optimization in your fork of Candle! Do you plan to submit these improvements upstream? I'm sure they would be very useful for the community!

EricLBuehler:
I am currently still working on those optimizations. Once they are done, I plan to submit a PR containing some of the major changes.

PR #12, opened 2024-03-13
Add instructions for cargo install

EricLBuehler:


README.md
```diff
@@ -115,10 +115,15 @@ cargo build --release --features cuda
...
-    cargo build --release --features cuda
-    ```
+5) Build or install
+    - Build
+        ```bash
+        cargo build --release --features cuda
+        ```
+    - Install with `cargo install` for easy command line usage
+        ```bash
+        cargo build --path mistralrs-server --features cuda
```
ivanbaldo, line 125: Shouldn't this be `install` instead of `build`?

EricLBuehler: Yes, you are correct, thanks for catching that! It should be fixed now.

PR #13, opened 2024-03-13
Optimization: Add fused kernels

EricLBuehler:
Added:
- [x] Fused LayerNorm
- [x] KV Concat

Adding:
- [x] Fused RoPE


EricLBuehler:
Note: will want to optimize start_offsets_kernel* by passing an &mut (not &, to maintain aliasing) instead of cloning.

PR #17, opened 2024-03-18
Add non granular scalings to optimize

EricLBuehler:


EricLBuehler:
Adds 1.8 T/s on an A6000.

PR #28, opened 2024-03-23
Fix Zephyr

EricLBuehler:


EricLBuehler:
It appears the issue stems from the RmsNorm implementation.

PR #34, opened 2024-03-26
Remove as many htod copies as possible

EricLBuehler:


EricLBuehler:
htos (1)

EricLBuehler:
htos (1)

EricLBuehler:
It looks like synchronization was the mysterious factor. After calling `CudaDevice::synchronize`:

synced

PR #36, opened 2024-03-27
Server: Allow `TokenSource` to be configured via cli

LLukas22:
Currently the `mistralrs-server` always uses the `TokenSource::CacheToken` TokenSource which is a bit unintuitive, especially if i want to use it in a docker container. 

This PR allows the `TokenSource` to be defined via the cli as follows: 

- `CacheToken` (Default): Same as current behaviour
- `Literal`: `--token-source literal:hf_xxxxxxxxxxx` uses the provided token
- `EnvVar`: `--token-source env:ENV` uses the provided environment variable, if no variable is provided it defaults to `HUGGING_FACE_HUB_TOKEN`  (similar behaviour as TGI/TEI)
-  `Path`: `--token-source path:.\token` uses the provided path

### DIFF
--- a/mistralrs-core/src/pipeline/mod.rs
+++ b/mistralrs-core/src/pipeline/mod.rs
@@ -3,6 +3,7 @@ mod llama;
 mod mistral;
 mod mixtral;
 use candle_sampling::logits_processor::Logprobs;
+use core::fmt;
 use either::Either;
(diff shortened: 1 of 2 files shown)


EricLBuehler (approved):
Looks good!

Issue #63, opened 2024-04-03
Streaming requests strip delta chunks


EricLBuehler:
Streaming requests return delta chunks which have no whitespace. This was first reported in #39:

```
Completion:
===============
The meaning of life is a philosophical and personal question that has been debated throughout history. Different people may have different answers to this question, depending on their beliefs, values, and experiences. Some people believe that the meaning of life is to seek happiness, while others believe that it is to fulfill a specific purpose or to achieve a certain goal. Ultimately, the meaning of life is a deeply personal and subjective concept that each individual must define for themselves.</s>

Streaming:
===============
Themeaningoflifeisaphilosophicalandpersonalquestionthathasbeendebatedthroughouthistory.Differentpeoplemayhavedifferentanswerstothisquestion,asitultimatelydependsonindividualbeliefs,values,andexperiences.Somemayfindmeaninginreligiousorspiritualbeliefs,whileothersmayfindmeaninginpursuingpersonalgoals,helpingothers,orcreatingmeaningfulrelationships.Ultimately,themeaningoflifeisadeeplypersonalandsubjectiveconceptthateachindividualmustexploreanddefineforthemselves.</s>
```

PR #64, opened 2024-04-03
Fix streaming strip

EricLBuehler:
Refs #63.


LLukas22:
Hey, i don't think this will fix the stripped chunks, as some tokenizers can contain multi-token words or other edge cases. Accoring to https://github.com/huggingface/tokenizers/issues/1141 we fixed this in LLM.rs by comparing the newly detokenized sequence to the last decoded sequenze (see [here](https://github.com/rustformers/llm/blob/main/crates/llm-base/src/inference_session.rs#L323-L326)). If you want i could implement this here.

EricLBuehler:
Thank you for offering! If you want to implement it that would be great. I will close this PR.

Issue #66, opened 2024-04-03
Optimization of sampling


EricLBuehler:
Sampling currently happens on CPU. By moving it to GPU it should be much faster.

This is a tracking issue for this implementation, which will be largely inspired by #34.

EricLBuehler:
It appears that the sorting is the major bottleneck. I have optimized the sampling code, and T/s is now 10% faster on an A10.

EricLBuehler:
Further optimizations in #90 have increased performance by 14%.

PR #67, opened 2024-04-03
Sample on GPU

EricLBuehler:
Refs #66.


EricLBuehler:
It appears that the sorting takes the longest amount of time.

PR #68, opened 2024-04-04
Fix Decoding during Streaming

LLukas22:
See #63.

Only yield chunks if the tokenizer decoded the sequence correctly, this should also handle multi token sequences. 

For now the `decoded_tokens` are keept in the `Sequence` to get the actual delta.


mistralrs-core/src/sequence.rs
```diff
@@ -7,6 +7,7 @@ use std::{
 
 use candle_core::Tensor;
 use candle_sampling::logits_processor::{LogitsProcessor, Logprobs};
```
EricLBuehler, line 9: I recently moved the sampler into `mistralrs-core`. Could you do a git merge to update?

EricLBuehler (changes requested):

EricLBuehler (approved):

LLukas22:
Forgot to sync my fork, should be good to go now 👍

EricLBuehler:
Thanks for adding this! Also, thanks for adding the `config.toml`!

PR #71, opened 2024-04-04
Use par iter in sampling

EricLBuehler:


EricLBuehler:
Causes a performance regression due to spawning 32000 threads.

PR #74, opened 2024-04-05
Add some QoL features to the Server

LLukas22:
Just adds some basic QoL features: 

- Configurable keep-alive-interval for sse via `KEEP_ALIVE_INTERVAL` environment variable
- Health check route on `/health` and `/` (mainly usefull for docker compose)
- OpenAPI doc and Swagger-UI via `utoipa` on `/docs` with example for `/v1/chat/completions`

### DIFF
--- a/mistralrs-server/Cargo.toml
+++ b/mistralrs-server/Cargo.toml
@@ -17,6 +17,8 @@ candle-core.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 axum = { version = "0.7.4", features = ["tokio"] }
+utoipa = { version = "4.2", features = ["axum_extras"] }
+utoipa-swagger-ui = { version = "6.0", features = ["axum"]}
 clap = { version = "4.5.1", features = ["derive"] }
 mistralrs-core = { version = "0.1.0", path = "../mistralrs-core" }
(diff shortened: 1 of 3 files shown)


EricLBuehler:
Thanks for the PR, this looks really great! I cannot seem to access the `/docs` endpoint with `curl http://localhost:1234/docs`, am I doing something wrong? It looks like `curl http://localhost:1234/api-docs/openapi.json` returns the expected docs.

EricLBuehler (approved):

LLukas22:
> Thanks for the PR, this looks really great! I cannot seem to access the `/docs` endpoint with `curl http://localhost:1234/docs`, am I doing something wrong? It looks like `curl http://localhost:1234/api-docs/openapi.json` returns the expected docs.

I don't think you are doing anything wrong. I also get no reply when I use `curl` on the `/docs` endpoint, but if I visit it in any browser, I get the SwaggerUI. I don't know if this is an error in my implementation or directly in `utoipa`, but all my [`text-embeddings-interface`](https://github.com/huggingface/text-embeddings-inference/tree/main) containers have exactly the same behavior where using `curl` on the `/docs` endpoint yields no result, but all the UIs show up if you actually visit them in a browser.

Here's how the UI looks on my side:
 
grafik

PR #85, opened 2024-04-06
Implement Phi2 model

EricLBuehler:
Introduces use of candle_nn::RotaryEmbedding::new_partial for the phi partial RoPE.


EricLBuehler:
This does not add a GGUF version.

PR #93, opened 2024-04-08
Implement merging for Lora layers

EricLBuehler:
In preparation for full LoRA support via merged layers to not degrade performance.


EricLBuehler:
QLoraLinear may be more difficult because we need to modify the weights.

PR #94, opened 2024-04-08
Allow input prompt message to be a String prompt

EricLBuehler:
Updates:
- `http.md` docs
- `mistralrs.pyi` type hints

Changes:
- `ChatCompletionReuqest::new` parsing/downcasting
- `ChatCompletionRequest` handles the string
- `Engine` handles not running the chat template if the prompt is already provided.

### DIFF
--- a/examples/http.md
+++ b/examples/http.md
@@ -79,7 +79,7 @@ curl http://localhost:<port>/docs
 OpenAI compatible request.
 ```rust
 pub struct ChatCompletionRequest {
-    pub messages: Vec<Message>,
+    pub messages: Either<Vec<Message>, String>,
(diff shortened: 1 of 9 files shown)


lucasavila00:
@EricLBuehler wouldn't it be better to support both Chat Completion API and Completion API?

This way it maintains OpenAI compatibility.

Check for instance vLLM that supports both https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/api_server.py#L86-L112

The difference between the 2 APIs is the chat template.

EricLBuehler:
Yes, that is probably the right way to do this. I'll add a PR for this later today.

Issue #91, opened 2024-04-08
Prompt Caching


lucasavila00:
# Use case: 

In a conversation, skip prefilling for the entire prompt every time a new message is sent.

Store the previously calculated KVs if there is enough VRAM for it.

# Examples

[Automatic Prefix Caching](https://github.com/vllm-project/vllm/issues/2614)

[Fast and Expressive LLM Inference with RadixAttention and SGLang](https://lmsys.org/blog/2024-01-17-sglang/)

PR #95, opened 2024-04-08
Implement prefix caching

EricLBuehler:
Refs #91.

This PR implements prefix caching, a concept introduced in [RadixAttention](https://lmsys.org/blog/2024-01-17-sglang/). It works by caching the KV cache resulting from running the model over the prompt (the "prefill step"), which is expensive. We implement a PrefixCacheManager which stores input prompt tokens with their prefilled KV caches and which avoids OOM by evicting caches.

**Caching**
- [x] The prefix cache manager holds all the KV caches from the prompts of previous sequences. Internally, this looks like a `HashMap<token ids, kv cache>`. The caching step is very simple and does not copy the KV cache, which is left on the device and only the KV cache `Tensor`s are cloned.

**Evicting**
- [x] When the scheduler runs, it will check if there is enough memory for the KV cache for the next token to be allocated. If there is not enough memory, the prefix manager will evict some KV caches from device memory to the CPU.

**Loading**
- [x] When a sequence is received, it will check with the prefix cache manager to see if it can skip the prefill step based on its prompt.
- [x] Further step: if a prompt exists in the prefix cache manager which contains more tokens _after_ the input prompt (that is, we can truncate the KV cache), we can still skip the prefill step.


EricLBuehler:
@lucasavila00, what do you think about this?

lucasavila00:
> @lucasavila00, what do you think about this? It is ready for merging, but I wanted to get your opinion.

This looks amazing, thank you!

I read the code and my little understanding saw no issue.

I'll clone and run it.

lucasavila00:
I added a few dbg! and I can see that in Chat a second message doesn't use the KV cache of the first.

Ideally this would be supported. When doing constrained generation of say Markdown, one needs to send the request with the `#` that marks the heading, then the model completes the heading, then re-send the prompt + heading for the model to generate a list and so on.

Besides that, even when the cache was hit, I could not tell a performance difference on my system :thinking:

EricLBuehler:
Yes, it doesn't cache the entire KV cache of the previous request, only the prompt. Given sampling, I don't think that would be a good idea, but please let me know.

I have fixed the lack of performance improvement, it is now much faster!

lucasavila00:
>Yes, it doesn't cache the entire KV cache of the previous request, only the prompt. Given sampling, I don't think that would be a good idea, but please let me know.

That would be good enough. In my example it'd re-calculate a handful of tokens every request. My understanding was that it wouldn't use any part of the cache if the prompt is not entirely cached.

> I have fixed the lack of performance improvement, it is now much faster!

Yes, it is for me too

EricLBuehler (commented):
Looks good.

PR #96, opened 2024-04-09
Remove sync calls and reorder cache ops

EricLBuehler:


EricLBuehler:
Performance improvement: 13%. **We are now at 93% of llama.cpp speed!**

Issue #97, opened 2024-04-09
Status code of HTTP error messages is not 4xx or 5xx


lucasavila00:
I can fix that. It's similar to https://github.com/EricLBuehler/mistral.rs/pull/80

Here: https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-server/src/main.rs#L140

`e.to_string().into_response()` does not carry the information that string is of an error, thus we get wrong HTTP status code

In the end this makes the OpenAI client return an unhelpful message:

```
Traceback (most recent call last):
  File "/home/lucas/oss/mistral.rs/examples/server/chat.py", line 25, in <module>
    resp = completion.choices[0].message.content
AttributeError: 'str' object has no attribute 'choices'
```

EricLBuehler:
Thanks for raising this! If you can submit a PR that would be great.

PR #98, opened 2024-04-09
Handle errors in forward pass, delete cache and cancel all sequences (Survive OOM)

lucasavila00:
I was playing around with the codebase, adding a few dbg! around and seeing what goes on under the hood.

While doing so I stumbled upon an Out of Memory error. I was using a very long prompt, and it won't fit my GPU indeed.

When this happened, the server no longer worked, the process was still alive and GPU memory had not been freed.

This behavior can be simulated with panicking anywhere in the forward function of the pipeline: https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/pipeline/mistral.rs#L371

I changed the Pipeline trait to return errors on the forward pass instead of panicking. I handled the errors by setting all sequence states to error, deleting the GPU cache and sending an error message to the receiver.

This makes the server continue to work and frees all GPU memory, allowing new requests to go on normally.

@EricLBuehler What are your thoughts?

If this approach is any good, I'll make the errors OpenAI API compatible too.

Also, I noticed the Status Code of the error message was 200, not 4xx or 5xx (but that's a different issue https://github.com/EricLBuehler/mistral.rs/issues/97)

### DIFF
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -68,7 +68,7 @@ impl Engine {
                     Self::clone_in_cache(&mut *pipeline, &mut scheduled.completion);
                 }
                 let logits = pipeline.forward(&scheduled.completion, false);
-                let logits = handle_pipeline_forward_error!(logits, &mut scheduled.completion, pipeline, 'lp);
+                let logits = handle_pipeline_forward_error!("completion", logits, &mut scheduled.completion, pipeline, 'lp);
 
                 if !self.no_kv_cache {
                     Self::clone_out_cache(&mut *pipeline, &mut scheduled.completion);
@@ -88,8 +88,7 @@ impl Engine {
                 // Run the prompt seqs
                 Self::set_none_cache(&mut *pipeline);
                 let logits = pipeline.forward(&scheduled.prompt, true);
-                let logits =
-                    handle_pipeline_forward_error!(logits, &mut scheduled.prompt, pipeline, 'lp);
+                let logits = handle_pipeline_forward_error!("prompt", logits, &mut scheduled.prompt, pipeline, 'lp);
 
                 if !self.no_kv_cache {
                     Self::clone_out_cache(&mut *pipeline, &mut scheduled.prompt);
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -68,7 +68,6 @@ impl Engine {
                     Self::clone_in_cache(&mut *pipeline, &mut scheduled.completion);
                 }
                 let logits = pipeline.forward(&scheduled.completion, false);
-
                 let logits = handle_pipeline_forward_error!(logits, &mut scheduled.completion, pipeline, 'lp);
 
                 if !self.no_kv_cache {
@@ -368,7 +367,7 @@ impl Engine {
                 // NOTE(EricLBuehler): Unwrap reasoning: The receiver should really be there, otherwise it is their fault.
                 request
                     .response
-                    .send(Response::Error(
+                    .send(Response::ValidationError(
                         format!("Prompt sequence length is greater than {}, perhaps consider using `truncate_sequence`?", get_mut_arcmutex!(self.pipeline).get_max_seq_len()).into(),
                     ))
                     .unwrap();
@@ -413,7 +412,7 @@ impl Engine {
                     // NOTE(EricLBuehler): Unwrap reasoning: The receiver should really be there, otherwise it is their fault.
                     request
                         .response
-                        .send(Response::Error(
+                        .send(Response::ValidationError(
                             format!("Stop sequence '{s:?}' encodes to multiple tokens when it should only encode to 1.").into(),
                         ))
                         .unwrap();
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -55,7 +55,7 @@ impl Engine {
     }
 
     pub fn run(&mut self) {
-        loop {
+        'lp: loop {
             if let Ok(request) = self.rx.try_recv() {
                 self.add_request(request);
             }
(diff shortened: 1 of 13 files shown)


mistralrs-core/src/engine/mod.rs
```diff
@@ -69,6 +69,22 @@ impl Engine {
...
+                                .send(Response::Error(e.to_string().into()))
+                                .unwrap();
+                            seq.set_state(SequenceState::Error);
+                        }
+                        Self::set_none_cache(&mut *pipeline);
+
+                        continue 'lp;
+                    }
+                };
+
```
lucasavila00, line 87: I don't like this code but I haven't given much thought to organization but rather behavior.

I think the waiting requests should also be set to errors? I'm not sure if `set_none_cache` alters them.

lucasavila00:
Another alternative is to exit the process here https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/pipeline/mistral.rs#L418

Tokio won't finish the process on panic https://users.rust-lang.org/t/panic-in-tokio-task-does-not-end-the-program-execution/45731

mistralrs-core/src/engine/mod.rs
EricLBuehler: No, the set_none_cache function just clears the model's cache. I don't think the waiting requests need to be set to Error, as they are not accessible in the loop by design.

I agree that this code is a bit repetitive, can you make this a macro in `utils`?

EricLBuehler:
Thanks for adding this, I like the idea of just raising an error instead of just `panic`ing. I think it may if it would be good to send back whatever was generated up until the error as a ChatCompletionResponse.

mistralrs-core/src/utils/mod.rs
```diff
@@ -36,14 +36,79 @@ macro_rules! handle_seq_error_stateaware {
...
+macro_rules! handle_pipeline_forward_error {
+    ($stage: tt, $fallible:expr, $seq_slice:expr, $pipeline:expr, $label:tt) => {
+        match $fallible {
+            Ok(v) => v,
+            Err(e) => {
+                use $crate::response::Response;
+                use $crate::sequence::SequenceState;
+                use $crate::Engine;
+                use $crate::response::SYSTEM_FINGERPRINT;
+                println!("{} - Model failed with error: {:?}", $stage, &e);
```
EricLBuehler, line 57: Could you change this to be a `error!` invocation? I think it is better to use `tracing`.

mistralrs-server/src/main.rs
```diff
@@ -109,7 +109,10 @@ impl futures::Stream for Streamer {
         }
         match self.rx.try_recv() {
             Ok(resp) => match resp {
-                Response::Error(e) => {
```
EricLBuehler, line 112: It looks like we ignore Response::ModelError, is there a reason why we do not handle it? Additionally, I think we should mark the streamer as done when an error comes through to prevent a hang, I wish we could signal that we got an error.

EricLBuehler (changes requested):

mistralrs-server/src/main.rs
lucasavila00: No reason at all. The `match` used a wildcard and I did not see that I should add another arm.

> I think we should mark the streamer as done when an error comes through to prevent a hang, I wish we could signal that we got an error.

This is my first time dealing with SSE, I also don't know what's the best way to deal with it...

mistralrs-server/src/main.rs
```diff
@@ -112,6 +120,13 @@ impl futures::Stream for Streamer {
                 Response::ValidationError(e) => {
                     Poll::Ready(Some(Ok(Event::default().data(e.to_string()))))
                 }
+                Response::ModelError(msg, _) => {
+                    MistralRs::maybe_log_error(
+                        self.state.clone(),
+                        &ModelErrorMessage(msg.to_string()),
+                    );
+                    Poll::Ready(Some(Ok(Event::default().data(msg))))
+                }
```
lucasavila00, line 129: `candle_core::Error` does not implement clone or copy, so I could not duplicate it to set as the response for every sequence...

So I decided to carry it around as a string. That's unfortunate here, where an ad-hoc error needs to be created

mistralrs-server/src/main.rs
EricLBuehler: I checked the Event docs, and there doesn't seem to be a way to signal an error. The only option is to set it in the comment, but I think that it would be hard to find there.

mistralrs-server/src/main.rs
EricLBuehler: I think it should be fine: we only need to optimize the happy path, as the hope is that there will be no errors!

mistralrs-server/src/main.rs
```diff
@@ -109,7 +117,17 @@ impl futures::Stream for Streamer {
         }
         match self.rx.try_recv() {
             Ok(resp) => match resp {
-                Response::Error(e) => {
+                Response::ValidationError(e) => {
+                    Poll::Ready(Some(Ok(Event::default().data(e.to_string()))))
```
EricLBuehler, line 121: It looks like there is a missing call to `maybe_log_error`, could you please add that?

mistralrs-server/src/main.rs
```diff
@@ -251,10 +298,15 @@ async fn chatcompletions(
...
 
         match response {
-            Response::Error(e) => {
+            Response::InternalError(e) => {
                 MistralRs::maybe_log_error(state, &*e);
-                ChatCompletionResponder::Error(e)
+                ChatCompletionResponder::InternalError(e)
+            }
+            Response::ModelError(msg, response) => {
+                MistralRs::maybe_log_response(state, &response);
```
EricLBuehler, line 306: I think this should be a call to `maybe_log_error`.

mistralrs-server/src/main.rs
```diff
@@ -251,10 +298,15 @@ async fn chatcompletions(
...
+                ChatCompletionResponder::ModelError(msg, response)
             }
+            Response::ValidationError(e) => ChatCompletionResponder::ValidationError(e),
```
EricLBuehler, line 309: Can you please add a call to `maybe_log_error`?

EricLBuehler (changes requested):

mistralrs-server/src/main.rs
lucasavila00: That was intentional, but sure let's add it.

In general, if the user made the mistake I don't log it.

EricLBuehler: Ah, that is a good idea. Lets not add it in that case.

mistralrs-server/src/main.rs
lucasavila00: Yeah, my bad

mistralrs-server/src/main.rs
EricLBuehler: > Ah, that is a good idea. Lets not add it in that case.

EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #103, opened 2024-04-10
Constrained decoding

lucasavila00:
The public interface is the same used by TGI.

A grammar parameter that can have type "regex" or "yacc". (TGI has regex and json-schema)

# Example 1 - [Token healing](https://towardsdatascience.com/the-art-of-prompt-design-prompt-boundaries-and-token-healing-3b2448b0be38)
```
curl http://localhost:1234/v1/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer EMPTY" \
-d '{
"model": "",
"grammar": {
    "type": "regex",
    "value": "Sure! (?s:.)*"
},
"messages": [
{
    "role": "user",
    "content": "Tell me a joke."
}
]
}'
```

# Example 2 - Markdown List

```
curl http://localhost:1234/v1/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer EMPTY" \
-d '{
"model": "",
"grammar": {
    "type": "regex",
    "value": "(- [^\n]*\n)+(- [^\n]*)(\n\n)?"
},
"messages": [
{
    "role": "user",
    "content": "Write a list of jokes. Return a markdown list where each item is a joke."
}
]
}'
```


lucasavila00:
@EricLBuehler What do you think about this approach?

The Windows build fails because the linker complains about missing functions that exist just for WASM. I'm not sure how to deal with this...

Also, I'm not a fan of using a library from Git. Should we vendor the code or something?

I haven't tried optimizing it yet, but on my system, this creates no overhead.

mistralrs-core/src/pipeline/llama.rs
```diff
@@ -444,6 +447,7 @@ impl Loader for LlamaLoader {
 
         Ok(Box::new(Mutex::new(LlamaPipeline {
             model,
+            tok_trie: build_tok_trie(tokenizer.clone()).into(),
```
lucasavila00, line 450: Each pipeline has the trie.

mistralrs-core/src/engine/mod.rs
```diff
@@ -457,6 +472,7 @@ impl Engine {
                 group.clone(),
                 response_index,
                 now.as_secs(),
+                Self::build_sequence_recognizer(&request.constraint),
```
lucasavila00, line 475: Each sequence has a stateful recognizer

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -301,9 +310,43 @@ pub trait Pipeline: Send + Sync {
...
+                self.tok_trie()
+                    .append_token(rx.as_mut(), second_logprobs_response.token);
+            }
+            SequenceRecognizer::Cfg(ref mut cfg) => {
+                self.tok_trie()
+                    .append_token(cfg.as_mut(), second_logprobs_response.token);
+            }
+            SequenceRecognizer::None => {}
+        }
+        Ok(second_logprobs_response)
```
lucasavila00, line 349: It samples, check if the samples token was accepted, fix it otherwise, advances the state machine, returns the original or fixed token

EricLBuehler:
This looks really great, thank you so much for adding it! However, I am concerned about the build failures on Windows, which I do want to support. An AICI issue: https://github.com/microsoft/aici/issues/42 says that they do not support native Windows, which is a problem.

Perhaps we could support using `aici_abi` on non-Windows platforms, but use kalosm_sample's [RegexParser](https://docs.rs/kalosm-sample/0.2.1/kalosm_sample/struct.RegexParser.html) otherwise or on all platforms for consistency? This is really great, and I wish that Windows support was available. Do you think you could modify it to use RegexParser? For example, they run the parser [here](https://github.com/floneum/floneum/blob/c494ec3a2445dc80b6efe01f1603eb94a627c92e/interfaces/language-model/src/structured.rs#L80-L86) and modify the logits.

lucasavila00:
> This looks really great, thank you so much for adding it! However, I am concerned about the build failures on Windows, which I do want to support. An AICI issue: https://github.com/microsoft/aici/issues/42 says that they do not support native Windows, which is a problem.

I tried vendoring the code. It's 2k lines and by vendoring I was able to remove ~20 transient dependencies too.

> Perhaps we could support using aici_abi on non-Windows platforms, but use kalosm_sample's RegexParser otherwise or on all platforms for consistency? This is really great, and I wish that Windows support was available. Do you think you could modify it to use RegexParser? For example, they run the parser here and modify the logits.

The parser is in "string-space" and we need to operate on "token-space". This would require a lot of glue code to do this conversion. I hope to make AICI code work for us, if it doesn't then I look into RegexParser

EricLBuehler:
Thank you for pointing that out. Looks like the Windows build is passing; I think we are probably close to a merge once we do some testing, and you mark it as ready!

lucasavila00:
@EricLBuehler I think this is ready for review and merge.

There is work left to do on optimization, and I haven't tested the `yacc` implementation yet.

But since the MR is already big, it might be a good point to merge.

mistralrs-core/src/utils/bintokens.rs
```diff
@@ -0,0 +1,181 @@
...
+                            char_map
+                                .get(&c)
+                                .copied()
+                                .ok_or_else(|| anyhow!("missing char: {}", c))
+                        })
+                        .collect();
+                    let bytes = match bytes {
+                        Ok(b) => b,
+                        Err(e) => {
+                            println!("error: {} for {:?}", e, tok_name);
```
EricLBuehler, line 148: If this is an error, can we propagate it out, or if it is not really an error, log it with `tracing::error!`

mistralrs-pyo3/src/lib.rs
```diff
@@ -122,6 +124,7 @@ impl Runner {
                 response: tx,
                 return_logprobs: request.logprobs,
                 is_streaming: request.stream,
+                constraint: Constraint::None,
```
lucasavila00, line 127: I never dealt with Py/Rust FFI. Python API has no support for constraints...

mistralrs-core/src/engine/mod.rs
```diff
@@ -347,6 +349,19 @@ impl Engine {
...
     }
 
+    fn build_sequence_recognizer(constraint: &Constraint) -> SequenceRecognizer {
+        match constraint {
+            Constraint::Regex(rx) => {
+                SequenceRecognizer::Regex(StackRecognizer::from(RecRx::from_rx(rx)).into())
+            }
+            Constraint::Yacc(cfg) => SequenceRecognizer::Cfg(
+                // TODO: handle error
+                CfgParser::from_yacc(cfg).unwrap().into(),
```
lucasavila00, line 359: need to handle this

EricLBuehler (changes requested):

mistralrs-pyo3/src/lib.rs
EricLBuehler: I will add this is a seperate PR.

mistralrs-core/src/engine/mod.rs
EricLBuehler: This should be easy to handle, right? Can you please return a Result from this function?

EricLBuehler (changes requested):

EricLBuehler:
I wrote some comments regarding error handling, other than that this looks excellent. Could you please add the example requests above into the `examples/` directory?

mistralrs-core/src/utils/bintokens.rs
lucasavila00: Sure. The AICI code had some of these, I removed a few but missed this.

The Yacc code has a bunch of these too.

EricLBuehler: Great, thanks!

mistralrs-core/src/engine/mod.rs
lucasavila00: fixed it

lucasavila00:
I added a Yacc example and tested that it works as expected.

EricLBuehler:
Thank you!

PR #108, opened 2024-04-11
Optimize qllama MQA impl

EricLBuehler:


EricLBuehler:
~3% performance improvement.

PR #109, opened 2024-04-11
Add multi-token stop sequence

### LINKED ISSUE #104: Multi-tokens stop sequence
I would like to use mistral.rs as a backend on https://github.com/lucasavila00/LmScript

LmScript hopes to hide the token concept from users so it needs support for multi-token stop sequences.

I can try implementing it, if it is a desired feature.

lucasavila00:
Closes #104


mistralrs-server/src/main.rs
```diff
@@ -230,8 +230,6 @@ fn parse_request(
     let stop_toks = match oairequest.stop_seqs {
         Some(StopTokens::Multi(m)) => Some(InternalStopTokens::Seqs(m)),
         Some(StopTokens::Single(s)) => Some(InternalStopTokens::Seqs(vec![s])),
-        Some(StopTokens::MultiId(m)) => Some(InternalStopTokens::Ids(m)),
```
lucasavila00, line 233: This felt like a footgun to me.
Since there is no discriminator, if one passes a number as stop sequence it might be interpreted as token.

There is no unambiguous way of passing a number as stop sequence, so I removed this from open ai api.

mistralrs-core/src/sequence.rs
```diff
@@ -244,7 +267,14 @@ impl Sequence {
         } else if self.tokens.len().saturating_sub(self.prompt_len) == max_model_len {
             Some(StopReason::ModelLength(max_model_len))
         } else {
-            None
+            let text_acc = self.generated_text(tokenizer);
+            if self.stop_strings.iter().any(|s| text_acc.contains(s)) {
+                // TODO: a different stop reason?
+                // TODO: strip the stop string from the generated text?
+                Some(StopReason::StopTok(tok))
```
lucasavila00, line 274: StopReason implements Copy, so I could not create a variant for the String. I need to think more about this, maybe it just shouldn't be copy.

mistralrs-core/src/sequence.rs
```diff
@@ -231,10 +234,30 @@ impl Sequence {
...
+        &self,
+        tok: u32,
+        eos_tok: u32,
+        max_model_len: usize,
+        tokenizer: &Tokenizer,
+    ) -> Option<StopReason> {
         if tok == eos_tok {
             Some(StopReason::Eos)
         } else if self.stop_tokens.contains(&tok) {
+            // TODO: strip the stop token from the generated text?
```
lucasavila00, line 260: In general engines remove the stop token from text... But whatever, the downstream consumer can do it too easily.

mistralrs-core/src/sequence.rs
```diff
@@ -244,7 +267,14 @@ impl Sequence {
         } else if self.tokens.len().saturating_sub(self.prompt_len) == max_model_len {
             Some(StopReason::ModelLength(max_model_len))
         } else {
-            None
+            let text_acc = self.generated_text(tokenizer);
+            if self.stop_strings.iter().any(|s| text_acc.contains(s)) {
+                // TODO: a different stop reason?
+                // TODO: strip the stop string from the generated text?
```
lucasavila00, line 273: In general engines remove the stop token from text... But whatever, the downstream consumer can do it too easily.

mistralrs-core/src/engine/mod.rs
```diff
@@ -452,28 +457,48 @@ impl Engine {
...
+                            stop_toks.push(toks[0]);
+                        }
+                    } else {
+                        stop_strings.push(stop_txt.clone());
+                    }
                 }
-                stop_toks.push(toks[0]);
-                stop_toks
+
+                (stop_toks, stop_strings)
```
lucasavila00, line 501: Suppose the stop sequence is a prefix like " " and we stop when generating " moon". There is no way to know if the tokens, when stringified, have a " " without checking the whole string.

So we'll use very inneficient string contains operations for strings.

So this still uses stop_toks as before, but only if the stop contains just a single token, and is not a prefix token. If both are not satisfied, we fall back to string checks.

mistralrs-core/src/engine/mod.rs
```diff
@@ -452,28 +457,48 @@ impl Engine {
...
+                for id in i {
+                    // We can't use ` ` (space) as a stop token because other tokens like ` moon` start with a space.
+                    if tok_trie.has_extensions(tok_trie.token(*id)) {
+                        request
+                            .response
+                            .send(Response::ValidationError(
+                                format!("Stop token {:?} is also a prefix of other tokens and cannot be used as a stop token.", tok_trie.token_str(*id)).into(),
+                            ))
+                            .unwrap();
+                        return;
```
lucasavila00, line 474: I re-used the validation for valid stop tokens here.

mistralrs-core/src/sequence.rs
```diff
@@ -231,10 +234,30 @@ impl Sequence {
...
-    pub fn is_done(&self, tok: u32, eos_tok: u32, max_model_len: usize) -> Option<StopReason> {
+    pub fn generated_text(&self, tokenizer: &Tokenizer) -> String {
+        let token_ids = self
+            .logprobs()
+            .iter()
+            .map(|lp| lp.token)
+            .collect::<Vec<_>>();
+        let text = tokenizer
+            .decode(&token_ids, false)
+            .expect("Failed to decode tokens generated by model");
```
lucasavila00, line 245: Can the tokenizer fail here? Like if the last token doesn't complete a valid UTF codepoint yet

mistralrs-core/src/sequence.rs
lucasavila00: @EricLBuehler what do you think? Removing Copy raised a lot of errors...

lucasavila00: I made it save the index of the string in the list of stop strings. Now it's copy too.

mistralrs-core/src/sequence.rs
lucasavila00: I'll create a different issue for this.

mistralrs-core/src/sequence.rs
lucasavila00: I'll create another issue for this.

mistralrs-core/src/sequence.rs
```diff
@@ -231,20 +235,47 @@ impl Sequence {
...
         } else {
-            None
+            if !self.stop_strings.is_empty() {
+                let text_acc = self.generated_text(tokenizer)?;
+                for (idx, s) in self.stop_strings.iter().enumerate() {
+                    if text_acc.contains(s) {
+                        return Ok(Some(StopReason::StopString(idx)));
+                    }
+                }
+            }
```
lucasavila00, line 277: I could not think of how to optimize this, besides not running if not requested

mistralrs-core/src/engine/mod.rs
EricLBuehler: Sounds good.

mistralrs-core/src/sequence.rs
EricLBuehler: Yes, I think this is a good solution.

mistralrs-server/src/main.rs
EricLBuehler: Thanks, you are absolutely correct!

EricLBuehler (approved):
@lucasavila00, thanks for the PR! I've pushed a commit to implement it for the PyO3 bindings too.

EricLBuehler:
@lucasavila00, I moved this to #111 as I needed to make some changes. Sorry for any confusion!

PR #112, opened 2024-04-12
Handle regex compilation error

lucasavila00:
That original `.unwrap` is from AICI. There might be more.


EricLBuehler:
@lucasavila00, I detected a severe performance drop upon merging of #109/#111. Master is now reverted to a state before that. Can you please re-open a new PR, which ideally maintains the T/s? Thank you!

EricLBuehler (changes requested):
Thank you for implementing this! Can you please re-open and base your changes on the updated master? #109 servery reduced performance, and it looks like this PR has those commits too. If you could open a new PR similar to #109 that does not incur a performance hit. Sorry for the trouble, it was my mistake.

Issue #99, opened 2024-04-09
Streaming in Tokens is significantly slower than not streaming them in


LLukas22:
If streaming is enabled the generation slows down significantly. 

The following script: 
```python 
Runs = 4

def request(stream:bool):
    client = openai.Client(api_key="foobar", base_url=ENDPOINT)
    return client.chat.completions.create(
            model="mistral",
            messages=[{"role": "user", "content": "What is the meaning of life?"}],
            stream=stream,
            max_tokens=200,
            temperature=0.0
        )
    
def run():
    for run in range(Runs):
        print("\nCompletion: ")
        print("="*15)
        
        now = datetime.now()
        answer = request(stream=False)
        finished = datetime.now()
        
        usage = answer.usage.completion_tokens      
        token_per_sek = usage / (finished-now).total_seconds()
        print(f"Duration: {finished-now}")
        print(f"Tokens per second: {token_per_sek}")
        
        print("\nStreaming: ")
        print("="*15)
        
        now = datetime.now()
        counter = 0
        stream = request(stream=True)
        for message in stream:
            counter += 1 
        finished = datetime.now()
        
        print(f"Duration: {finished-now}")
        token_per_sek = counter / (finished-now).total_seconds()
        print(f"Tokens per second: {token_per_sek}") 
        
        
if __name__ == "__main__":
    run()

```

Produces the following output, when run against a server hosted on a single L40: 
```
Completion:
===============
Duration: 0:00:01.335570
Tokens per second: 67.38695837732205

Streaming:
===============
Duration: 0:00:02.141581
Tokens per second: 42.02502730459413
```

These are the results from an ollama server running on the same hardware: 
```
Completion:
===============
Duration: 0:00:01.338007
Tokens per second: 84.45396772961577

Streaming:
===============
Duration: 0:00:01.380666
Tokens per second: 81.84455907511303
```

EricLBuehler:
I was getting 93% of llama.cpp T/s last night (72 vs 79) on an A10 for a completion request. When I was benchmarking the sampling code, the `finish_seq` function adds 7ms to the normal sampling time of 2ms (finish_seq is called in the sampling process). I think that this may be caused by the use of a Responder to send the data across threads, which takes significantly longer. I think the only other way to avoid having 2 threads (one for the server, one for the engine) is to rework the server code significantly. I would assume that the ollama/llama.cpp server does not rely on multithreading to do streaming.

[...]

PR #116, opened 2024-04-12
Support multiple stop tokens

lucasavila00:


lucasavila00:
@EricLBuehler how should I profile this change?

Should I run on CPU to avoid CUDA's async API? The only place I feel I can grasp the sampling duration is on the visual profiler :thinking:

EricLBuehler:
@lucasavila00, actually, I was thinking of something a lot simpler: comparing the completion T/s on master to this branch. Ideally the cost would be opt-in, and when it is used it would not be too costly.

It looks like your usage of `decoded_tokens` is actually incorrect: it stores the decoded tokens for streaming (the naming is not the best, we should probably change it). Perhaps you could cache the last generated token in `Sequence` (entrypt: `add_token`) or pass it to `is_done` (extract in `sample_seqs`)?

EricLBuehler:
I think this is still unfortunately incorrect, as getting the delta is only for streaming sequences. Can you try to get the `.bytes` field of `Logprobs` (returned by the sampler) and pass that to the `is_done` function instead?

lucasavila00:
> I think this is still unfortunately incorrect, as getting the delta is only for streaming sequences. Can you try to get the .bytes field of Logprobs (returned by the sampler) and pass that to the is_done function instead?

Sure, but won't that have just the last token's bytes?

The check must be done on the whole string, no? For instance, I can set the stop to "a word and spaces" and to know if we have generated this text we can't really optimize it, no?


I was able to reproduce the slowdown even when not being used, I'm trying to figure that out first.

EricLBuehler:
Hmm, that is a good point. Perhaps we could keep a String containing the entire decoded sequence, and then check against that? I think that would be faster than decoding all the tokens every pass. I also found the slowdown when not being used, that was very confusing.

lucasavila00:
@EricLBuehler it was caused by cloning the tokenizer. I moved the tokenizer to an Arc and it fixed the issue.

lucasavila00:
>Ah, great that you found that! Is there a performance regression when it checks against the stop seqs?

I'm starting on it.

BTW the ARC will probably fix the streaming issues #99 as it also cloned the tokenizer here. I assumed that it was cheap because of it.

lucasavila00:
> Is there a performance regression when it checks against the stop seqs?

When used, on https://github.com/EricLBuehler/mistral.rs/pull/116/commits/d3d15cb107e5445396b4b8f92d6acd37566f7f0f there is a performance regression.

I'm getting 73 for this PR and 77 for master (warm run, prefilled cache)

Let me try avoiding re-tokenizing everything all the time now.

lucasavila00:
I cannot figure out how to re-build the text from the tokens iteratively.

If I get the values directly from tokenizer using `id_to_token` special characters appear in the text

```
2024-04-12T03:03:41.495875Z  INFO mistralrs_server: Serving on http://0.0.0.0:1234.
[mistralrs-core/src/sequence.rs:269:17] &self.completion_text = "▁Once"
[mistralrs-core/src/sequence.rs:269:17] &self.completion_text = "▁Once▁upon"
[mistralrs-core/src/sequence.rs:269:17] &self.completion_text = "▁Once▁upon▁a"
```

or if I use `tokenizer.decode` then text has no spaces.
```
2024-04-12T02:50:17.690452Z  INFO mistralrs_server: Serving on http://0.0.0.0:1234.
[mistralrs-core/src/sequence.rs:269:17] &self.completion_text = "Once"
[mistralrs-core/src/sequence.rs:269:17] &self.completion_text = "Onceupon"
```

I organized the code and left it still re-tokenizing everything all the time.

lucasavila00:
I think the TokTrie might have utilities that help with this. Let me give it a shot.

lucasavila00:
After https://github.com/EricLBuehler/mistral.rs/pull/116/commits/404af701f0cef06d9850901db62eebc7c8e1117c

I measure 76t/s when enabled, 77t/s when disabled.

mistralrs-core/src/sequence.rs
```diff
@@ -243,21 +249,45 @@ impl Sequence {
...
+    fn is_sub<T: PartialEq>(mut haystack: &[T], needle: &[T]) -> bool {
+        if needle.is_empty() {
+            return true;
+        }
+        while !haystack.is_empty() {
+            if haystack.starts_with(needle) {
+                return true;
+            }
+            haystack = &haystack[1..];
+        }
```
lucasavila00, line 261: Maybe it's worth it to use a more optimized version of this is_sub?

eg: https://docs.rs/galil-seiferas/latest/galil_seiferas/ from https://stackoverflow.com/a/47045355

lucasavila00: I added it...

lucasavila00: I'm not sure if this is the best library...

EricLBuehler (approved):

EricLBuehler:
Thank you!

Issue #121, opened 2024-04-12
Crash on prefix cache


lucasavila00:
```
import openai

openai.api_key = "EMPTY"
openai.base_url = "http://localhost:1234/v1/"

completion = openai.chat.completions.create(
    model="mistral",
    messages=[
        {
            "role": "user",
            "content": "Tell me a joke.",
        },
        {
            "role": "assistant",
            "content": "Sure! Here's a joke: Why did the tomato turn red? Because it saw the salad dressing!",
        },
        {
            "role": "user",
            "content": "Tell me another joke.",
        },
    ],
    max_tokens=256,
    frequency_penalty=1.0,
    top_p=0.1,
    temperature=0,
)


completion = openai.chat.completions.create(
    model="mistral",
    messages=[
        {
            "role": "user",
            "content": "Tell me a joke.",
        }
    ],
    max_tokens=256,
    frequency_penalty=1.0,
    top_p=0.1,
    temperature=0,
)

```

```
thread '<unnamed>' panicked at mistralrs-core/src/prefix_cacher.rs:95:29:
range start index 49 out of range for slice of length 13
``` 

I'll try to fix this.

lucasavila00:
Once I fixed the crash I stumbled upon the issue I reported on https://github.com/EricLBuehler/mistral.rs/pull/122#discussion_r1563545992

For now I'm not looking into fixing this. I feel like I still don't understand enough of the code to handle it

EricLBuehler:
This should be fixed.

PR #122, opened 2024-04-12
Remove stop from response

### LINKED ISSUE #113: Remove stop token from response
Most engines do it.

It's very useful if one has multiple stop sequences and they don't want to loop over all possibilities to remove the matched option, since the engine can do it...

I think the PR doing it will be simple. I'll give it a shot and we can discuss if it is a good idea.

lucasavila00:
Closes #113

### DIFF
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -216,33 +216,28 @@ impl Engine {
         };
 
         let text = match reason {
-            StopReason::Length(_) | StopReason::ModelLength(_) => handle_seq_error!(
-                pipeline
-                    .tokenizer()
-                    .decode(&seq.get_toks()[seq.prompt_tokens()..], false),
-                seq.responder()
-            ),
+            StopReason::Length(_) | StopReason::ModelLength(_) => pipeline
+                .tokenizer()
+                .decode(&seq.get_toks()[seq.prompt_tokens()..], false),
+
             StopReason::Eos | StopReason::StopTok(_) => {
                 let toks = seq.get_toks();
-                handle_seq_error!(
-                    pipeline
-                        .tokenizer()
-                        .decode(&toks[seq.prompt_tokens()..toks.len() - 1], false),
-                    seq.responder()
-                )
+                pipeline
+                    .tokenizer()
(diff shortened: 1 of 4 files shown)


mistralrs-core/src/engine/mod.rs
```diff
@@ -215,19 +215,36 @@ impl Engine {
             None
         };
 
-        let res = handle_seq_error!(
-            pipeline
+        let text = match reason {
```
lucasavila00, line 218: The use of `get_toks` here was and is incorrect when a prefix cache item is being used (https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/sequence.rs#L204-L209)

I could not execute it because the prefix cache crashes https://github.com/EricLBuehler/mistral.rs/issues/121

I'm trying to understand the prefix cache code to propose a fix. Ideally the fix would add support for caching the other way around, where the cache has a subset of a prompt.

EricLBuehler (changes requested):
Thank you for implementing this. Can you please update the examples of the server (the python files) accordingly?

mistralrs-core/src/engine/mod.rs
EricLBuehler: Actually, that is intentional and only used if we used a cache which had a prompt that was a superset of our prompt. In that case `get_toks` returns the remaining tokens. I don't think this problem affects this PR?

lucasavila00: I think it does... I don't understand how `get_toks` would return the decoded tokens of the response... They contain the tokens of the prompt, no?

lucasavila00:
This doesn't affect streaming, so I did not change the streaming example.

String stop sequences will not work while streaming. We might want to skip sending finish tokens though?

mistralrs-core/src/engine/mod.rs
EricLBuehler: No, it returns the tokens which were not in the cached prompt, please see PrefixCacheManager::search_for_matching_cache, specifically the cases where it returns Subset. The toks which are in that variant are returned by `get_toks` for the prompt pass.

EricLBuehler (approved):

EricLBuehler:
I think this looks good. If you could open another PR to remove stop tokens from streaming requests that would be great!

PR #128, opened 2024-04-13
Optimize streaming

### LINKED ISSUE #99: Streaming in Tokens is significantly slower than not streaming them in

lucasavila00:
It seems Axum/OpenAI libraries add ~300ms of overhead on opening/closing the stream.

I can always get this 300ms diff no matter the prompt length.

Using 400 tokens like in the .py file I'm sending I get 96% efficiency, close to Ollama's 97%.

It doesn't look like there's much to improve here.

Closes #99

### DIFF
--- a/examples/server/stream_completion_bench.py
+++ b/examples/server/stream_completion_bench.py
@@ -1,0 +1,49 @@
+import openai
+from datetime import datetime
+
+Runs = 4
+
+ENDPOINT = "http://localhost:1234/v1/"
+
+
+def request(stream: bool):
+    client = openai.Client(api_key="foobar", base_url=ENDPOINT)
+    return client.chat.completions.create(
+        model="mistral",
+        messages=[
+            {
+                "role": "user",
+                "content": "What is the meaning of life? Write a long story.",
+            }
+        ],
+        stream=stream,
+        max_tokens=400,
+        temperature=0.0,
+    )
+
+
+def run():
+    for run in range(Runs):
+        print("\nCompletion: ")
+        print("=" * 15)
+
+        now = datetime.now()
+        answer = request(stream=False)
+        finished = datetime.now()
+
+        print(f"Duration: {finished-now}")
+
+        print("\nStreaming: ")
+        print("=" * 15)
+
+        now = datetime.now()
+        stream = request(stream=True)
+        for _message in stream:
+            pass
+        finished = datetime.now()
+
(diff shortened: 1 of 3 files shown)


lucasavila00:
@EricLBuehler I think you want to re-record the Readme video.

This is on an RTX 2070 :open_mouth: 

[Screencast from 13-04-2024 14:54:47.webm](https://github.com/EricLBuehler/mistral.rs/assets/12750442/ebd76d76-c3fb-4ca5-91ee-8c5a4cc448f1)

EricLBuehler:
Yes, that would be amazing!

EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #137, opened 2024-04-14
Fix interactive mode stripping end of message

lucasavila00:
Due to the stream rate limit there was a chance we'd not print the latest 2 tokens of the message.


EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #138, opened 2024-04-14
Print queue state

lucasavila00:


EricLBuehler:
@lucasavila00, there have been some changes since this PR has been opened, can you please run `git merge`? Additionally, can we gate the printing so that it is only enabled if there is a debug build?

lucasavila00:
I don't think we need to merge this. It's easy to add the dbg!s when required.

Issue #126, opened 2024-04-13
Prompt caching forces model to answer with first word of first response


lucasavila00:
This is weird. I've been debugging it for 10 minutes and I'm pretty sure the pattern holds.

```
2024-04-13T14:51:45.250836Z  INFO mistralrs_server::interactive_mode: Starting interactive loop with sampling params: SamplingParams { temperature: Some(0.1), top_k: Some(32), top_p: Some(0.1), top_n_logprobs: 0, frequency_penalty: Some(0.1), presence_penalty: Some(0.1), stop_toks: None, max_len: Some(4096), logits_bias: None, n_choices: 1 }
> Tell me a joke.
 Why don't scientists trust atoms?

Because they make up everything!
> Tell me another joke.
 Why:
Why did the tomato turn red?
Because it saw the salad dressing!

I hope that brought a smile to your face! If you have any other requests, feel free to ask.
> Tell me a third joke, that does not start with 'why" 
 Why is the ocean always at the beach? Because it's where the wavessss are! [I hope you found that joke amusing. If you have any other requests or questions, feel free to ask!]
```

If I edit `prefix_cacher.rs` and make it always find no cache, returning `Ok(None))` from `search_for_matching_cache`:

```
2024-04-13T14:50:46.587752Z  INFO mistralrs_server::interactive_mode: Starting interactive loop with sampling params: SamplingParams { temperature: Some(0.1), top_k: Some(32), top_p: Some(0.1), top_n_logprobs: 0, frequency_penalty: Some(0.1), presence_penalty: Some(0.1), stop_toks: None, max_len: Some(4096), logits_bias: None, n_choices: 1 }
> Tell me a joke.
 Why don't scientists trust atoms?

Because they make up everything!
> Tell another joke.
 Why did the scarecrow win an award?

Because he was outstanding in his field!
> A different one! One that does not start with why
 Sure thing! Here's a classic one:

Why did the chicken cross the playground?
```

I could also reproduce with other tokens. For instance:

```
$ python3 chat.py 
Enter system prompt >>> 
>>> Tell me a random number between 0 and 100.
I'd be happy to help, but I cannot provide a random number between 0 and 100 that is truly random as I don't have the ability to generate numbers randomly. However, if you need a specific number for demonstration purposes or for any other reason, feel free to ask and I can provide one. For example, let me give you the number 42. Is there anything particular you would like to do with this number?
>>> Ok, generate another.
[...]

PR #140, opened 2024-04-15
Rework the prefix cacher

EricLBuehler:
Refs #126.

This changes:
- Searching strategy:
  - [x] The previous "subset" method is the only method as it also covers the "verbatim" method
- Caching strategy:
  - [x] Cache the sequences after they are done (https://lmsys.org/blog/2024-01-17-sglang/) 
  - [x] Always maximize cache usage


EricLBuehler:
@lucasavila00, this seems to fix the problems in #126 in my testing. Does it also fix the problems for you?

PR #143, opened 2024-04-15
Fix get_prompt_input underflow & crash

lucasavila00:


lucasavila00:
@EricLBuehler 

I think the off by one here:

```
[mistralrs-core/src/pipeline/mod.rs:376:9] &max_len = 201
[mistralrs-core/src/pipeline/mod.rs:377:9] ctxt.len() = 202
```

Is a sign of a bigger issue, and we should not just saturate sub?

The issue is that `seq.len() != seq.get_toks().len()`

lucasavila00:
I think this latest revision is a good fix.

If we use the KV cache for len, we need to add 1 because the KV cache does not contain the current token

EricLBuehler (approved):

EricLBuehler:
Yes, this was a mistake. Thank you for catching it.

PR #146, opened 2024-04-16
Add more settings to prompt mode

lucasavila00:


mistralrs-server/src/main.rs
```diff
@@ -92,6 +92,14 @@ struct Args {
...
     #[arg(long, default_value_t = 16)]
     prefix_cache_n: usize,
+
+    /// Number of prompt completions to run concurrently in prompt mode.
+    #[clap(long, default_value_t = 1)]
+    prompt_concurrency: usize,
+
+    /// Number of prompt tokens to generate.
+    #[clap(long, default_value_t = 128)]
+    prompt_max_tokens: usize,
```
lucasavila00, line 102: Should these be a sub-command of sorts? I feel like they could confuse someone.

These are only used in prompt mode. I change these settings a lot when profiling

EricLBuehler: Yes, if you could make this a sub command that would be great!

lucasavila00: Clap doesn't allow 2 sub-commands, so I made the prompt arguments require prompt, and improved docs.

EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #147, opened 2024-04-16
Add env-var based scheduled status printing

lucasavila00:


mistralrs-core/src/engine/mod.rs
```diff
@@ -63,10 +64,14 @@ impl Engine {
             truncate_sequence,
             no_kv_cache,
             prefix_cacher: PrefixCacheManager::new(device, prefix_cache_n, is_xlora),
+            is_debug: std::env::var("RUST_LOG")
+                .unwrap_or_default()
+                .contains("debug"),
```
lucasavila00, line 69: This doesn't really match the [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html) behavior of tracing crate... But I don't know how to match it exactly...

This is just an optimization to avoid creating the vecs before calling tracing::debug!

EricLBuehler: I think it should be fine, as long as it is documented.

PR #151, opened 2024-04-16
Remove sampling metrics

lucasavila00:
Sampling is done on the GPU, so it's not possible to measure this timing without GPU synchronization.

It is misleading and I think it can be removed.

The only way to measure sampling is through nvidia profiler.


EricLBuehler (approved):

EricLBuehler:
I wonder how llama.cpp measures their sampling time? But I think this is a non-essential feature and like you said it is a bit misleading.

PR #152, opened 2024-04-16
Sampling optimizations

lucasavila00:
Makes sampling fully async, removing synchronizations.

Generation goes from 54t/s to 59t/s using `./target/profiling/mistralrs-server --prompt "Tell me 3 jokes." mistral-gguf`

Same generation speed as Llama.cpp :smile:

### DIFF
--- a/mistralrs-core/src/sampler.rs
+++ b/mistralrs-core/src/sampler.rs
@@ -2,7 +2,7 @@
 
 use std::{collections::HashMap, iter::zip, sync::Arc};
 
-use candle_core::{bail, Device, Error, Result, Tensor, D};
+use candle_core::{bail, Error, Result, Tensor, D};
 use rand::{
     distributions::{Distribution, WeightedIndex},
     SeedableRng,
@@ -43,7 +43,6 @@ pub struct Sampler {
     topk: i64,
     topp: f64,
     vocab_size: usize,
-    device: Device,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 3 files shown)


mistralrs-core/src/sampler.rs
```diff
@@ -208,11 +201,18 @@ impl Sampler {
...
 
+        if let Some(penalty) = penalty {
+            for (token_id, logit) in probs.iter_mut().enumerate() {
+                *logit += penalty[token_id];
+            }
+        }
+
         // Sort by descending probability.
-        argsort_indices.sort_by(|&i, &j| probs[j].partial_cmp(&probs[i]).unwrap());
+        argsort_indices.sort_unstable_by(|&i, &j| probs[j].partial_cmp(&probs[i]).unwrap());
```
lucasavila00, line 215: `This sort is unstable (i.e., may reorder equal elements), in-place (i.e., does not allocate), and O(n * log(n)) worst-case.`

The benchmarks here show the difference between stable and unstable. (I don't think adding the lib is required)

https://github.com/orlp/glidesort
image

This saves 500ms per token

mistralrs-core/src/sampler.rs
```diff
@@ -247,22 +247,22 @@ impl Sampler {
...
+        let mut logits = vec![0.0; vocab_size];
         for (token_id, logit) in logits.iter_mut().enumerate() {
             let count = context.iter().filter(|x| **x as usize == token_id).count();
             *logit = *logit
                 - count as f32 * frequency_penalty
                 - if count > 0 { 1. } else { 0. } * presence_penalty;
         }
-        let logits_len = logits.len();
-        Tensor::from_vec(logits, logits_len, device)
+        Ok(logits)
```
lucasavila00, line 263: Since this reads the context we can't pre-allocate the tensor.

Allocating the tensor here was triggering a synchronization that would cost 500ms

EricLBuehler:
Thank you!

PR #162, opened 2024-04-16
Optimize quantized masked fill

### LINKED ISSUE #161: Quantized Mistral: Optimize masked fill
We should copy the changes from this PR https://github.com/huggingface/candle/pull/1770/files

lucasavila00:
Closes #161 

This MR:

```
Prompt T/s = 70.70707
Completion T/s = 61.244022
```

Master:

```
Prompt T/s = 59.82906
Completion T/s = 61.538464
```

### DIFF
--- a/mistralrs-core/src/models/quantized_llama.rs
+++ b/mistralrs-core/src/models/quantized_llama.rs
@@ -122,15 +122,14 @@ struct LayerWeights {
     n_kv_head: usize,
     head_dim: usize,
     rotary: RotaryEmbedding,
+    neg_inf: Tensor,
 }
 
-fn masked_fill(on_false: &Tensor, mask: &Tensor, on_true: f32) -> Result<Tensor> {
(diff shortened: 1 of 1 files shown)


lucasavila00:
It reduces completion speed too much.

I wonder if that's because of bad measurements or real. I'm profiling it.

lucasavila00:
It was a measurement issue, fixed by https://github.com/EricLBuehler/mistral.rs/pull/163

EricLBuehler:
@lucasavila00, I just merged #163. Can you please update the benchmarks?

lucasavila00:
@EricLBuehler I just did.

I'm having a hard time measuring small changes, significantly.

I think we need something like https://github.com/ggerganov/llama.cpp/blob/master/examples/llama-bench/README.md instead of the --prompt setup to benchmark.

I ran it a bunch of times and completion speed is unchanged. Prompt speed improved.

EricLBuehler:
That is a great idea, I think that we should add that in light of future improvements. I'll merge this as I think the performance gains are very significant!

PR #166, opened 2024-04-17
mistralrs-bench

### LINKED ISSUE #164: Infra: Create mistralrs-bench
I think it should be mostly a copy of https://github.com/ggerganov/llama.cpp/blob/master/examples/llama-bench/README.md

Makes it easy to compare the 2 projects

Our current `--prompt` setup is good enough for profiling but not for benchmarking.

Refs https://github.com/EricLBuehler/mistral.rs/pull/162#issuecomment-2060055590

lucasavila00:
Closes #164 

llama.cpp

```
$ ./llama-bench -m /home/lucas/.cache/huggingface/hub/models--TheBloke--Mistral-7B-Instruct-v0.1-GGUF/snapshots/731a9fc8f06f5f5e2db8a0cf9d256197eb6e05d1/mistral-7b-instruct-v0.1.Q4_K_M.gguf 
| model                          |       size |     params | backend    | ngl | test       |              t/s |
| ------------------------------ | ---------: | ---------: | ---------- | --: | ---------- | ---------------: |
| llama 7B Q4_K - Medium         |   4.07 GiB |     7.24 B | CUDA       |  99 | pp 512     |  1682.64 ± 58.17 |
| llama 7B Q4_K - Medium         |   4.07 GiB |     7.24 B | CUDA       |  99 | tg 128     |     64.46 ± 1.25 |
```

mistral.rs
```
$ cargo run --release --features cuda
+----------------------------------------+---------+--------+----------+-----------+-------------+--------------+
| model                                  | backend | test   | t/s      | ms/t      | concurrency | throughput/s |
+----------------------------------------+---------+--------+----------+-----------+-------------+--------------+
| TheBloke/Mistral-7B-Instruct-v0.1-GGUF | CUDA    | tg 128 | 59.78263 | 16.727266 |           1 |     59.78263 |
+----------------------------------------+---------+--------+----------+-----------+-------------+--------------+
| TheBloke/Mistral-7B-Instruct-v0.1-GGUF | CUDA    | pp 512 | 654.6095 | 1.5276283 |           1 |     654.6095 |
+----------------------------------------+---------+--------+----------+-----------+-------------+--------------+
```


mistralrs-bench/src/main.rs
```diff
@@ -0,0 +1,320 @@
...
+    repetitions: usize,
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let tok_model_id = "mistralai/Mistral-7B-Instruct-v0.1";
+    let tokenizer_json = None;
+    let quantized_model_id = "TheBloke/Mistral-7B-Instruct-v0.1-GGUF";
+    let quantized_filename = "mistral-7b-instruct-v0.1.Q4_K_M.gguf";
+    let repeat_last_n = 64;
```
lucasavila00, line 215: Model was hardcoded, I don't know how to re-use the `model_selected.rs` file here.

IDK if it's best to create a new library, move it to core, make server also a lib...

mistralrs-core/src/lib.rs
```diff
@@ -50,15 +50,80 @@ pub struct MistralRs {
...
+            no_prefix_cache,
+            prefix_cache_n,
+            disable_eos_stop,
+        } = config;
+
+        let truncate_sequence = truncate_sequence.unwrap_or(false);
+        let no_kv_cache = no_kv_cache.unwrap_or(false);
+        let no_prefix_cache = no_prefix_cache.unwrap_or(false);
+        let prefix_cache_n = prefix_cache_n.unwrap_or(16);
+        let disable_eos_stop = disable_eos_stop.unwrap_or(false);
```
lucasavila00, line 125: allows adding a new setting without having to change every caller

mistralrs-bench/src/main.rs
EricLBuehler: I think it would be best here to re-use the `model_selected.rs` file. Unfortunately, the logic for the server is, as you pointed out, duplicated here. I think this is something we can avoid: perhaps by adding a method to the `MistralRs` struct which acts like a constructor and takes the `ModelSelected`?

EricLBuehler (changes requested):

lucasavila00:
@EricLBuehler I made model an argument of the benchmark, by moving that part to core.

I would like to merge this ASAP.

The PR is too big already.

There's a lot to improve, but I'd prefer to make the improvements in further smaller PRs

EricLBuehler (approved):

EricLBuehler:
Great, thanks for adding this,

Issue #139, opened 2024-04-14
Quantized Mistral: Batching is slower than non batches


lucasavila00:
I added some code that prints the queue state: https://github.com/EricLBuehler/mistral.rs/pull/138

I ran it on a single generation:

```
2024-04-14T17:34:50.601969Z  INFO mistralrs_core::engine: Prompt[] Completion[210] - 21ms
```

And on batches:

```
2024-04-14T17:34:49.138483Z  INFO mistralrs_core::engine: Prompt[] Completion[269, 154] - 278ms

...

2024-04-14T17:34:34.634905Z  INFO mistralrs_core::engine: Prompt[] Completion[269, 217, 102] - 354ms
```

I'm trying to figure out why...

lucasavila00:
It seems to be because of mask-fill

![image](https://github.com/EricLBuehler/mistral.rs/assets/12750442/f23425f7-5707-4c6c-bad5-8f85772e4449)

lucasavila00:
The masked fill can be optimized https://github.com/huggingface/candle/pull/1770/files

But it doesn't change much.

When I look into the nvidia profiler different kernels are used if batching:

Batch=2
![image](https://github.com/EricLBuehler/mistral.rs/assets/12750442/1e42ee80-f6d3-42e0-8abb-4e01d269b7bd)

Batch=1
![image](https://github.com/EricLBuehler/mistral.rs/assets/12750442/f9c626a7-52a2-4b48-954c-10859af218c9)

lucasavila00:
I guess candle has the optimized version just for batch=1 https://github.com/huggingface/candle/blob/main/candle-core/src/quantized/cuda.rs#L310-L321 indeed

LLukas22:
@lucasavila00 

I think Laurent added the batched kernels today: https://github.com/huggingface/candle/pull/2060

EricLBuehler:
Yes, I've been tracking that. I will merge the upstream changes to our branch.

lucasavila00:
It's 3x faster now!

But still slower than not parallelizing up to 5 requests.

20ms/token batch=1
133ms/token batch=2
144ms/token batch=3

lucasavila00:
Candle seems to be missing a PR from Llama.cpp which makes it use the `vec` kernels if <= 4 batches.

Refs https://github.com/huggingface/candle/issues/2074

[...]

PR #195, opened 2024-04-22
Sample seqs in parallel

EricLBuehler:
Refs #139.


mistralrs-core/src/sequence.rs
```diff
@@ -94,6 +94,11 @@ pub struct Sequence {
     group: Rc<RefCell<SequenceGroup>>,
     state: Cell<SequenceState>,
 }
+
+/// SAFETY: The scheduler only gives out mutable refs, and the `sample_seqs` function only modifies
```
EricLBuehler, line 98: I think this is actually unsound...

PR #196, opened 2024-04-22
Move kernels into repo

EricLBuehler:


EricLBuehler:
Doesn't work because we need from_storage, which is a private API.

PR #202, opened 2024-04-24
Add the Phi3 model

EricLBuehler:
Since most of the llama-architecture models have the same code, it probably makes sense to do a major refactor to handle that sometime in the near future. This is pending on #157.


EricLBuehler:
Adds the following:
- Normal Phi3
- LoRA Phi3
- X-LoRa Phi3
- Quantized Phi3 (already implemented due to the same "llama" architecture)

EricLBuehler:
128K support with LongRope will be coming soon, see: https://github.com/huggingface/transformers/blob/bc274a28a9a72c5fe98b070e2083fe689c8fd5e9/src/transformers/models/phi3/modeling_phi3.py#L131

PR #198, opened 2024-04-23
Async sampling

lucasavila00:
```
./target/profiling/mistralrs-bench -p 0 -g 64 -r 1 -c 8  gguf -t mistralai/Mistral-7B-Instruct-v0.1 -m TheBloke/Mistral-7B-Instruct-v0.1-GGUF -f mistral-7b-instruct-v0.1.Q4_K_M.gguf
```

Master

image

This PR

image

PR #210, opened 2024-04-26
Async engine

lucasavila00:
Continuation of https://github.com/EricLBuehler/mistral.rs/pull/198

If I understand it correctly, blocking in the server's async handler is not correct https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-server/src/chat_completion.rs#L245

This will block a tokio thread from doing other work, thus negating the benefits of the async axum server, and basically using a threadpool where each thread gets a request.


lucasavila00:
Not worth it to continue with this PR, it has too many conflicts.

I can look into this again after we've settled about #198

PR #227, opened 2024-04-27
docs: update ADAPTER_MODELS.md

eltociear:
minor fix


EricLBuehler:
@eltociear, thanks for fixing this! I think it should be "We provide an", could you please add that?

EricLBuehler (approved):
Thanks!

PR #231, opened 2024-04-27
Fix debug log timing of first token

lucasavila00:
If running in interactive mode it was calculating the time it takes to type the message as time of the first token


EricLBuehler (approved):

EricLBuehler:
Thank you

Issue #216, opened 2024-04-26
Batched & chunked prefill


lucasavila00:
Similar to what was described here https://github.com/huggingface/candle/issues/2108

"When prompts get longer than trivial sizes, the memory usage spikes as the prompt is thrown into one Tensor and sent off to a forward pass in the model at whatever length it comes in as. These spikes can be reduced by processing the batch in chunks."

There's a candle implementation here https://github.com/huggingface/candle/pull/2111

Let's say we configure a setting batch_size = 512.

The scheduler would need to be aware of it and only schedule 2 prompts if they're less than 512 tokens combined.

And the engine should be aware of it and if a sequence is larger than 512 tokens, split it.

To reproduce it locally, use the benchmark with a high enough `-p` and you get an OOM

``` 
./mistralrs-bench -p 2048 -g 0 -r 1 -c 1  gguf -t mistralai/Mistral-7B-Instruct-v0.1 -m TheBloke/
Mistral-7B-Instruct-v0.1-GGUF -f mistral-7b-instruct-v0.1.Q4_K_M.gguf

2024-04-26T20:17:25.483829Z ERROR mistralrs_core::engine: prompt - Model failed with error: Cuda(Cuda(DriverError(CUDA_ERROR_OUT_OF_MEMORY, "out of memory")))
```

But generating this same amount of tokens work

``` 
./mistralrs-bench -p 0 -g 2048 -r 1 -c 1  gguf -t mistralai/Mistral-7B-Instruct-v0.1 -m TheBloke/
Mistral-7B-Instruct-v0.1-GGUF -f mistral-7b-instruct-v0.1.Q4_K_M.gguf

+------------------------------------+---------+---------+--------------+--------------+-------------+--------------+
| model                              | backend | test    | t/s          | ms/t         | concurrency | throughput/s |
+------------------------------------+---------+---------+--------------+--------------+-------------+--------------+
| mistralai/Mistral-7B-Instruct-v0.1 | CUDA    | tg 2048 | 26.297±0.000 | 38.027±0.000 |           1 |    26.296867 |
+------------------------------------+---------+---------+--------------+--------------+-------------+--------------+
```

EricLBuehler:
@lucasavila00, this looks great. It'll require modifying the attention mask calculation of every model, so it may be helpful to factor those out into a `layers.rs` in `mistralrs-core`.

EricLBuehler:
@lucasavila00, I am actually going to end up adding this in #242.

[...]

PR #234, opened 2024-04-28
Batched & chunked prefill

lucasavila00:
Continuing https://github.com/EricLBuehler/mistral.rs/pull/219

Closes https://github.com/EricLBuehler/mistral.rs/issues/216

### DIFF
--- a/mistralrs-core/src/pipeline/mod.rs
+++ b/mistralrs-core/src/pipeline/mod.rs
@@ -363,11 +363,48 @@ pub trait NormalModel {
     }
 }
 
-struct InputMetadata {
-    input: Tensor,
-    positions: Vec<usize>,
-    positions_kernel: Tensor, // [bs, seq len]
-    context_lens: Vec<usize>,
+#[derive(Debug)]
+pub struct InputMetadata {
+    pub input: Tensor,
+    pub positions: Vec<usize>,
+    pub positions_kernel: Tensor, // [bs, seq len]
+    pub context_lens: Vec<usize>,
+}
+
+pub fn get_prompt_input_batched(
+    input_seqs: &[&mut Sequence],
+    device: &Device,
+    chunk_size: usize,
+) -> Result<Vec<InputMetadata>> {
+    assert!(input_seqs.len() == 1);
+    let seq = &input_seqs[0];
+
+    let mut acc = vec![];
+
+    for (chunk_index, chunk) in seq.get_toks().chunks(chunk_size).enumerate() {
+        let input = Tensor::new(chunk, &device)?.unsqueeze(0)?;
+        let pos_idx = chunk_index * chunk_size;
+        let positions = vec![pos_idx];
+
+        let this_chunk_size = chunk.len();
+        let positions_kernel = Tensor::from_slice(
+            &(pos_idx..this_chunk_size + pos_idx)
+                .map(|x| x as i64)
+                .collect::<Vec<_>>(),
+            this_chunk_size,
+            device,
+        )?;
+
+        let context_lens = vec![this_chunk_size - 1];
+        acc.push(InputMetadata {
+            input,
+            positions,
+            positions_kernel,
+            context_lens,
+        });
+    }
+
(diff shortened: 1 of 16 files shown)


mistralrs-core/src/xlora_models/mistral.rs
```diff
@@ -525,26 +525,20 @@ impl XLoraModel {
...
+        let t = tgt_len;
+        let u = seqlen_offset + tgt_len;
+        let mask: Vec<_> = (0..t)
             .flat_map(|i| {
-                (0..tgt_len).map(move |j| {
-                    if i < j || j + sliding_window < i {
-                        f32::NEG_INFINITY
-                    } else {
+                (0..u).map(move |j| {
+                    if j + t + self.sliding_window.unwrap_or(tgt_len + 1) > i + u {
```
lucasavila00, line 533: I'm not sure the sliding window part is right

mistralrs-core/src/xlora_models/mixtral.rs
```diff
@@ -675,25 +675,20 @@ impl XLoraModel {
...
+        let t = tgt_len;
+        let u = seqlen_offset + tgt_len;
+        let mask: Vec<_> = (0..t)
             .flat_map(|i| {
-                (0..tgt_len).map(move |j| {
-                    if i < j || j + self.sliding_window < i {
-                        f32::NEG_INFINITY
-                    } else {
+                (0..u).map(move |j| {
+                    if j + t + self.sliding_window > i + u {
```
lucasavila00, line 683: I'm not sure the sliding window part is right

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -363,11 +366,45 @@ pub trait NormalModel {
...
+    pub positions: Vec<usize>,
+    pub positions_kernel: Tensor, // [bs, seq len]
+    pub context_lens: Vec<usize>,
+}
+
+fn calculate_inputs_prompt_batched(
+    seq: &mut Sequence,
+    device: &Device,
+    chunk_size: usize,
+) -> Result<Vec<InputMetadata>> {
```
lucasavila00, line 381: Is it worth it making `InputMetadata` public? Or should this function return `ModelInputs`?

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -219,6 +219,9 @@ pub trait Pipeline: Send + Sync {
     fn get_max_seq_len(&self) -> usize;
     fn is_xlora(&self) -> bool;
     fn has_no_kv_cache(&self) -> bool;
+    fn get_prefill_chunk_size(&self) -> usize {
+        512
+    }
```
lucasavila00, line 224: Should this be configurable?

This value should be as large as possible for performance, under the constraint that it doesn't OOM the system.

mistralrs-core/src/xlora_models/phi3.rs
```diff
@@ -460,16 +460,12 @@ impl Model {
         tgt_len: usize,
         seqlen_offset: usize,
```
lucasavila00, line 461: This function already passed the `seqlen_offset`. I think it was already being correctly calculated for the current use-case, but I'm not sure

mistralrs-core/src/xlora_models/gemma.rs
```diff
@@ -561,16 +561,12 @@ impl XLoraModel {
         tgt_len: usize,
         seqlen_offset: usize,
```
lucasavila00, line 562: This function already passed the `seqlen_offset`. I think it was already being correctly calculated for the current use-case, but I'm not sure

lucasavila00:
@EricLBuehler I'd appreciate help on testing this.

It changed a lot of models, some of them I have never used.

I did test the quantized llama I usually run a lot though. Both on `mistral-bench` and talking to it with the interactive mode.

EricLBuehler:
@lucasavila00, absolutely, I'll test the models out.

PR #236, opened 2024-04-28
Async channels

lucasavila00:
Closes https://github.com/EricLBuehler/mistral.rs/issues/235

PR #237, opened 2024-04-28
Fix quantized example

lucasavila00:
I found it while testing https://github.com/EricLBuehler/mistral.rs/pull/236

### DIFF
--- a/mistralrs-core/src/sampler.rs
+++ b/mistralrs-core/src/sampler.rs
@@ -20,7 +20,7 @@ pub enum StopTokens {
     Ids(Vec<u32>),
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 /// Sampling params are used to control sampling.
 pub struct SamplingParams {
     pub temperature: Option<f64>,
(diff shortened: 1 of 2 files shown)


mistralrs-core/src/sampler.rs
```diff
@@ -35,6 +35,23 @@ pub struct SamplingParams {
...
+            temperature: None,
+            top_k: None,
+            top_p: None,
+            top_n_logprobs: 0,
+            frequency_penalty: None,
+            presence_penalty: None,
+            stop_toks: None,
+            max_len: None,
+            logits_bias: None,
+            n_choices: 1,
```
lucasavila00, line 50: it was defaulting to 0

lucasavila00: I'm not sure if this should use the Completion Request `best_of` instead. These 2 sources of truth are confusing

EricLBuehler: @lucasavila00 the `best_of` parameter sets the number of seqs to consider when they are done by cumulative logprob. It should be <= `n_choices`.

EricLBuehler (approved):

EricLBuehler:
Thank you

Issue #232, opened 2024-04-28
Accelerate topk, topp sampling with `argsort`


EricLBuehler:
Argsort was just added to Candle (https://github.com/huggingface/candle/pull/2132). Using an argsort kernel will accelerate the current CPU sorting part of `topk` or `topp` sampling, which takes a lot of time.

PR #239, opened 2024-04-28
Sampling: Sort indices with argsort kernel

EricLBuehler:
Refs #232.


EricLBuehler:
Causes large performance regression due to dual-synchronization.

PR #244, opened 2024-04-29
Sliding window for phi3

EricLBuehler:
2 major lessons:
- Need to slice the kv cache for sliding window
- RoPE freqs cannot be cast into low precision quickly, only after the cos/sin application: https://github.com/huggingface/transformers/pull/29285


EricLBuehler:
It looks like this issue is because of losing precision on long context windows: https://github.com/huggingface/transformers/pull/29285

PR #255, opened 2024-04-30
Add automatic pypi upload and docker build on release

EricLBuehler:


EricLBuehler:
@LLukas22, does this look correct to you?

LLukas22:
TEI defined [their triggers](https://github.com/huggingface/text-embeddings-inference/blob/main/.github/workflows/build_89.yaml#L5C1-L9C14) as follows: 
```
 on:
   workflow_dispatch:
   push:
     branches:
       - 'main'
     tags:
       - 'v*'
``` 

We only need the `tags` part of it. The `release` trigger you are currently using could also work, but i don't know if it contains the tag/version informations the workflow needs to tag the images 🤔

EricLBuehler:
Ok, I updated it:
```
on:
  workflow_dispatch:
  release:
    types: [published]
    tags:
       - 'v*'
```

Would that work?

LLukas22:
Probably, easiest way to find out is to just try it. 

Github actions can be very finicky sometimes 😅

EricLBuehler:
Ok, we will find out tomorrow with the rolling release 👍.

PR #274, opened 2024-05-08
Add build.rs to PyO3 to improve compat when extension_module

EricLBuehler:


EricLBuehler:
This should fix, once and for all, any errors like #35.

PR #284, opened 2024-05-10
fix typos

maximus2600:
mistralrs_server should be mistralrs-server

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ To install mistral.rs, one should ensure they have Rust installed by following [
         ```
 6) The build process will output a binary `misralrs-server` at `./target/release/mistralrs-server` which may be copied into the working directory with the following command:
     ```
-    cp ./target/release/mistralrs-server ./mistralrs_server
+    cp ./target/release/mistralrs-server ./mistralrs-server
     ```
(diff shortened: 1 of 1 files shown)


EricLBuehler:
I think that we do this intentionally, to avoid a conflict? When we copy the binary out, the folder `mistralrs-server` conflicts with the filename `mistralrs-server` so it does not put it in the correct location. By using `mistralrs_server` it just copies it. Do you have a better name idea?

maximus2600:
Ah I missed the step `cp ./target/release/mistralrs-server ./mistralrs_server`, IMHO `-` and `_` and got me confused.
Maybe the directories could be just named `bench`, `core`, `lora`, `pyo3` and `server`? Maybe `mistralrs-*` is a bit redundant since they all reside in the directoy mistral-rs.

Just my 2 cents.

EricLBuehler:
The convention of `project-subfolder` is a Rust convention unfortunately. Maybe we could copy it out to `server`: cp ./target/release/mistralrs-server ./server`?

EricLBuehler:
@maximus2600, were you planning on working on this PR?

PR #300, opened 2024-05-13
bug fix: llama kv cache part

keisuke-niimi-insightedge-jp:
I'm wondering if the part of using kv cache is wrong because of wrong indexing.


EricLBuehler:
Do you experience strange behavior?

EricLBuehler (approved):

EricLBuehler:
@keisuke-niimi-insightedge-jp Thank you!

PR #302, opened 2024-05-14
Update containers to cuda 12.4 + Fix missing libraries

LLukas22:


EricLBuehler (approved):

EricLBuehler:
This may be a solution to huggingface/candle#2175.

polarathene:
FWIW, I vaguely recall having an issue with 12.4 when my host system had only Cuda 12.2 installed. Thus this may reduce the UX a little.

When I upgraded my host (Windows 11) to Cuda 12.4, it required not only bringing down all actively running containers (managed via Docker Desktop) but also all WSL2 terminal sessions I had open too to shutdown WSL. Only then could I use Cuda again in containers / WSL2 without segfaults.

**UPDATE:** Knowing better now, this was probably because I ran into the situation of needing to use compat drivers when my system had the CUDA 12.2 driver, but upgrading to CUDA 12.4 there also works. The official nvidia images that are being used provide compat libs at `/usr/local/cuda/compat` for the version of CUDA that runtime image is bundled with.

PR #303, opened 2024-05-14
Make `pyo3` an optional dependency in `mistralrs-core`

LLukas22:
Puts `pyo3` behind the `pyo3_macros` feature flag, which is set per default in the `mistralrs-pyo3` create. 

This allows us to compile the benchmark/server without any python depencies via: 

```
cargo build --release --workspace --exclude mistralrs-pyo3
```

### DIFF
--- a/mistralrs-pyo3/Cargo.toml
+++ b/mistralrs-pyo3/Cargo.toml
@@ -17,7 +17,7 @@ doc = false
 
 [dependencies]
 pyo3.workspace = true
-mistralrs-core = { version = "0.1.7", path = "../mistralrs-core" }
+mistralrs-core = { version = "0.1.7", path = "../mistralrs-core", features = ["pyo3_macros"] }
 serde.workspace = true
 serde_json.workspace = true
 candle-core.workspace = true
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM rust:latest as builder
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-    python3-dev \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /mistralrs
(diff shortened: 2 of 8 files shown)


mistralrs-pyo3/Cargo.toml
```diff
@@ -17,7 +17,7 @@ doc = false
 
 [dependencies]
 pyo3.workspace = true
-mistralrs-core = { version = "0.1.7", path = "../mistralrs-core" }
+mistralrs-core = { version = "0.1.7", path = "../mistralrs-core", features = ["pyo3_macros"] }
```
EricLBuehler, line 20: Can you please also update `mistralrs-pyo3/Cargo_template.toml`?

EricLBuehler (changes requested):

mistralrs-pyo3/Cargo.toml
LLukas22: Yeah forgot to add it there, is the `Cargo_template.toml` only used to select features for the `pypi` uploads?

EricLBuehler: Yeah, it's used for creating packages with different features. I don't love that way of doing it, though, because we have so many packages uploaded. Do you know of another way to do this? Ideally, we could gate features on install, but maturin is missing functionality to do that.

LLukas22: Well it seams to be possible to set feature flags via the `MATURIN_PEP517_ARGS` environment, as mentioned [here](https://github.com/PyO3/maturin/discussions/1090). But that is very unintuitive. 

Maybe we should create a ticket to enable features via pips optional dependency flags e.g. mistralrs[cuda] to use the `cuda` feature.

LLukas22: For now i just added `pyo3_macros` as a forced feature. Should be good to go 👍

EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #308, opened 2024-05-14
Support loading model into pipeline from local filesystem

Jeadie:
## Motivation
- Support loading models from local filesystem to make it easier for productionising using `mistral.rs`. 

## Changes
- Change trait `Loader`. From `load_model`
    - `load_model_from_hf`
    - `load_model_from_path`
- Both of these can use the same logic once HF files have been downloaded locally.


mistralrs-server/src/main.rs
```diff
@@ -282,7 +282,7 @@ async fn main() -> Result<()> {
         warn!("Using flash attention with a quantized model has no effect!")
     }
     info!("Model kind is: {}", loader.get_kind().to_string());
-    let pipeline = loader.load_model(
+    let pipeline = loader.load_model_from_hf(
```
EricLBuehler, line 285: If I understand correctly, all the logic within `get_paths!` is still there, so if a user specifies a local path, and they use this API, nothing will change?

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -82,6 +83,37 @@ pub struct SimpleModelPaths<P> {
     lora_preload_adapter_info: Option<HashMap<String, (P, LoraConfig)>>,
 }
 
+impl<P> SimpleModelPaths<P> {
```
EricLBuehler, line 86: This API was internal before, so not really well documented. Can you please add some docs to explain what each path should point to?

mistralrs-core/src/pipeline/speculative.rs
```diff
@@ -30,27 +31,28 @@ pub struct SpeculativeLoader {
 }
 
 impl Loader for SpeculativeLoader {
-    fn load_model(
+    #[allow(clippy::type_complexity, clippy::too_many_arguments)]
+    fn load_model_from_hf(
```
EricLBuehler, line 35: Like mentioned below, can we document that this falls back to searching locally if it cannot find the files on HF hub?

EricLBuehler (commented):
Thank you for adding this! I think it's a great addition to the Rust API and just clears up the code a bit. I left a few minor comments regarding docs and one small question.

mistralrs-server/src/main.rs
Jeadie: Correct. For example, take a look at https://github.com/EricLBuehler/mistral.rs/pull/308/files#diff-79ee2143b95e1682589015af457db1e09fe5d24b70501b22750104eb761084c3R350-R371

mistralrs-core/src/pipeline/mod.rs
Jeadie: Done

mistralrs-core/src/pipeline/speculative.rs
Jeadie: Added to `Loader` trait.

mistralrs-core/src/pipeline/mod.rs
EricLBuehler: That looks good. Thanks for adding it!

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -82,6 +109,37 @@ pub struct SimpleModelPaths<P> {
...
```
EricLBuehler, line 112: Can we rename this to `LocalModelPaths` to emphasize its purpose?

Jeadie: Done.

EricLBuehler:
@Jeadie, I think there are some formatting/linting issues. After those are fixed, I think this should be ready to merge.

Jeadie:
Fixed and ran both `cargo fmt --all --check` and `cargo clippy --workspace --tests --examples -- -D warnings` locally.

EricLBuehler:
Thank you!

polarathene:
@Jeadie Is the local filesystem support only a few days old? Can you explain if HF API is mandatory?

Active issue: https://github.com/EricLBuehler/mistral.rs/issues/326

- It's much simpler to run a local GGUF model via `llama-cpp`.
- The expected parameters (redundant?) can also be a tad confusing. [This GGUF](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B-GGUF) presumably needs the additional files sourced from [here](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B/tree/main), but there is no `tokenizer.json`? From what I've read all the relevant metadata should already be available in the GGUF file itself, and `llama-cpp` happily runs with only the GGUF file provided.
- Without a HF token provided (`--token-source none`), the HF API is still hit but encounters an "Unauthorized" response (401), while only 404 responses have an exception which triggers a panic.

Perhaps there is some benefit with local filesystem going through HF API? Otherwise it'd probably be better to avoid that, or at least offer a way to opt-out if possible.

PR #328, opened 2024-05-18
chore: Simplify `utils/token.rs:get_token()`

polarathene:
Presently when a token is not considered valid via this utility method, it is treated as an empty string `Some("")` to hand over to the `hf-hub` crate API, but it should have instead been provided as `None`?

From the `hf-hub` crate, you can see that this would [avoid adding a blank authorization HTTP header](https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L143-L157)

```rs
    /// Sets the token to be used in the API
    pub fn with_token(mut self, token: Option<String>) -> Self {
        self.token = token;
        self
    }

    fn build_headers(&self) -> HeaderMap {
        let mut headers = HeaderMap::new();
        let user_agent = format!("unkown/None; {NAME}/{VERSION}; rust/unknown");
        headers.insert(USER_AGENT, user_agent);
        if let Some(token) = &self.token {
            headers.insert(AUTHORIZATION, format!("Bearer {token}"));
        }
        headers
    }
```

---

This PR change won't avoid this info log showing multiple times however:

```
INFO hf_hub: Token file not found "/root/.cache/huggingface/token"
```

Which is due to calling [`ApiBuilder::new()`](https://github.com/EricLBuehler/mistral.rs/blob/ca9bf7d1a8a67bd69a3eed89841a106d2e518c45/mistralrs-core/src/pipeline/macros.rs#L133-L138) which [initializes by checking a cache dir path for a `token` file](https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L105-L118) (_via [`cache.token()`](https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/lib.rs#L50-L68)_)

<details>
<summary>Added context</summary>

```rs
    pub fn new() -> Self {
        let cache = Cache::default();
        Self::from_cache(cache)
    }

// ...

    pub fn from_cache(cache: Cache) -> Self {
        let token = cache.token();
```

Since `mistral.rs` is already setting the token via `with_token()` the upstream internal `cache.token()` seems redundant? But would presently require recreating `from_cache()` without the `cache.token()` line and providing a [default `Cache` input](https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/lib.rs#L194-L208):

```rs
// hf-hub should really have a separate default method similar to:
pub fn new_api_builder() -> ApiBuilder {
    Self {
        endpoint: "https://huggingface.co".to_string(),
        url_template: "{endpoint}/{repo_id}/resolve/{revision}/{filename}".to_string(),
        cache: Cache::default(),
        token: None,
        progress: true,
    }
}

// ...

let api = new_api_builder()
  .with_progress(true)
  .with_token(get_token(token_source)?)
  .build()?;
```

</details>

**UPDATE:** I've sent a [PR upstream to address that](https://github.com/huggingface/hf-hub/pull/60), however it doesn't look like the crate is being maintained? Simple to review contributions already exist but have had no response.

### DIFF
--- a/mistralrs-core/src/utils/tokens.rs
+++ b/mistralrs-core/src/utils/tokens.rs
@@ -22,16 +22,8 @@ pub(crate) fn get_token(source: &TokenSource) -> Result<Option<String>> {
 
     let token = match source {
         TokenSource::Literal(data) => Some(data.clone()),
-        TokenSource::EnvVar(envvar) => {
-            env::var(envvar)
-                .ok()
-                .or_else(|| skip_token(envvar))
-        },
-        TokenSource::Path(path) => {
-            fs::read_to_string(path)
-                .ok()
-                .or_else(|| skip_token(path))
-        },
+        TokenSource::EnvVar(envvar) => env::var(envvar).ok().or_else(|| skip_token(envvar)),
+        TokenSource::Path(path) => fs::read_to_string(path).ok().or_else(|| skip_token(path)),
         TokenSource::CacheToken => {
             let home = format!(
                 "{}/.cache/huggingface/token",
@@ -43,7 +35,7 @@ pub(crate) fn get_token(source: &TokenSource) -> Result<Option<String>> {
             fs::read_to_string(home.clone())
                 .ok()
                 .or_else(|| skip_token(&home))
-        },
+        }
         TokenSource::None => None,
     };
 
--- a/mistralrs-core/src/utils/tokens.rs
+++ b/mistralrs-core/src/utils/tokens.rs
@@ -14,27 +14,24 @@ enum TokenRetrievalError {
 
 /// This reads a token from a specified source. If the token cannot be read, a warning is logged with `tracing`
 /// and *no token is used*.
-pub(crate) fn get_token(source: &TokenSource) -> Result<String> {
-    Ok(match source {
-        TokenSource::Literal(data) => data.clone(),
+pub(crate) fn get_token(source: &TokenSource) -> Result<Option<String>> {
+    fn skip_token(input: &str) -> Option<String> {
+        info!("Could not load token at {input:?}, using no HF token.");
+        None
+    }
+
+    let token = match source {
+        TokenSource::Literal(data) => Some(data.clone()),
         TokenSource::EnvVar(envvar) => {
-            let tok = env::var(envvar);
-            if let Ok(tok) = tok {
-                tok
-            } else {
-                info!("Could not load token at {envvar:?}, using no HF token.");
-                "".to_string()
-            }
-        }
+            env::var(envvar)
+                .ok()
+                .or_else(|| skip_token(envvar))
+        },
         TokenSource::Path(path) => {
-            let tok = fs::read_to_string(path);
-            if let Ok(tok) = tok {
-                tok
-            } else {
-                info!("Could not load token at {path:?}, using no HF token.");
-                "".to_string()
-            }
-        }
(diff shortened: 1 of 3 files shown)


mistralrs-core/src/utils/tokens.rs
```diff
@@ -14,44 +14,30 @@ enum TokenRetrievalError {
...
-                "".to_string()
-            }
-        }
+pub(crate) fn get_token(source: &TokenSource) -> Result<Option<String>> {
+    fn skip_token(input: &str) -> Option<String> {
+        info!("Could not load token at {input:?}, using no HF token.");
+        None
+    }
+
+    let token = match source {
```
EricLBuehler, line 23: What about, if `EnvVar` or `Path` fail, we try `Cache`. So, the graph would look like this:
- Literal -> `return Some` <--- Same as current
- EnvVar | Path (maybe `return Some`) or -> notify, try CacheToken (maybe `return Some`) -> notify, None
- CacheToken (maybe `return Some`) -> notify, None  <--- Same as current

EricLBuehler (commented):
This looks really great. When reviewing it, though, I had the idea to implement a fallback system similar to our ISQ fallback system. What do you think?

mistralrs-core/src/utils/tokens.rs
polarathene: I had a go at it, but the branches for `EnvVar` / `Path` calling `try_cache()` won't propagate the error if encountered, or at least not with the awkward looking approach I put together to share the cache logic as fallback:

```rs
pub(crate) fn get_token(source: &TokenSource) -> Result<Option<String>> {
    fn skip_token() -> Option<String> {
        info!("Proceeding without an HF token.");
        None
    }

    fn log_token_missing(location: &str) {
        info!("Could not find a token at '{location:?}'.");
    }

    fn try_cache() -> Result<Option<String>> {
        let home = format!(
            "{}/.cache/huggingface/token",
            dirs::home_dir()
                .ok_or(TokenRetrievalError::HomeDirectoryMissing)?
                .display()
        );

        Ok(fs::read_to_string(home.clone()).ok().or_else(|| {
            log_token_missing(&home);
            skip_token()
        }))
    }

    let token = match source {
        TokenSource::Literal(data) => Some(data.clone()),
        TokenSource::EnvVar(envvar) => env::var(envvar)
            .ok()
            .ok_or_else(|| {
                log_token_missing(envvar);
                try_cache()
            })
            .ok(),
        TokenSource::Path(path) => fs::read_to_string(path)
            .ok()
            .ok_or_else(|| {
                log_token_missing(path);
                try_cache()
            })
            .ok(),
        TokenSource::CacheToken => try_cache()?,
        TokenSource::None => None,
    };

    Ok(token.map(|s| s.trim().to_string()))
}
```

> What do you think?

I am not sure about the implicit fallback, I guess it's ok but if there is no actual user demand for it I would probably just keep it simple as-is to maintain?

EricLBuehler: Yeah, implicit behavior is generally bad. I think it's fine the way it is. Thanks!

EricLBuehler (approved):

EricLBuehler:
Thank you!

Issue #326, opened 2024-05-17
Running model from a GGUF file, only


MoonRide303:
**Describe the bug**
Running model from a GGUF file using [llama.cpp](https://github.com/ggerganov/llama.cpp) is very straightforward, just like that:
`server -v -ngl 99 -m Phi-3-mini-4k-instruct-Q6_K.gguf`
and if model is supported, it just works.

I tried to do the same using mistral.rs, and I've got that:
```
mistralrs-server gguf -m . -t . -f .\Phi-3-mini-4k-instruct-Q6_K.gguf
2024-05-17T07:21:48.581660Z  INFO mistralrs_server: avx: true, neon: false, simd128: false, f16c: true
2024-05-17T07:21:48.581743Z  INFO mistralrs_server: Sampling method: penalties -> temperature -> topk -> topp -> multinomial
2024-05-17T07:21:48.581820Z  INFO mistralrs_server: Loading model `.` on Cuda(CudaDevice(DeviceId(1)))...
2024-05-17T07:21:48.581873Z  INFO mistralrs_server: Model kind is: quantized from gguf (no adapters)
2024-05-17T07:21:48.583625Z  INFO hf_hub: Token file not found "C:\\Users\\[REDACTED]\\.cache\\huggingface\\token"
2024-05-17T07:21:48.583707Z  INFO mistralrs_core::utils::tokens: Could not load token at "C:\\Users\\[REDACTED]/.cache/huggingface/token", using no HF token.
thread 'main' panicked at mistralrs-core\src\pipeline\gguf.rs:290:58:
File "tokenizer.json" not found at model id "."
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Why it asks me for a tokenizer file, when it's [included in the GGUF file](https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#tokenizer)? I understand having this as an option (if I wanted to try out different tokenizer / configuration), but by default it should just use information provided in the gguf file itself.

Next attempt, when I copied tokenizer.json from the original model repo:
```
mistralrs-server gguf -m . -t . -f .\Phi-3-mini-4k-instruct-Q6_K.gguf
2024-05-17T07:28:34.987235Z  INFO mistralrs_server: avx: true, neon: false, simd128: false, f16c: true
2024-05-17T07:28:34.987332Z  INFO mistralrs_server: Sampling method: penalties -> temperature -> topk -> topp -> multinomial
[...]

PR #331, opened 2024-05-19
chore: `SimpleModelPaths` should be renamed to `LocalModelPaths`

polarathene:
This was originally carried out in https://github.com/EricLBuehler/mistral.rs/commit/07ea5b3039ab7a92a4b9e38967bc5c1e1b8df026 as part of https://github.com/EricLBuehler/mistral.rs/pull/308

When I was troubleshooting https://github.com/EricLBuehler/mistral.rs/issues/326#issuecomment-2118647910 I came across this comment but wasn't quite sure what it was referencing. Later I spotted the mentioned commit when looking at recent history and realized this line was accidentally missed 😅


EricLBuehler:
@polarathene, thank you!

PR #333, opened 2024-05-19
Remove candle-layer-norm dep

EricLBuehler:


EricLBuehler:
This removes the long build time of candle layer norm.

PR #334, opened 2024-05-19
chore: Use `strum` to simplify `GGUFArchitecture` maintenance

polarathene:
Minor improvement to consider?

- Minimizes some noise when assigning `arch` from metadata (_that is read via the `candle` crate_).
- Custom `FromStr` can be replaced by `strum` derive, making it DRY, adding a new variant no longer needs to update the string mapping.
- Both concerns now handled in smaller `from_value()` method.

### DIFF
--- a/mistralrs-core/src/pipeline/gguf.rs
+++ b/mistralrs-core/src/pipeline/gguf.rs
@@ -22,7 +22,7 @@ use crate::{
     utils::tokens::get_token,
     xlora_models::{XLoraQLlama, XLoraQPhi3},
 };
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use candle_core::quantized::{
     gguf_file::{self, Value as GgufValue},
     GgmlDType,
@@ -316,6 +316,7 @@ impl Loader for GGUFLoader {
             .map_err(|e| e.with_path(paths.get_weight_filenames().first().unwrap()))?;
         let arch = model.metadata["general.architecture"]
             .to_string()
+            .context("Model metadata should have declared an architecture")
             .and_then(GGUFArchitecture::from_value)?;
 
         info!("Model config:");
--- a/mistralrs-core/src/pipeline/gguf.rs
+++ b/mistralrs-core/src/pipeline/gguf.rs
@@ -36,6 +36,7 @@ use std::fs;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
+use strum::EnumString;
 use tokenizers::Tokenizer;
 use tokio::sync::Mutex;
 use tracing::info;
@@ -73,7 +74,8 @@ pub struct GGUFLoader {
     tgt_non_granular_index: Option<usize>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, EnumString)]
+#[strum(serialize_all = "kebab-case")]
 enum GGUFArchitecture {
     Llama,
     Mpt,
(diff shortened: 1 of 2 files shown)


mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -88,24 +90,14 @@ enum GGUFArchitecture {
...
-            "phi3" => Ok(GGUFArchitecture::Phi3),
-            a => Err(format!("Unknown GGUF architecture `{a}`")),
-        }
+// Wraps from_str() for some convenience:
+// - Case-insensitive variant matching (TODO: is this desirable?)
+// - Customized error until potential upstream support: https://github.com/Peternator7/strum/issues/332
+impl GGUFArchitecture {
+    fn from_value<T: AsRef<str> + std::fmt::Display>(value: T) -> Result<Self> {
+        Self::from_str(&value.as_ref().to_ascii_lowercase())
+            .with_context(|| format!("Unknown GGUF architecture `{value}`"))
```
EricLBuehler, line 99: Looks like there are some Clippy failures here.

mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -322,11 +314,9 @@ impl Loader for GGUFLoader {
         let mut file = std::fs::File::open(paths.get_weight_filenames().first().unwrap())?;
         let model = gguf_file::Content::read(&mut file)
             .map_err(|e| e.with_path(paths.get_weight_filenames().first().unwrap()))?;
-        let arch: GGUFArchitecture = model.metadata["general.architecture"]
+        let arch = model.metadata["general.architecture"]
             .to_string()
-            .unwrap()
-            .parse()
-            .map_err(anyhow::Error::msg)?;
+            .and_then(GGUFArchitecture::from_value)?;
```
EricLBuehler, line 319: Looks like there are some Clippy failures here.

EricLBuehler (commented):
Thanks for adding this, it makes the code much more idiomatic now. I think there are some Clippy lint failures, though. After those are fixed, I will be happy to merge.

polarathene:
I assume you have to approve the workflow run? I have no errors with clippy or rustfmt on my end. If it complains about `strum` this was resolved by `cargo add strum --features derive` for me, despite `Cargo.toml` already having that added 🤷‍♂️

polarathene:
All green, can we merge this? I'll then continue with https://github.com/EricLBuehler/mistral.rs/pull/335

EricLBuehler:
> BTW, I noticed you're using merge commits? Do you prefer that I clean up the commit history?

Absolutely! If you could do that it would be great.

EricLBuehler (approved):
Looks good!

polarathene:
> If you could do that it would be great.

Just updated the history into a single commit but was too slow 😂 

---

Thanks! I'll get the other PR sorted now 👍

EricLBuehler:
> Just updated the history into a single commit but was too slow 😂

Perhaps you could open a separate PR for this?

---
Looking forward to the other PR!

polarathene:
> Perhaps you could open a separate PR for this?

Sorry?

- This PR was 2 commits, the 2nd commit was a correction to the original. It could have been squashed which is what I did.
- You already merged the PR. It would require rewriting the main branch history now, not something I tend to encourage.

---

You may want to consider squash merge for future PRs though. You just select it via the drop-down instead of your current default merge strategy:

image

EricLBuehler:
Yes, I do plan on doing that in the future, though.

PR #335, opened 2024-05-19
refactor: `ModelKind` with `strum` + `derive_more`

polarathene:
This is something to consider, but would be a bit awkward without replacing `ModelKind` as a breaking change.

It was something I explored after https://github.com/EricLBuehler/mistral.rs/pull/334 and depends on it for `strum`, likewise I've not added the derive-more crate here either. Going forward with such a change for this type would be rather large, but can be done if deemed worthwhile.

I noticed quite a bit of logic to support the different kinds is mostly repeated with very little if any differences. Meanwhile some logic looks like it's already falling out of sync, thus reducing the noise throughout the codebase may help focus on minimizing that concern?

---

I'm not sure if `Lora` quantized kinds were intentionally excluded here, could simply be a call to `kind.is_quantized()`?:

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-bench/src/main.rs#L316-L326

There seems to be some logic to detect a difference between `lora` and `xlora` types? (_the `ModelKind` could better indicate that or similar?_):

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/gguf.rs#L406-L407

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/gguf.rs#L473-L476

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/gguf.rs#L498-L507

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/normal.rs#L297

---

Many of these would be collapsed to the compound type proposed by the PR:

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/normal.rs#L231-L234

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/normal.rs#L277-L284

Whereas these would need additional refactoring, but all have the same shape for parameters that they could pass in a struct?

https://github.com/EricLBuehler/mistral.rs/blob/1d5f9f3a4cb5a0e4bd408cabd1f2b8fb549b4c57/mistralrs-core/src/pipeline/normal.rs#L235-L276

---

The builder methods also seem rather suitable for deferring to a crate to derive, and that would work better with a breaking change away from these large parameterized methods funneling data through them.

I won't have time to tackle such any time soon, but something to consider as more models and features are added before this sprawls further 😅

### DIFF
--- a/mistralrs-bench/src/main.rs
+++ b/mistralrs-bench/src/main.rs
@@ -2,7 +2,7 @@ use candle_core::Device;
 use clap::Parser;
 use cli_table::{format::Justify, print_stdout, Cell, CellStruct, Style, Table};
 use mistralrs_core::{
-    Constraint, DeviceMapMetadata, Loader, LoaderBuilder, MistralRs, MistralRsBuilder, ModelKind,
+    Constraint, DeviceMapMetadata, Loader, LoaderBuilder, MistralRs, MistralRsBuilder,
     ModelSelected, NormalRequest, Request, RequestMessage, Response, SamplingParams,
     SchedulerMethod, TokenSource, Usage,
 };
@@ -313,15 +313,7 @@ fn main() -> anyhow::Result<()> {
     if use_flash_attn {
         info!("Using flash attention.");
     }
-    if use_flash_attn
-        && matches!(
-            loader.get_kind(),
-            ModelKind::QuantizedGGML
-                | ModelKind::QuantizedGGUF
-                | ModelKind::XLoraGGML
-                | ModelKind::XLoraGGUF
-        )
-    {
+    if use_flash_attn && loader.get_kind().is_quantized() {
         warn!("Using flash attention with a quantized model has no effect!")
     }
     info!("Model kind is: {}", loader.get_kind().to_string());
--- a/mistralrs-core/src/pipeline/mod.rs
+++ b/mistralrs-core/src/pipeline/mod.rs
@@ -244,21 +244,153 @@ pub enum ModelKind {
     },
 }
 
+// TODO: Future replacement for `ModelKind` above:
+#[derive(Default, derive_more::From, strum::Display)]
+pub enum ModelKindB {
+    #[default]
+    #[strum(to_string = "normal (no quant, no adapters)")]
+    Plain,
+
+    #[strum(to_string = "quantized from {quant} (no adapters)")]
+    Quantized { quant: QuantizationKind },
+
+    #[strum(to_string = "{adapter}, (no quant)")]
+    Adapter { adapter: AdapterKind },
+
+    #[strum(to_string = "{adapter}, quantized from {quant}")]
+    AdapterQuantized {
+        adapter: AdapterKind,
+        quant: QuantizationKind,
+    },
+
+    // TODO: This would need to be later changed to reference `Self`, but this current way
+    // avoids having to handle the conversion logic with `ModelKind`.
+    #[strum(to_string = "speculative: target: `{target}`, draft: `{draft}`")]
+    Speculative {
+        target: Box<ModelKind>,
+        draft: Box<ModelKind>,
+    },
+}
+
+#[derive(Clone, Copy, strum::Display, strum::EnumIs)]
+#[strum(serialize_all = "kebab-case")]
+pub enum QuantizationKind {
+    Ggml,
+    Gguf,
+}
+
+#[derive(Clone, Copy, strum::Display, strum::EnumIs)]
+#[strum(serialize_all = "kebab-case")]
+pub enum AdapterKind {
+    Lora,
+    XLora,
+}
+
+impl ModelKindB {
+    // Quantized helpers:
+    pub fn is_quantized(&self) -> bool {
+        self.quantized_kind().iter().any(|q| q.is_some())
+    }
+
+    pub fn is_quantized_and(&self, mut f: impl FnMut(QuantizationKind) -> bool) -> bool {
+        self.quantized_kind().iter().any(|q| q.is_some_and(&mut f))
+    }
+
+    pub fn quantized_kind(&self) -> Vec<Option<QuantizationKind>> {
+        use ModelKindB::*;
+
+        match self {
+            Plain | Adapter { .. } => vec![None],
+            Quantized { quant } | AdapterQuantized { quant, .. } => vec![Some(*quant)],
+            Speculative { target, draft } => {
+                let t = ModelKindB::from(*target.clone());
+                let d = ModelKindB::from(*draft.clone());
+
+                [t.quantized_kind(), d.quantized_kind()].concat()
+            }
+        }
+    }
+
+    // Adapter helpers:
+    pub fn is_adapted(&self) -> bool {
+        self.adapted_kind().iter().any(|a| a.is_some())
+    }
+
+    pub fn is_adapted_and(&self, mut f: impl FnMut(AdapterKind) -> bool) -> bool {
+        self.adapted_kind().iter().any(|a| a.is_some_and(&mut f))
+    }
+
+    pub fn adapted_kind(&self) -> Vec<Option<AdapterKind>> {
+        use ModelKindB::*;
+
+        match self {
+            Plain | Quantized { .. } => vec![None],
+            Adapter { adapter } | AdapterQuantized { adapter, .. } => vec![Some(*adapter)],
+            Speculative { target, draft } => {
+                let t = ModelKindB::from(*target.clone());
+                let d = ModelKindB::from(*draft.clone());
+
+                [t.adapted_kind(), d.adapted_kind()].concat()
+            }
+        }
+    }
+}
+
+// TODO: Temporary compatibility layers follow (until a future PR follow-up introduces a breaking change)
 impl Display for ModelKind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            ModelKind::Normal => write!(f, "normal (no quant, no adapters)"),
-            ModelKind::QuantizedGGML => write!(f, "quantized from ggml (no adapters)"),
-            ModelKind::QuantizedGGUF => write!(f, "quantized from gguf (no adapters)"),
-            ModelKind::XLoraNormal => write!(f, "x-lora (no quant)"),
-            ModelKind::XLoraGGML => write!(f, "x-lora, quantized from ggml"),
-            ModelKind::XLoraGGUF => write!(f, "x-lora, quantized from gguf"),
-            ModelKind::LoraGGUF => write!(f, "lora, quantized from gguf"),
-            ModelKind::LoraGGML => write!(f, "lora, quantized from ggml"),
-            ModelKind::LoraNormal => write!(f, "lora (no quant)"),
-            ModelKind::Speculative { target, draft } => {
-                write!(f, "speculative: target: `{target}`, draft: `{draft}`")
-            }
+        write!(f, "{}", ModelKindB::from(self.clone()))
+    }
+}
+
+// Delegate to `ModelKindB` methods:
+impl ModelKind {
+    // Quantized helpers:
+    pub fn is_quantized(&self) -> bool {
+        let k = ModelKindB::from(self.clone());
+        k.is_quantized()
+    }
+
+    pub fn is_quantized_and(&self, f: impl FnMut(QuantizationKind) -> bool) -> bool {
+        let k = ModelKindB::from(self.clone());
+        k.is_quantized_and(f)
+    }
+
+    pub fn quantized_kind(&self) -> Vec<Option<QuantizationKind>> {
+        let k = ModelKindB::from(self.clone());
+        k.quantized_kind()
+    }
+
+    // Adapter helpers:
+    pub fn is_adapted(&self) -> bool {
+        let k = ModelKindB::from(self.clone());
+        k.is_adapted()
+    }
+
+    pub fn is_adapted_and(&self, f: impl FnMut(AdapterKind) -> bool) -> bool {
+        let k = ModelKindB::from(self.clone());
+        k.is_adapted_and(f)
+    }
+
+    pub fn adapted_kind(&self) -> Vec<Option<AdapterKind>> {
+        let k = ModelKindB::from(self.clone());
+        k.adapted_kind()
+    }
+}
+
+impl From<ModelKind> for ModelKindB {
+    fn from(kind: ModelKind) -> Self {
+        match kind {
+            ModelKind::Normal => ModelKindB::Plain,
+            ModelKind::QuantizedGGML => (QuantizationKind::Ggml).into(),
+            ModelKind::QuantizedGGUF => (QuantizationKind::Gguf).into(),
+            ModelKind::XLoraNormal => (AdapterKind::XLora).into(),
+            ModelKind::XLoraGGML => (AdapterKind::XLora, QuantizationKind::Ggml).into(),
+            ModelKind::XLoraGGUF => (AdapterKind::XLora, QuantizationKind::Gguf).into(),
+            ModelKind::LoraNormal => (AdapterKind::Lora).into(),
+            ModelKind::LoraGGML => (AdapterKind::Lora, QuantizationKind::Ggml).into(),
+            ModelKind::LoraGGUF => (AdapterKind::Lora, QuantizationKind::Gguf).into(),
+            ModelKind::Speculative { target, draft } => (target, draft).into(),
         }
     }
 }
--- a/mistralrs-core/src/pipeline/normal.rs
+++ b/mistralrs-core/src/pipeline/normal.rs
@@ -228,10 +228,9 @@ impl Loader for NormalLoader {
             Device::Cpu
         };
 
-        let mut is_lora = false;
+        let is_lora = self.kind.is_adapted_and(|a| a.is_lora());
+
         let mut model = match self.kind {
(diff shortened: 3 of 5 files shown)


EricLBuehler:
@polarathene, thanks for proposing these changes. Currently, this code is a bit obscure, so any changes to make it more idiomatic is much appreciated. I left some comments below. It also looks like the Clippy lints are failing like #334, can you please take a look?

> I'm not sure if Lora quantized kinds were intentionally excluded here, could simply be a call to kind.is_quantized()?:

Yes, that is a mistake. Can you please add an `kind.is_quantized()` function?

> There seems to be some logic to detect a difference between lora and xlora types? (the ModelKind could better indicate that or similar?):

I think some methods on ModelKind would make this possible.

> Many of these would be collapsed to the compound type proposed by the PR:

Sounds good!

> Whereas these would need additional refactoring, but all have the same shape for parameters that they could pass in a struct?

Yes, I think we could do that. Please feel free to draft something here!1

polarathene:
Sure I'll tackle the feedback later today, current tasks for this project are:
- Wrap another refactor PR (larger).
- Get the earlier strum PR merged.
- Rebase here and resolve concerns + additions.

EricLBuehler:
Great! I think that making the code more idiomatic now before I merge #309 (Idefics 2 vision model) or #324 (Nomic embedding model) which would add lots of complexity.

polarathene:
> Can you please add an `kind.is_quantized()` function?

@EricLBuehler how should this work for `Speculative`? At a glance it seems both `target` and `draft` fields can vary in model used, the `toml-selectors`  examples I saw have one that is a base + quantized and another that is quantized + quantized.

I don't know how this type/feature works, so what would be the expectation if `is_quantized()` was called on that kind variant? Likewise I was thinking to add a `quantized_type()` and `adapter_type()` that returns an option of the quantized/adapter kind when valid else `None`. Since `target` and `draft` can differ in `ModelKind` I'm not sure how I should approach that beyond `None` 🤷‍♂️

EricLBuehler:
I think for these methods, we should return a vector of the resulting data or perhaps some sort of struct. That way, there is much more type-safe control, which really fits the spirit of this PR. For example, you can check if any are quantized with `.any()`.

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -244,23 +244,155 @@ pub enum ModelKind {
     },
 }
 
-impl Display for ModelKind {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+// TODO: Future replacement for `ModelKind` above:
+#[derive(Default, derive_more::From, strum::Display)]
+pub enum ModelKindB {
```
polarathene, line 249: **Original comment removed from commit:**
- No need to `impl Display`, strum can handle derive `Display` with customized presentation.
- DRY via compound variants (_minimize growing permutations of adapters + formats_)
- Can derive `strum::EnumIs` for parent type to impl `is_quantized()` / `is_adapter()` methods that call
   - `kind.is_quantized() || kind.is_adapter_quantized()`
   - `kind.is_adapter() || kind.is_adapter_quantized()`
- Alternatively derive `strum::EnumProperty` to assign similar information to variants and query via `get_str(prop)`

---

Due to the `Speculative` variant and advice to return a Vec of options instead, the `strum::EnumIs` was only used on the sub variants (`QuantizationKind` / `AdapterKind`). As the logic for `is_quantized()` need to be a bit more complicated.

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -244,23 +244,155 @@ pub enum ModelKind {
...
+    Ggml,
+    Gguf,
+}
+
+#[derive(Clone, Copy, strum::Display, strum::EnumIs)]
+#[strum(serialize_all = "kebab-case")]
+pub enum AdapterKind {
+    Lora,
+    XLora,
+}
```
polarathene, line 287: **Original commit commentary:**
- The compound variants are much easier to manage here as individual components via these enums.
- When `strum::Display` derives, the variants are transformed to strings as `kebab-case` (_retaining the existing `x-lora` format string_).

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -244,23 +244,155 @@ pub enum ModelKind {
...
+        }
+    }
+}
+
+// TODO: Temporary compatibility layers follow (until a future PR follow-up introduces a breaking change)
+impl Display for ModelKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", ModelKindB::from(self.clone()))
+    }
+}
```
polarathene, line 344: **Original commit commentary:**
- To work with the existing `ModelKind` type for the time being, it is cloned to convert into the new `ModelKindB`.
- `ModelKindB` has `Display` derived via strum, thus in future this will be redundant, but is already greatly simplified 👍 
- Next the `ModelKindB` methods are delegated/proxied to keep this PR simple.  A future PR will broadly apply `ModelKindB` by replacing the existing `ModelKind` enum, provided a breaking change is acceptable to introduce.

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -244,23 +244,155 @@ pub enum ModelKind {
...
+            ModelKind::QuantizedGGUF => (QuantizationKind::Gguf).into(),
+            ModelKind::XLoraNormal => (AdapterKind::XLora).into(),
+            ModelKind::XLoraGGML => (AdapterKind::XLora, QuantizationKind::Ggml).into(),
+            ModelKind::XLoraGGUF => (AdapterKind::XLora, QuantizationKind::Gguf).into(),
+            ModelKind::LoraNormal => (AdapterKind::Lora).into(),
+            ModelKind::LoraGGML => (AdapterKind::Lora, QuantizationKind::Ggml).into(),
+            ModelKind::LoraGGUF => (AdapterKind::Lora, QuantizationKind::Gguf).into(),
+            ModelKind::Speculative { target, draft } => (target, draft).into(),
+        }
+    }
```
polarathene, line 395: Simple compatibility requirement for converting `ModelKind` to `ModelKindB`.

**Original commit commentary:**
- The `derive_more` crate enables the new approach to remain simple for compound types.
- A tuple of the named field variants (_provided the type combination is unique_) can simply leverage the `Into` trait for assignment (_or as parameter input_).
  - ```rs
     let kind: ModelKindB = (XLora, GGUF).into();
     ```
   - Compared to without `derive_more::From`: 
     ```rs
      let kind = ModelKindB::AdapterQuantized { adapter: XLora, quant: GGUF };
      ```

mistralrs-core/src/pipeline/normal.rs
```diff
@@ -228,10 +228,9 @@ impl Loader for NormalLoader {
             Device::Cpu
         };
 
-        let mut is_lora = false;
+        let is_lora = self.kind.is_adapted_and(|a| a.is_lora());
```
polarathene, line 231: The PR enables setting `is_lora` prior to the match arms which simplifies that a bit 👍 

---

Not quite sure what you envisioned for the API here.

I decided to mirror `is_some` + `is_some_and` that exist for `Option` type, this way you have a bit more flexibility and under the hood it handles the vec iteration without having to think about it.

**NOTE:** I have not applied this to `gguf.rs` and `ggml.rs` as I have a large refactoring PR already touching that. I can apply the same changes there after this PR is merged.

mistralrs-core/src/pipeline/normal.rs
```diff
@@ -258,30 +257,26 @@ impl Loader for NormalLoader {
...
+                in_situ_quant.is_some(),
+                device.clone()
+            ),
+            ModelKind::QuantizedGGUF
+            | ModelKind::QuantizedGGML
+            | ModelKind::XLoraGGUF
+            | ModelKind::XLoraGGML
+            | ModelKind::LoraGGUF
+            | ModelKind::LoraGGML
+            | ModelKind::Speculative { .. } => unreachable!(),
```
polarathene, line 279: Without needing to set `is_lora` anymore, we can drop an indent level.

All other variants that are `unreachable!()` have been bundled together below the valid match arms. `Speculative` doesn't use any fields in the match so `..` is more appropriate 👍

mistralrs-bench/src/main.rs
```diff
@@ -313,14 +313,7 @@ fn main() -> anyhow::Result<()> {
...
     }
-    if use_flash_attn
-        && matches!(
-            loader.get_kind(),
-            ModelKind::QuantizedGGML
-                | ModelKind::QuantizedGGUF
-                | ModelKind::XLoraGGML
-                | ModelKind::XLoraGGUF
-        )
+    if use_flash_attn && !loader.get_kind().is_quantized()
```
polarathene, line 316: This is now simpler and covers both adapter kinds as a bugfix.

**NOTE:** `Speculative` kind will also trigger this now. I am not 100% sure if that's desired? This is due to the `any()` call, as I assume if either `target` or `draft` are quantized it may not be compatible, but likewise this does not guarantee `is_quantized == true` equates to `Speculative` having both as true.

polarathene (commented):
Context for review added, hope it's helpful 👍

polarathene:
For the CI failed checks, I am a bit confused why these do not show up for me locally? Do you have any idea?:

```bash
# Commands used by CI:
cargo clippy --workspace --tests --examples -- -D warnings
cargo fmt --all -- --check
```

I have been using those before my commit and push, and both gave me changes to apply. But running them on the same content as what I pushed does not produce the CI failures 🤔 

---

**`clippy`:**
```
error: unused import: `ModelKind`
Error:   --> mistralrs-server/src/main.rs:12:23

error: unused import: `ModelKind`
Error:  --> mistralrs-bench/src/main.rs:5:88
```

**`rustfmt`:**

```diff
-    if use_flash_attn && !loader.get_kind().is_quantized()
-    {
+    if use_flash_attn && !loader.get_kind().is_quantized() {
         warn!("Using flash attention with a quantized model has no effect!")
     }
```

mistralrs-core/src/pipeline/mod.rs
EricLBuehler: I think this breaking change is acceptable in a future PR. We're pre v1, so it is fine.

mistralrs-core/src/pipeline/mod.rs
EricLBuehler: Looks good. It seems like there are a few compat `impl`s here which perhaps you can remove in a future PR by using `ModelKindB`? It seems much cleaner.

mistralrs-core/src/pipeline/normal.rs
EricLBuehler: > I decided to mirror is_some + is_some_and that exist for Option type, this way you have a bit more flexibility and under the hood it handles the vec iteration without having to think about it.

I like this API. The old way with the boolean wasn't very idiomatic anyway.

> NOTE: I have not applied this to gguf.rs and ggml.rs as I have a large refactoring PR already touching that. I can apply the same changes there after this PR is merged.

Sounds like a good idea.

mistralrs-core/src/pipeline/normal.rs
EricLBuehler: Looks good.

mistralrs-server/src/main.rs
```diff
@@ -269,15 +269,7 @@ async fn main() -> Result<()> {
...
-    if use_flash_attn
-        && matches!(
-            loader.get_kind(),
-            ModelKind::QuantizedGGML
-                | ModelKind::QuantizedGGUF
-                | ModelKind::XLoraGGML
-                | ModelKind::XLoraGGUF
-        )
-    {
+    if use_flash_attn && !loader.get_kind().is_quantized() {
```
EricLBuehler, line 272:
```suggestion
    if use_flash_attn && loader.get_kind().is_quantized() {
```
Same as above, I think the logic should be inverted.

EricLBuehler (commented):
Thanks for the PR. I think the logic is inverted on the checks for flash attention, but other than that, I am happy to accept a breaking change for the `ModelKind` enum and look forward to the quantized pipeline refactor.

polarathene:
> I think the logic is inverted?

Good spotting! 😬 

I'll resolve that shortly 👍 

---

**Just to confirm** this concern I raised (_as it'll be hidden in the resolved review comment_):

https://github.com/EricLBuehler/mistral.rs/pull/335#discussion_r1609259445

> _**NOTE:** `Speculative` kind will also trigger this now. I am not 100% sure if that's desired?_
> _This is due to the `any()` call, as I assume if either `target` or `draft` are quantized it may not be compatible, but likewise this does not guarantee `is_quantized == true` equates to `Speculative` having both as true._

`ModelKind::Speculative { .. }` may only be partially quantized (_I've seen the `draft` always as quantized? While `target` may be quantized_).
- I assume it's better for `is_quantized()` to return `true` when there is any hint of quantization in the kind?
- Unlike every other variant, you'd have a caveat where this must be kept in mind that it does not guarantee a kind that stores multiple kinds are all quantized.
- Same applies for `is_adapted()` and the `_and(|| { ... })` methods (_which also use `.any()`_).

Using `.all()` instead doesn't resolve the concern either.. so it might need to be documented with some context that variants like `Speculative` are prone to bugs?

---

> Looks good. It seems like there are a few compat `impl`s here which perhaps you can remove in a future PR by using `ModelKindB`? It seems much cleaner.

Yes, as I wasn't sure about breaking changes being approved I've been a bit cautious. A follow-up PR can remove the compat layers here when `ModelKindB` becomes the new `ModelKind` 👍 

> I think this breaking change is acceptable in a future PR. We're pre v1, so it is fine.

I'll tackle that after my next PR (_dependent upon changes here_) is sorted out 👍

polarathene:
> > I decided to mirror is_some + is_some_and that exist for Option type, this way you have a bit more flexibility and under the hood it handles the vec iteration without having to think about it.
> 
> I like this API. The old way with the boolean wasn't very idiomatic anyway.
> 
> > NOTE: I have not applied this to gguf.rs and ggml.rs as I have a large refactoring PR already touching that. I can apply the same changes there after this PR is merged.
> 
> Sounds like a good idea.

In my upcoming PR it was a little awkward/noisy with this new approach to check if the kind has an `Adapter` and set variables `is_lora` + `is_xlora`.

`Speculative` isn't supported in the match arms for GGUF/GGML pipeline (_probably a good reason for that?_), so I can probably just assume only the first element returned from `adapter_type()` is relevant and use `if let Some(a)` with `a.is_lora()` + `a.is_xlora()` called directly instead of as a closure 👍 

---

As I'm not really familiar with these features yet, am I right to assume that **future adapters may be supported that aren't "LoRA" specific?**

I am new to this domain, so I only have a surface understanding that LoRA is a technique to direct output towards a preferred style (_I've noticed it can be applied with image, audio and text generation models_). I think it's often referred to as a way to approach "fine-tuning" as a light-weight alternative via a separate file(s), as opposed to duplicating the model (weights?) data but with the adjustments merged (_requiring much more disk space and inefficient for changing styles/adapters at runtime? Which the [LoRAX project](https://github.com/predibase/lorax) describes_).

I have no idea about X-LoRA, and haven't found time to look into it yet.

**Not too important to resolve** for now, I'll continue with my refactoring and adjust when appropriate 😅 

**EDIT:** I just took a glance at [LoRAX docs for Adapters](https://loraexchange.ai/models/adapters/#medusa) and noticed an Adapter type called `Medusa` that mentions "Speculative Decoding", which I assume is what `Speculative` is meant to represent?

polarathene:
I rebased and squashed the commit suggestions into the prior commit.

The two `main.rs` conditions for flash attention have been inverted. Should be good to merge 👍 

---

If you're new to squash merge on Github, it should provide by default the content of each commit message I wrote, and use the PR name as the squash commit name (_with link ref to this PR_).

EricLBuehler:
> Speculative isn't supported in the match arms for GGUF/GGML pipeline (probably a good reason for that?), so I can probably just assume only the first element returned from adapter_type() is relevant and use if let Some(a) with a.is_lora() + a.is_xlora() called directly instead of as a closure 👍

`Speculative` needs 2 models, so all other pipelines can just ignore the possibility of loading it.

> EDIT: I just took a glance at [LoRAX docs for Adapters](https://loraexchange.ai/models/adapters/#medusa) and noticed an Adapter type called Medusa that mentions "Speculative Decoding", which I assume is what Speculative is meant to represent?

Yes, but the Medusa method is a bit different. Speculative Decoding is a sampling technique which uses a (smaller, faster, less precise) "draft" model to accelerate inference of a (larger, slower, more precise) "target" model.

EricLBuehler (approved):
Looks good, thank you for this change! I think removing the `ModelKind` enum is not even a breaking change, actually, as it is not exported publically. Perhaps you could remove it in the GGUF/GGML pipeline refactor?

PR #336, opened 2024-05-19
Use different RoPE impl for bs=1

EricLBuehler:


EricLBuehler:
Causes performance degredation.

PR #345, opened 2024-05-25
Add support for using GGUF tokenizer

EricLBuehler:
This adds support for using a GGUF tokenizer as documented here:
https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#tokenizer

- [x] llama
  - Unigram 
- [ ] replit
  - Unigram 
- [ ] gpt
  - BPE
- [ ] rwkv
  - RWKV


Jeadie:
What remains on this PR? Need GGUF tokenizer support so happy to contribute.

EricLBuehler:
> What remains on this PR? Need GGUF tokenizer support so happy to contribute.

Currently, it doesn't work. In this PR I tried to convert the GGUF tokenizer to a HF tokenizer for easy integration with the rest of mistral.rs, but I ran into some problems with how the decoder/post processor/normalizer parts of the HF tokenizer are being set up. Additionally, it looks like the Mistral GGUF doesn't contain any merges, but the HF tokenizer itself does. I'm not sure if there are sensible defaults or ways to calculate those values from the token types that I can use.

So, the current state of this PR is that it is half working. If you could perhaps take a look and see if you can get it to work, that would be amazing!

Jeadie:
What example GGUF are you using for mistral? I don't see any reference to Mistral in [ggerganov/ggml](https://github.com/ggerganov/ggml)

EricLBuehler:
I'm using this one: https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/tree/main

Mistral uses a `llama` tokenizer, which [ggerganov/ggml](https://github.com/ggerganov/ggml) says should be a SentencePiece tokenizer. In this PR I use a BPE tokenizer, however, because the HF tokenizer is BPE (maybe that is the problem). Perhaps we can use [this crate](https://github.com/guillaume-be/rust-tokenizers?tab=readme-ov-file)?

EricLBuehler:
@Jeadie, I made some progress! It mostly works now, and I think there is just one small bug left. With this PR you can run models fully locally, specifying paths for the chat template and GGUF file:
```
./mistralrs-server --chat-template <chat_template> gguf -m . -f Phi-3-mini-128k-instruct-q4_K_M.gguf
```

PR #349, opened 2024-05-26
Propogating Regex init error

gregszumel:
I found that certain Regex constraints can panic and bring down the whole engine. For example, take this constraint: `constraint: Constraint::Regex("[a-z]*".to_string()).` We get:

```
thread '<unnamed>' panicked at mistralrs-core/src/aici/rx.rs:35:14:
dfa has no universal start state; make sure it doesn't match empty string
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at mistralrs/examples/quantized/main.rs:57:39:
called `Option::unwrap()` on a `None` value
fish: Job 1, 'cargo run --example quantized -…' terminated by signal SIGSEGV (Address boundary error)
```

Not sure if this is a solid fix, but I just pushed the error up into StackRecognizer creation. Now, the engine can handle this initialization error and return a Validation error to the client, rather than killing the whole engine.

### DIFF
--- a/mistralrs-core/src/aici/recognizer.rs
+++ b/mistralrs-core/src/aici/recognizer.rs
@@ -30,6 +30,12 @@ impl<S: Copy, R: FunctionalRecognizer<S>> StackRecognizer<S, R> {
         };
         Ok(rec)
     }
+
+    pub fn reset(&mut self) -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


EricLBuehler (approved):

EricLBuehler:
Thank you!

Issue #347, opened 2024-05-26
Enabling prefix cache for llama3 gguf


joshpopelka20gmail:
For my use case, I've been informed that prefix caching may help me reduce inference time (I'm working on an internal web service). Looking through the codebase, I see that there is some code for creating a prefix cache i.e. `prefix_cacher.rs`. 

I'm really struggling to understand if this cacher is currently being used. It doesn't seem like the no_prefix_cache is available in the python api (though I do see `with_prefix_cache_n`). Is the prefix cache being used across messages? I don't plan on batching requests as I'd like a real time service so, for my use case, it'll be one message at a time. Since it'll be a service, the prompt will have the same prefix each time.

EricLBuehler:
> I'm really struggling to understand if this cacher is currently being used. 

It is being used by default. If you set the prefix cache n to be 0, then it will not be used as no prefixes will be cached.

> Since it'll be a service, the prompt will have the same prefix each time.

Currently, we only allow verbatim matches. I'm working on allowing subset matches in #350 which should make the rate of cache hits much higher and accelerate performance for use cases such as yours.

joshpopelka20gmail:
This is excellent news! I'm looking forward to testing the new feature when you are finished. Thank you for taking the time to work on it. I was thinking I might try to write the code, but I couldn't follow the flow of the cacher to know if it was being used.

EricLBuehler:
Yes, the prefix cache is a bit complicated and is spread out over multiple files. Here's an outline, for your and future reference, of how it works:
- During the sampling stage of a pipeline step if a sequence is done, we add the final kv cache to the prefix cacher:
  - The prefix cacher object is passed: `Pipeline::step` method -> `Pipeline::sample` method -> `do_sample` macro -> `finish_and_add_tokens_to_seq` macro
  - We add and then may evict sequences here:
    > https://github.com/EricLBuehler/mistral.rs/blob/0a2fd4db244b16d0ed830aa3034b67ca5b6740ae/mistralrs-core/src/pipeline/sampling_pipeline.rs#L39-L42
- When a new sequence comes in, try to match a prefix cache:
  > https://github.com/EricLBuehler/mistral.rs/blob/0a2fd4db244b16d0ed830aa3034b67ca5b6740ae/mistralrs-core/src/engine/mod.rs#L375-L378

joshpopelka20:
I've been reading up more on prefix caching and I'm looking for a little clarification to improve my understanding. 

[...]

PR #350, opened 2024-05-27
Allow subsets of sequences in prefix cacher

EricLBuehler:
Refs #347.


EricLBuehler:
> nice, this is cool! I'll work on deleting from the prefix-cache if some byte-threshold was exceeded. Long term we could bring the trie structure back and do some vLLM style stuff with pointers to caches?

Yeah, a byte threshold metric would be a big improvement!

Regarding the `trie` structure, it should be doable for non-quantized models where I can modify the matmul implementation. I think that would improve the scaling for non-quantized models, too.

EricLBuehler:
So close:

**master** (correct):
```
> hi
Hello! How can I assist you today?
> what is graphene
Graphene is a two-dimensional material made of carbon atoms arranged in a hexagonal lattice. It is a very strong and flexible material, with a high electrical conductivity and thermal conductivity. Graphene has many potential applications in fields such as electronics, energy storage, and materials science. It is considered to be one of the most promising materials for the future, due to its unique properties and potential for widespread use.
```

**prefix_cacher_subseq**:
```
> hi
Hello! How can I assist you today?
> what is graphene
Graphene


Graphene is a two-dimensional material made of carbon atoms arranged in a hexagonal lattice. It is a highly conductive and flexible material with a high surface area and unique electronic properties. Graphene has been shown to have potential applications in various fields, including electronics, energy storage, and biomedicine.
```

Looks like a RoPE seqlen offset problem.

PR #354, opened 2024-05-28
Refactor `deserialize_chat_template`

Jeadie:
## Changes
- The macro `deserialize_chat_template` returns two objects `ChatTemplate` and `GenerationConfig`. 
- These two objects are independent when instantiating from a `impl ModelPaths`.
- Create new function to create `ChatTemplate`
- Creating a `GenerationConfig` is a one liner, no function needed.
- Also, fixed some documentation.


Jeadie:
The current implementation of `load_model_from_path` uses the `chat_template` from the `get_template_filename` regardless if the `Loader` has been instantiated with a `chat_template: Option<String>` override (e.g. [`GGUFLoader`](https://github.com/EricLBuehler/mistral.rs/blob/d2953a8ad3f8bbcdbbdb6c1546f0bba5f3889a40/mistralrs-core/src/pipeline/gguf.rs#L75). 

I think the correct behaviour should be that if the use has provided a specific `chat_template`, use it instead of the one derived from `get_template_filename`.

EricLBuehler (approved):
Looks great, thanks!

PR #356, opened 2024-05-28
refactor: GGUF + GGML Loaders with `ModelKind`

polarathene:
This is mostly where I wanted to get it functionally. Still plenty more to do, but there is a lot of churn going on lately that it's starting to introduce conflicts when rebasing, so I'd rather get this through review.

I'll provide more context in about 12 hours, but the commit history is iterative with detailed commit messages on changes. Some of it became redundant and replaced, but diff may be easier to follow if helpful.

---

- `ModelKind` workarounds dropped, so the newer version is now used as requested.
- A similar workaround is on the model `from_gguf()` and `from_ggml()` methods, I can address that in a follow-up PR. These methods now use a trait that needs to be imported to use them.
- Some potential bugs spotted and others fixed.

I'll handle the clippy and rustfmt feedback when I return 😓


mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+        // X-LoRA support:
+        let mut xlora_paths: Vec<PathBuf> = vec![];
+        let mut xlora_config: Option<XLoraConfig> = None;
+        if is_xlora {
+            xlora_paths = vec![paths.get_classifier_path().as_ref().unwrap().to_path_buf()];
+            xlora_config = Some(paths.get_classifier_config().as_ref().unwrap().clone());
+        }
+
+        // Create VarBuilder:
+        // TODO: `from_mmaped_safetensors` has `xlora_paths` as the 2nd param, is this a bug?
```
EricLBuehler, line 69: No, perhaps we should rename `xlora_paths` to `xlora_classifier_path` though as that is really what it carries here.

The 2nd param of `from_mmaped_safetensors` takes the adapters paths, perhaps we should rename that accordingly too?

mistralrs-core/src/xlora_models/gemma.rs
```diff
@@ -736,6 +736,7 @@ impl NormalModel for XLoraModel {
     fn device(&self) -> &Device {
         &self.device
     }
+    // BUG?: This is the only x-lora model with `is_xlora() -> false`
```
EricLBuehler, line 739: *Yes*! Thanks for catching that. As a long term and safer solution, maybe we can get this information from the `ModelKind` instead of using hardcoded values? Just an idea.

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -457,10 +396,13 @@ pub struct GeneralMetadata {
...
     pub has_no_kv_cache: bool,
-    pub is_xlora: bool,
     pub num_hidden_layers: usize,
     pub eos_tok: Vec<u32>,
-    pub is_lora: bool,
+    // Model utilizes an adapter (LoRA or X-LoRA):
+    pub has_adapter: bool,
+    // TODO: There is a possibility that code mistakenly queries only `is_xlora`,
+    // rather than `has_adapter`. Thus may actually only support X-LoRA? (potential bugs?)
+    pub is_xlora: bool,
```
polarathene, line 405: Might need your input here for clarification if `is_xlora` is exclusive to X-LoRA or also represents plain LoRA, and if `AdapterKind` were to have another variant is there value in separate `has_adapter` and `is_xlora`?

The enum itself `AdapterKind` could potentially be stored instead in future to cover both needs.

mistralrs-core/src/utils/model_config.rs
EricLBuehler, line 1: This is really great, it makes the code much more type safe and (I think) easier to read! It looks like there are a few clippy warnings here, some of which, like the enum size one, can probably be ignored.

EricLBuehler (commented):
@polarathene, this looks excellent, and I think it makes it much more idiomatic and type safe which is always better. I wrote some comments, and it looks like there are a few clippy warnings, one of which can be ignored.

mistralrs-core/src/utils/model_config.rs
polarathene, line 69: I've looked at this utils file and saw that it effectively repeats code 4 times for slight variations. I have a WIP change to condense that down considerably.

There was some churn with refactorings in the past though, and it did seem a bit odd with how you've got `xlora_paths` as the first param here, but it's treated as the 2nd param in the method itself. Perhaps at one point they were different, or this accidentally happened when you introduced xlora support?

If you could let me know I'll address it in a future PR, I've not changed the param order below from what it is presently prior to this PR.

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+#[buildstructor::buildstructor]
+impl<'a, Q: QuantParams> ModelParams<'a, Q> {
+    #[builder]
+    pub fn new<'b: 'a>(quant: Q, adapter: Option<Adapter<'b>>) -> Self {
+        match adapter {
+            None => Self::Quantized((quant, NoAdapter {}).into()),
+            Some(a) => Self::Adapted((quant, a).into()),
+        }
+    }
+}
```
polarathene, line 141: An example of `buildstructor`. Usage is simple as seen in the loaders, `builder()` + method per parameter. Once `build()` is called it'll run the logic in this method which produces the GGUF / GGML enum from earlier.

If you were ever to implement support for EXL2 or some other format, it should keep it fairly simple 👍

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+        reader: &mut R,
+        device: &candle_core::Device,
+        lora_config: &[((String, String), LoraConfig)],
+        vb: &VarBuilder,
+        ordering: &Ordering,
+        xlora_config: Option<XLoraConfig>,
+        mapper: DeviceMapMetadata,
+        preload_adapters: &Option<HashMap<String, (VarBuilder, LoraConfig)>>,
+    ) -> Result<Self, candle_core::Error> where Self: Sized;
+}
```
polarathene, line 183: All this will get replaced in a future PR.

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+            device,
+            lora_config,
+            &vb,
+            ordering,
+            xlora_config,
+            mapper,
+            &preload_adapters,
+        )
+    }
+}
```
polarathene, line 276: All this is just mapping to those traits expected sequence of params. It'll disappear too once those models `from_gguf()` / `from_ggml()` take the param structs instead.

I don't think there's much value in calling those directly (_if that's supported by the Rust / Python APIs_) vs the loaders, so I assume the more logical grouping with structs for fewer param inputs (and less strict order) will be an improvement 👍

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    impl TryFrom<ModelParams<'_, GGUF<'_>>> for *models_gguf_a {
+        type Error = candle_core::Error;
+
+        fn try_from(params: ModelParams<'_, GGUF<'_>>) -> Result<Self, Self::Error> {
+            let config = params.expect_adapted("`Config` should be GGUF Quantized with an Adapter");
+            config.try_into_model()
+        }
+    }
+}
```
polarathene, line 328: Here's the `akin` + `variantly` usage. Should be fairly simple to grok.

`akin` has you wrap the block of code to duplicate and define vars with `let &var_name` assigned with an array of values. The code block is then repeated with each item replacing the deref `*var_name`. You can have multiple arrays, but should keep the length consistent, otherwise it'll attempt to fill in the blanks via items from another array.

`variantly` usage is quite basic. That `Quantized`/`Adapted` enum from earlier is passed in as `params` and the wrapped `Config` struct is extracted like an options `expect()` / `unwrap()` would do, basically `Option<_>` methods but for enums.

mistralrs-core/src/xlora_models/gemma.rs
polarathene, line 739: Pretty sure I spotted a bug here? Could you confirm?

polarathene (commented):
Context for review @EricLBuehler 

I'll go take care of appeasing the lint gods now, followed by resolving the conflict.

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+        let mut xlora_paths: Vec<PathBuf> = vec![];
+        let mut xlora_config: Option<XLoraConfig> = None;
+        if is_xlora {
+            xlora_paths = vec![paths.get_classifier_path().as_ref().unwrap().to_path_buf()];
+            xlora_config = Some(paths.get_classifier_config().as_ref().unwrap().clone());
+        }
+
+        // Create VarBuilder:
+        // TODO: `from_mmaped_safetensors` has `xlora_paths` as the 2nd param, is this a bug?
+        let vb = from_mmaped_safetensors(
```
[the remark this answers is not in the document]
polarathene: > The 2nd param of from_mmaped_safetensors takes the adapters paths, perhaps we should rename that accordingly too?

Yes. I think more clearer names or at least some associated comments can assist with maintenance. Especially those that aren't familiar with all the ML / LLM jargon and only understand how to read code 😓 

Can address that in a smaller follow-up PR and you can suggest the new names then? As mentioned I already was working on `from_mmaped_safetensors()` rewrite. Hopefully no merge conflicts there when I return to it 😝

mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -657,8 +583,8 @@ impl Pipeline for GGUFPipeline {
         }
     }
     fn activate_adapters(&mut self, adapter_names: Vec<String>) -> anyhow::Result<usize> {
-        if !self.metadata.is_lora {
-            anyhow::bail!("Cannot activate adapters non-LoRA models.")
+        if !self.metadata.has_adapter {
```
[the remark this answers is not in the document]
polarathene: Ok, could we better communicate this?

LoRA models share the same as X-LoRA when they're created? GGUF/GGML is converted into `XLoraLlama` for example, but this has an `is_xlora() -> true` method.

Something seems off, or for someone like me not too familiar with these terms, other than you were involved in developing X-LoRA, I don't know how the usage differs in respect to `XLoraLlama` vs `GeneralMetadata` since `is_xlora` and `is_xlora()` kind of conflict there, but the model and pipeline are presumably related?

---

Is `activate_adapters()` only meant for LoRA? Or for any potential `AdapterKind` that isn't X-LoRA?

```rs
// Pseudo-code of what might be a better way to query?
if self.metadata.adapter.is_some_and(|a| !a.is_xlora()) { activate } else { bail }
```

mistralrs-core/src/utils/model_config.rs
```diff
@@ -0,0 +1,328 @@
...
+
+// NOTE: Below is a workaround to proxy params to the existing API methods `get_gguf()` / `get_gmml()` traits covered above.
+impl Config<GGML, NoAdapter> {
+    pub fn try_into_model<T: FromGGML>(self) -> Result<T, candle_core::Error> {
+        // Destructure props:
+        let GGML(
+            FileGGML { ct, gqa },
+        ) = self.quant;
+
+        // Forwards all structured fields above into the required flattened param sequence:
```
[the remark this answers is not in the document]
polarathene: Yes, I plan to, but as mentioned I wanted to minimize running into conflicts. My first approach was just getting something that was an improvement and moving most changes into this `model_config.rs` module to minimize conflicts as I spent quite a bit of time trying a few different things 😅 

I still have considerations for `ModelKind` to move the quantization and adapter into the `Plain` / `Normal` variant via struct fields.

After this PR I can look into addressing the concern with the prop drilling (_`main.rs` of each CLI seems to have a lengthy chain to go from CLI args to Loader_). There's plenty to tackle, I've just got to do it in smaller bursts I think due to the activity in the repo going on.

mistralrs-core/src/xlora_models/gemma.rs
polarathene: 💯 I was thinking that made sense too.

Should `is_xlora()` be `true` here when the `AdapterKind` is `Lora` instead of `XLora`? That was one of my concerns about how these two are treated since presently these would return `true` for `Lora` kind.

polarathene:
I think I've addressed the feedback to get this reviewed again and merged. I have a checklist above of feedback to still tackle but I don't think it needs to block this PR from merging.

EricLBuehler:
I'll write a review soon, but just responding to your comments:

> I'm not sure what i is intended for exactly, but with the current approach since there is a continue condition earlier, an entry can be skipped but i will increment. Is that context relevant? Or do the numbers assigned not need to be sequential, it's ok to skip some i?

`i` is the adapter number. It's +1 because adapter 0 is not what HF/peft does.

> EDIT: Oh, there's a markdown example with using Python that does use the ModelKind enum, I'm not quite sure what the syntax for that would be with the new variant ModelKind::Quantized { quant: QuantizationKind::Gguf }, perhaps that would benefit from a new_*() helper method 🤷‍♂️

If you could add a more type-safe option to the Python API that would be great too.

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -458,10 +397,11 @@ pub struct GeneralMetadata {
     pub repeat_last_n: usize,
     pub tok_trie: Arc<TokTrie>,
     pub has_no_kv_cache: bool,
-    pub is_xlora: bool,
     pub num_hidden_layers: usize,
     pub eos_tok: Vec<u32>,
-    pub is_lora: bool,
+    pub kind: ModelKind,
+    // TODO: Replace is_xlora queries to check via kind instead:
```
EricLBuehler, line 403: Do you want to do this in this PR or a future one?

EricLBuehler (commented):
Looks good, almost ready to merge! Just one question about a TODO.

mistralrs-core/src/pipeline/mod.rs
polarathene: I can do it in a follow-up PR if you want to get this merged. Either way I can tackle it after I've had some rest, it's my cut off point for the day now :)

EricLBuehler: Sounds good, I'll merge this.

EricLBuehler (approved):
@polarathene, is it OK to merge? Looks good on my end.

polarathene:
Yes happy to get this merged! 🚀

EricLBuehler:
Sounds good! I wanted to check and see if you wanted to implement the other points, but I assume you plan on doing that in a future PR?

EricLBuehler:
Thank you!

PR #357, opened 2024-05-29
Add an example

EricLBuehler:
Also some fixes


mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -356,7 +360,21 @@ impl Loader for GGUFLoader {
...
+        } = if paths.get_tokenizer_filename().to_string_lossy().is_empty() {
+            convert_ggml_to_hf_tokenizer(&model)?
+        } else {
+            ConversionResult {
+                tokenizer: get_tokenizer(paths.get_tokenizer_filename())?,
+                bos: None,
+                eos: None,
+                unk: None,
+            }
+        };
```
polarathene, line 377: I really don't want to deal with resolving conflicts again, this will be annoying for me to rebase against like the debug addition was.

polarathene: and... you just merged it 😑

EricLBuehler: Ah, sorry! I didn't see the comment in time. Is there anything I can do to help it?

polarathene: Not really, but due to the number of commits and my change being early at the point straight after this addition, I get conflicts rippled through the history.

I'll probably be lazy and just use a merge commit to resolve instead of a proper rebase. I raised the PR before going to bed since there is a lot of activity recently on the files I was refactoring and it's starting to get tedious, so best to get what I've done so far in when you have time to review 👍

EricLBuehler: Ok, sounds good, a merge commit is fine. I'll make a review now and then maybe you can do the rebase/merge?

PR #358, opened 2024-05-29
remove tracing_subscriber.init() from Loader

Jeadie:
- Running `tracing_subscriber.init()` within a library will break most uses of this library.


EricLBuehler (changes requested):
Sounds good. Can you please add this to the following crates (after the model has been loaded):
- `mistralrs-server`
- `mistralrs-bench`
- `misrtalrs-pyo3` (in the module initialization function towards the end of `lib.rs`)

EricLBuehler:
@Jeadie I just merged #428 which implements this plus a bit more.

PR #361, opened 2024-05-30
fix: Ensure committed files are normalized to LF

polarathene:
This avoids accidentally committing files with CRLF line endings which `rustfmt` will happily normalize to LF creating a very noisy diff from these invisibles.

---

## Additional context

As I'm presently developing from Windows with VSCode and I haven't adjusted defaults yet, when I create new files they default to CRLF for line endings.

I was only aware of this after running `rustfmt` and noticing every single line had been changed in the diff, even though there was no visual difference the invisibles change from CRLF to LF was the cause (_since I had already committed this file quite a while ago with CRLF before running `rustfmt`.

`.gitattributes` is a useful file to prevent that sort of mistake from occurring as the project gains more contributors. It also helps ensure better compatibility when certain files should be checked out with LF or CRLF specifically, which I've added a contextual comment for the maintainers benefit.

---

`Dockerfile` can be one of the cases where you want to ensure LF is used as checkout, since `docker build` can fail `RUN` directives under some contexts (_[as I have experienced in the past](https://github.com/docker-mailserver/docker-mailserver/pull/3755)_):

```Dockerfile
FROM alpine

# This will work:
RUN chmod +x /bin/sh

# However the HereDoc feature retains the
# trailing CRLF, triggering a build error:
# (Fails with sh/bash/ash)
RUN <<EOF
  chmod +x /bin/sh
EOF
```

```console
$ docker build --no-cache --tag crlf-example -f Dockerfile .

# ...                                                                                                                                                      
 => CACHED [1/2] FROM docker.io/library/alpine:latest
 => [2/3] RUN chmod +x /bin/sh
 => ERROR [3/3] RUN <<EOF (chmod +x /bin/sh)
------
 > [3/3] RUN <<EOF (chmod +x /bin/sh):
: No such file or directory
------
Dockerfile:9
--------------------
   8 |     # (Fails with sh/bash/ash)
   9 | >>> RUN <<EOF
  10 | >>>   chmod +x /bin/sh
  11 | >>> EOF
  12 |
--------------------
ERROR: failed to solve: process "/bin/sh -c   chmod +x /bin/sh\r\n" did not complete successfully: exit code: 1
```

Likewise for any scripts like `.sh` relying on the shebang to be executed that are added to the image via `COPY`, since the shebang line would have CRLF in environments that default to that otherwise without `eol=lf`. Similar for `.bat` which should use `eol=crlf`.

---

`Dockerfile-cuda-all` was renamed to a more conventional `Dockerfile.cuda-all`.

Most editors will better recognize `Dockerfile.*` or `*.Dockerfile` (_which [Dockers official docs suggest](https://docs.docker.com/build/building/packaging/#filename) as convention_). I went with `Dockerfile.` as a prefix for better co-location, similar to what the repo is already doing.


EricLBuehler (approved):
Thanks!

PR #370, opened 2024-06-02
Fix chat template and F16/BF16 CUDA GEMM when RUST_BACKTRACE

EricLBuehler:


EricLBuehler:
- Fixes chat template deserialization
- Fixes CUDA GEMM cap when `RUST_BACKTRACE=1`.

Issue #329, opened 2024-05-19
bug: If device layers requested exceed model layers, host layers overflow


polarathene:
## Describe the bug

If they number of device layers exceed the models, then the host layers to assign seems to wrap/overflow instead of the expected `0`.

**NOTE:** With `llama-cpp` you can configure a larger number of layers and host layers will remain `0` while only the needed layers are used as device layers.

```console
RUST_BACKTRACE=1 ./mistralrs-bench --num-device-layers 33 gguf -m . -t . -f Hermes-2-Pro-Mistral-7B.Q4_K_M.gguf
# ...
INFO mistralrs_core::device_map: Using 33 layers on device and 18446744073709551615 on host.

thread 'main' panicked at library/alloc/src/raw_vec.rs:25:5:
capacity overflow
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: alloc::raw_vec::capacity_overflow
   3: <T as alloc::vec::spec_from_elem::SpecFromElem>::from_elem
   4: mistralrs_core::device_map::DeviceMapMetadata::into_mapper
   5: mistralrs_core::models::quantized_llama::ModelWeights::from_gguf
   6: <mistralrs_core::pipeline::gguf::GGUFLoader as mistralrs_core::pipeline::Loader>::load_model_from_path
   7: <mistralrs_core::pipeline::gguf::GGUFLoader as mistralrs_core::pipeline::Loader>::load_model_from_hf
   8: mistralrs_bench::main
```

**Context:**
- [`Q4_KM` GGUF model](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B-GGUF) used, but this applies to any model where the layers is exact.
- Since `mistral.rs` seems to enforce loading first through HuggingFace API calls, I've worked around that by allowing 401 (Unauthorized) panic [as described here](https://github.com/EricLBuehler/mistral.rs/issues/326#issuecomment-2118647910), while unlike `llama-cpp` additional config files is enforced... I sourced those from [here](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B), but it lacked a `tokenizer.json` file so I gave it one from another model (_this has no relevance on the error encountered_).

---

## Additional feedback

`mistral.rs` output doesn't information like layers as clear to me as `llama-cpp`, and I don't know if there's some sort of `inspect` command to output/query the metadata?

[...]

PR #373, opened 2024-06-03
Fix tokenizer write on read only file system

EricLBuehler:


polarathene:
Not sure if this is relevant, but a [recent change in `tokenizers`](https://github.com/huggingface/tokenizers/pull/1513) ignores added tokens in the decoder. Which may affect the relevance of this utility / workaround? 🤷‍♂️ 

---

For context on this utility method, while the logic itself is pretty clear that it adds the token `content` field as a lookup key for the associated ID field of the token object in `added_tokens`, it may be helpful to have a little reference of the file content. So if anyone lands here via `git blame` I documented the following [here](https://github.com/EricLBuehler/mistral.rs/issues/329#issuecomment-2143348589) (_collapsed section at end of linked comment_):

<details>
<summary>Relevant context snippets (click to view)</summary>

> image
> image

The utility itself is to avoid these log lines, that I only seemed to trigger with a llama3 `tokenizer.json`:

> image
> image

</details>

Similar to what is found in `added_tokens.json` (_this file is at least present in [`NousResearch/Hermes-2-Pro-Mistral-7B`](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B) which lacks a `tokenizer.json` and has `tokenizer.model` instead_).

ArthurZucker:
The `added_tokens.json` is an old format btw

EricLBuehler:
> The added_tokens.json is an old format btw

We do not use `added_tokens.json` as a backend, but thank you for letting us know!

PR #378, opened 2024-06-04
adding reboot functionality

gregszumel:
This is something I whipped up, thoughts? The basic idea is to be able to restart the tokio runtime if it for some reason goes down.

It introduces a few new pieces, mainly:
- Reboot state, which just saves the args to `Engine::new
- adds the `reboot` function, which updates the `MistralRs` sender
- Modifies the `MistralRs.sender` to have a `RwLock`, so we can safely swap it out with a new sender if the engine goes down
- Since we now have a `RwLock` on the sender, the `get_sender()` function can potentially fail, so now `get_sender()` returns a Result.

Before I keep going on this, wanted to get anyone's thoughts on it!

Maybe still to do
- I have not worked on the PyO3 stuff, I'm not super familiar with it -- is anything needed there in terms of supporting a restart?
- Make more formalized errors to allow not using the anyhow Errors in the main MistralRs struct


mistralrs-core/src/lib.rs
```diff
@@ -234,8 +259,66 @@ impl MistralRs {
...
     }
 
-    pub fn get_sender(&self) -> Sender<Request> {
-        self.sender.clone()
+    /// attempts to reboot the engine, if the sender (only way to communicate with
+    /// the engine) is closed
+    /// TODO: GS don't use anyhow for these errors?
+    pub fn reboot_engine(&self) -> anyhow::Result<()> {
+        tracing::info!("attempting to reboot");
+        // only start a new runtime if the reciever was closed. this implies
```
EricLBuehler, line 267: Nit: Looks like there is a type here.

mistralrs-core/src/scheduler.rs
```diff
@@ -45,8 +44,9 @@ pub struct SchedulerOutput<'a> {
 /// step of the engine. For each scheduling step, the scheduler method is used if there
 /// are not only running, only waiting sequences, or none. If is it used, then it
 /// is used to allow waiting sequences to run.
+#[derive(Clone)]
 pub enum SchedulerMethod {
-    Fixed(UsizeBounded<1, { usize::MAX }, false>),
```
EricLBuehler, line 49: Is there a reason why you removed the UsizeBounded usage? It upheld the invariant that there bound will be strictly positive.

mistralrs-server/src/chat_completion.rs
```diff
@@ -336,7 +336,22 @@ pub async fn chatcompletions(
             return ChatCompletionResponder::InternalError(e.into());
         }
     };
-    let sender = state.get_sender();
+    let sender = state.get_sender().unwrap();
+
+    // if the sender is closed, try to reboot the engine, and get the sender
+    // TODO: GS - make more readable?
```
EricLBuehler, line 342: I assume you were trying to add your name here? If you could change this to `TODO(GS)` or something like that, it would make the code easier to grep and manage. Thanks!

mistralrs-server/src/completions.rs
```diff
@@ -160,7 +160,21 @@ pub async fn completions(
         );
     }
     let request = parse_request(oairequest, state.clone(), tx);
-    let sender = state.get_sender();
+    let sender = state.get_sender().unwrap();
+    // if the sender is closed, try to reboot the engine, and get the sender
+    // TODO: GS - make more readable?
```
EricLBuehler, line 165: Same as above.

mistralrs-core/src/lib.rs
```diff
@@ -234,8 +259,66 @@ impl MistralRs {
...
+                    engine.run().await;
+                });
+            });
+            Ok(())
+        } else {
+            Err(anyhow::Error::msg("Did not reboot, sender not closed"))
+        }
+    }
+
+    /// TODO: GS don't use anyhow for these errors?
```
EricLBuehler, line 298: Same as other comment, replace `TODO: GS` with `TODO(GS)` or something similar for easier discoverability.

EricLBuehler (commented):
Thank you for adding this functionality. My concern is that we will not know if the engine itself has crashed and therefore there is no more sender to reboot to. I think that if the sender was closed, we should check if the engine itself is alive somehow before restoring it. 

Also, I assume that you left out the rebooting logic from interactive mode on purpose? I think that we should include this logic though in the `mistralrs_pyo3::Runner` methods `send_completion_request` and `send_chat_completion_request`.

mistralrs-core/src/scheduler.rs
gregszumel: Didn't immediately realize that was the invariant it was holding up, but makes total sense now.

Removed it because I couldn't find a great way to get a clone from it with 5 minutes of trying, but I'll take another stab

gregszumel:
@EricLBuehler thanks for the thoughts! 

> My concern is that we will not know if the engine itself has crashed and therefore there is no more sender to reboot to. I think that if the sender was closed, we should check if the engine itself is alive somehow before restoring it.

Agreed - let me think more about this before my next commit here

> Also, I assume that you left out the rebooting logic from interactive mode on purpose? I think that we should include this logic though in the mistralrs_pyo3::Runner methods send_completion_request and send_chat_completion_request.

Will add those when I've made another stab at the above

gregszumel:
Okay this is my next stab, here's a quick overview:

- On init, save engine parameters into the `RebootState`, so we can reboot a 'fresh' engine if needed (not purely fresh, uses the same pipeline).
- On init, also save the engine thread handler into `engine_handler`, so we can check if the thread has completed
- On `get_sender`, checks to see if the engine thread has completed. If it has, then the engine is dead.
- Then we try to reboot the engine by setting up a new engine, and storing a new mspc channel sender + engine handler

This is a little more robust than the last approach; previously, maybe somehow the sender was dropped without the engine going down. In this new approach, the engine has definitely crashed if the thread has returned. 

Also this now auto-reboots on get_sender(), so no extra logic is needed on the server side.

I don't believe there's a potential for deadlock with the `RwLocks` on the `sender` + `engine_handler`, just added that comment in `reboot_engine` for clarity. It could probably be slightly more efficient; i.e., if multiple senders try to reboot the engine at the same time, we might have a bit of holdup on the locks, but I don't think it should be too drastic since we also check if the engine is alive in reboot_engine before actually rebooting (which prevents multiple subsequent reboots).

Thoughts on this newest version? Still some to-dos with the error stuff, but I'll handle that if this looks like a solid approach.

EricLBuehler (commented):
Looks great, I think this is almost ready to merge after you fix the errors you mentioned.

gregszumel:
Here's a version that adds some error types, reworks the logging (warns on a reboot needed), and resolves the merge conflict.

I was able to test that adding a panic into the engine.run() did crash the runtime and it was able to restart (before immediately crashing again because of the panic).

Not sure why the windows check failed, let me know if there's something I can do about that -- looks like a tokenizer didn't have permissions?

EricLBuehler:
Ok, great!

The Windows check is flaky for some reason.

mistralrs-core/src/lib.rs
```diff
@@ -240,11 +277,81 @@ impl MistralRs {
...
+        })
+    }
+
+    /// attempts to reboot the engine, if the sender (only way to communicate with
+    /// the engine) is closed
+    fn reboot_engine(&self) -> Result<(), MistralRsError> {
+        let (new_sender, rx) = channel(10_000);
+        let reboot_state = self.reboot_state.clone();
+        let mut sender_lock = self.sender.write().map_err(|_| {
+            tracing::warn!("couldn't get write lock on the sender during reboot attempt");
```
EricLBuehler, line 300: This is a small thing, but all the logging messages in mistralrs are capitalized :smile: If you could change them accordingly that would be great.

gregszumel:
I don't think the clippy stuff is from this PR -- let me know if there's anything else to do for this one!

EricLBuehler (approved):
Looks good, thank you!

PR #379, opened 2024-06-04
Support multiple GGUF files

EricLBuehler:
Support multiple GGUF files by refactoring the GGUF `Content` usage.


polarathene:
Oh no I was a bit too slow getting my PR ready 💀 

I'm not used to contributing on a project where the change I work on is so prone to conflicts from other activity 😓 Should I have a draft PR that I iterate on instead to better communicate this?

README.md
```diff
@@ -283,6 +283,14 @@ please consider using the method demonstrated in examples below, where the token
 **Supported GGUF tokenizer types**
 - `llama`
 
+Some GGUF models are very large and are sharded into multiple files. Mistral.rs supports this, and to use it, delimit the `.gguf` filenames with a space as such:
+
+```bash
+./mistralrs-server --chat-template <chat_template> gguf -m . -f "a.gguf b.gguf"
+```
+
+For the Python API, a list of strings is also accepted for this case.
```
polarathene, line 292: Do you have an example GGUF file where this is actually the case? Or was this a misunderstanding from referenced motivation in https://github.com/EricLBuehler/mistral.rs/issues/380 ?

EricLBuehler: Yes: https://huggingface.co/mradermacher/dolphin-2.9-mixtral-8x22b-GGUF/tree/main

EricLBuehler:
> I'm not used to contributing on a project where the change I work on is so prone to conflicts from other activity 😓 Should I have a draft PR that I iterate on instead to better communicate this?

Would that be for developing a new feature? If so, that sounds good. Otherwise, we have a Discord for this purpose :)

README.md
polarathene: I don't understand why the sharding is necessary other than as [this comment](https://www.reddit.com/r/LocalLLaMA/comments/1c2dfv6/comment/kzb1qpt/) suggests, to workaround size restrictions on a file host?

Seems like it'd be more appropriate to have a tool that concatenates the files together if there's a UX issue you want to address, but I don't really see the point in adding additional complexity / maintenance to support something that should realistically be addressed outside of `mistral.rs` (runtime) 🤷‍♂️ 

Similar to the tokenizer patching you added recently, these are workarounds that seem more appropriate as CLI tools to apply "fixes" 🤔

polarathene:
> Would that be for developing a new feature?

No, I was just refactoring on the GGUF tokenizer, but this PR touched quite a bit that I know rebasing my work won't be something I'm interested in doing.

> Otherwise, we have a Discord for this purpose :)

I think you saw my github comment on candle about what I was doing before this PR was merged, you had given it a 👍 

Good to know though, my mistake.

EricLBuehler:
Ah sorry, I thought that was a different refactor. I will roll back these changes, as it would be great to see your changes.

Generally, multiple GGUF files is a rare occurrence. Would your PR include support for those?

polarathene:
> Would your PR include support for those?

No, it was just focused on tidying up the GGUF metadata tokenizer file you already put together, and would then have had the `from_gguf()` model methods also leverage that to tidy that up a little.

On Linux, from what I've read the file is split in parts that you can just run something like:

```bash
cat goliath-120b.Q6_K.gguf-split-* > goliath-120b.Q6_K.gguf
```

A small CLI tool could do similar to glob the related files and merge the output. I don't know for sure but I assume users would be fine with piecing it back into a single file.

I haven't spent much time reading over your PR changes for the feature, so I could be mistaken on your integration to workaround multi-part files, but at a glance it seemed to interleave a bit (_multiple loops and other additions to carry support_).

Since I didn't implement my PR on the model files, the conflicting files aren't much. I just don't have the energy atm to compare changes I'd need to do.

EricLBuehler:
Nice! Perhaps we can add such a tool. This PR was a *bit* hackey, so I actually just rolled back this PR. There is just 1 (easy) conflict now in #389.

Issue #386, opened 2024-06-05
Load chat template from GGUF file


EricLBuehler:
As discussed in #326, we have support for loading the tokenizer from a GGUF file, but not the chat template. Currently using the tokenizer model ID or specifying a chat template file is necessary (see the example chat template files [here](https://github.com/EricLBuehler/mistral.rs/tree/master/chat_templates)). This feature will be helpful for loading a GGUF model from a GGUF file only.

EricLBuehler:
Implemented in #416.

PR #388, opened 2024-06-05
Support using GGUF chat template

EricLBuehler:
This references #386 and is the last step for #326.

- [x] Support discovering GGUF chat template
- [x] Support overriding chat template
- [ ] Update docs
- [ ] Testing


EricLBuehler:
Will recreate.

Issue #380, opened 2024-06-04
Support loading multiple GGUF files


EricLBuehler:
Currently, we only support loading one GGUF file, as noticed by @ShelbyJenkins. #379 will enable loading multiple files.

PR #389, opened 2024-06-05
Refactor: GGUF metadata tokenizer

polarathene:
This was my progress on a refactor, but I didn't communicate this effort well with maintainers.

- A similar effort was recently made and merged with a slightly different approach. I thus haven't adapted the `from_gguf()` methods in model files that could have used the approach I proposed here (_a simple `TryFrom` on a struct of metadata fields_). I was considering a derive macro to simplify that further, but it might have been overkill for the project 😅 
- I did [propose upstreaming this into Candle directly](https://github.com/huggingface/candle/issues/2245), but like my other issues/PRs raised upstream I don't know if I'll get a timely response (_my PR to the HF API crate still has no activity_).

The merged alternative was [rather extensive in changes](https://github.com/EricLBuehler/mistral.rs/pull/379), I don't have the motivation to go through all that. Some structs were renamed, methods changed, the gguf tokenizer file moved along with other changes that'll bring conflicts for rebasing.

---

If this PR were updated, it would simplify the `from_gguf()` metadata parameter extraction:

https://github.com/EricLBuehler/mistral.rs/blob/89dea1b1fc80213550eed53fb0020e188f092643/mistralrs-core/src/models/quantized_llama.rs#L274-L287

But without the derive proc macro approach, due to all the `usize` and defaults, this PR may not be that worthwhile of an improvement.

---

The related tests changeset is probably still worth having. It fixes a bug with the encoder test case being misconfigured. It also removes the need for the `dead_code` annotations.

Each commit has a commit message for additional context.


EricLBuehler:
Hi @polarathene thanks. If I roll back the changes s.t. #380 was effectively never merged, would that work for you?

EricLBuehler:
Ok, I reverted and there is just 1 simple merge conflict now :)

polarathene:
> If I roll back the changes such that #380 was effectively never merged, would that work for you?

Thanks! Much appreciated 🙏 

Sorry about this, I'll try to better communicate when I'm working on a refactor in future 😓 

I hope I didn't waste any of your time. You still had some nice changes I think, such as better restructuring of the files (_something I've not given too much thought to_).

EricLBuehler:
No problem! In the future, f you let me know in Discord, I'll see it immediately so we can coordinate better.

> You still had some nice changes I think, such as better restructuring of the files (something I've not given too much thought to).

Maybe you can do something similar? A key point is that each GGUF model has an` md_get` closure, which we can refactor with our own `Content` type? Perhaps you can find a better way to do this though :)

polarathene:
> A key point is that each GGUF model has an` md_get` closure

That's no longer necessary if you take the approach I did here with `PropsGGUF` (not the best name) stuct, just impl a similar `TryFrom` that maps all the keys and it should look a little bit more terser.

> which we can refactor with our own Content type?

I don't know if we need to beyond the `MetadataContext` (also not a great name choice) struct? Unless there are other parts that would benefit, I was just focused on a cleaner API for extracting the needed metadata fields :)

---

I'll take a look after some rest, past midnight here 😅 

Failing test is for the last encode test case:

HF tokenizer has `<s> ` prepended, that's the only difference in content vs GGUF tokenizer. I'm not sure why that is. I know there was a bug in this particular test case, so this might have just surfaced what was already there, rather than something I've introduced?

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -19,115 +20,71 @@ pub struct ConversionResult {
...
+        props.added_tokens.as_ref().map(|x| x.len()).unwrap_or(0),
+        props.merges.as_ref().map(|x| x.len()).unwrap_or(0),
+        props.scores.as_ref().map(|x| x.len()).unwrap_or(0),
+        model = props.model,
     );
     if DEBUG.load(Ordering::Relaxed) {
         info!("Tokenizer: {tokenizer:?}");
     }
+
+    let [bos_str, eos_str, unk_str] = special_tokens
```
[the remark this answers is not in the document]
polarathene: > In the original code, the unk token is optional.

`unk` is still optional. I did shift some of the logic around but `unk_str` began as unset, relying upon `unigram` to set it, otherwise panic.

All I did was have the `special_tokens` returned from the supported tokenizer (eg: unigram) which made more sense to me since the final portion adds these tokens in the same manner you were setting them.

I've not changed the expectation though, below you assume `unk_str` is some value and set it as `Some(unk_str)`. If `unk_str` should be `Option<String>` instead, I can adjust for that.

mistralrs-core/src/utils/gguf_metadata.rs
```diff
@@ -0,0 +1,93 @@
+use akin::akin;
+use anyhow::ensure;
+use anyhow::Result;
+use candle_core::quantized::gguf_file;
+use std::collections::HashMap;
+
+pub struct MetadataContext<'a> {
```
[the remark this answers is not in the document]
polarathene: > Can you please extend its usage to the `quantized_*.rs` in `models` and `xlora_models`?

Yes, I can tackle that in about 12 hours.

Just need guidance on the current test failure.

---

> we can incorporate some subtyping support for backwards compatibility
> (for example for GGUF, u64 fields were previously documented as being u32. A PR of mine was merged into Candle).

I'm aware of the PR I followed that one. It might be a bit awkward to get similar support here.

I match a rust primitive type with the expected `Value` method to get that type. If the wrapped value is not valid, the `Value` method will return an error. Working around that might be more hassle than improving the `Value` enum upstream?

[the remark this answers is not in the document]
EricLBuehler: > Just need guidance on the current test failure.

I'll take a look. 

> I match a rust primitive type with the expected Value method to get that type. If the wrapped value is not valid, the Value method will return an error. Working around that might be more hassle than improving the Value enum upstream?

The way I did it in the PR was to match all possible subtypes in addition to the target type. However, we probably actually do not need to do this as I cannot think of any use cases other than u64.

EricLBuehler:
@ShelbyJenkins thank you! 

> Not sure if it's an error in [this quant,](https://huggingface.co/MaziyarPanahi/Meta-Llama-3-8B-Instruct-GGUF/tree/main?show_file_info=Meta-Llama-3-8B-Instruct.IQ1_M.gguf) but it's listed as tokenizer.ggml.model: gpt2.

>
> Given how prolific this hugging face user is I imagine people are going to try their quants and run into the same issue I had (only support for llama). Should I reach out to them or is it something that will be handled by this library?

We do the conversion in this codebase, so I can add support for the `gpt2` tokenizer type. However, I don't think it is a mislabeling because the other fields for GPT2 are there. I'll try to add this soon.

EricLBuehler:
@polarathene I took a look at it, and I think that the GGUF one is actually wrong here. It is not prepending a BOS token where one should not be. Regardless, GGUF tokenizers are documented to be less accurate than the HF ones and the GGML docs themselves recommend using the HF one.

I have not seen any effect of this error, as the chat template adds that anyway. This can, however, be an issue if you are not using a model with a chat template.

I also did a cursory search for an API through `tokenizers` to enable this prepending behavior but could not find any. Could you perhaps find what we are missing? **A key difference is that we use a `unigram` tokenizer here (that is what the GGUF tokenzier provides for `llama` tokenizers), but the `tokenizer.json` which we are testing against uses a `BPE` tokenizer.**

polarathene:
> Could you perhaps find what we are missing?

I'm not quite sure how to? Is there no information in the GGUF to ensure the correct tokenizer is used?

Even with the improved tokenizer support after unigram, is the GGUF still lacking the information to use the appropriate tokenizer?

---

> As another note, I plan on merging this before #397 (which #388 will need for testing) so I am excited to merge!

Sorry for the delay, I'll get onto applying the requested changes.

I'm still not sure how to proceed with the test case issue though.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -19,115 +20,71 @@ pub struct ConversionResult {
...
+    added_tokens: Option<Vec<String>>,
+    scores: Option<Vec<f32>>,
+    merges: Option<Vec<String>>,
+    unk: Option<u32>,
+    eos: u32,
+    bos: u32,
+}
+
+// This approach is a workaround for candles GGUF `Value` enum type wrapper,
+// a better upstream approach would be to have serialize/deserialize support.
```
polarathene, line 35:
```suggestion
```

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -197,50 +200,61 @@ mod tests {
...
-        let gguf_decoded = gguf_tokenizer
-            .decode(gguf_tokenized.get_ids(), true)
-            .map_err(anyhow::Error::msg)?;
+        // With special tokens added
+        // SKIPPED:
+        // - Bugged the GGUF tokenizer does not prepend `<s> `
+        // - Due to HF tokenizer using BPE (tokenizer.json) while GGUF tokenizer uses Unigram (metadata)?
+        /*
+        let hf_decoded = codec_roundtrip(&hf_tokenizer, passage.as_str(), true)?;
+        let gguf_decoded = codec_roundtrip(&gguf_tokenizer, passage.as_str(), true)?;
```
polarathene, line 254: Test is now skipped as requested.

mistralrs-core/src/utils/gguf_metadata.rs
```diff
@@ -0,0 +1,125 @@
...
+        let actual_arch: String = self
+            .metadata
+            .get("general.architecture")
+            .cloned()
+            .try_value_into()?;
+
+        anyhow::ensure!(
+            actual_arch == expected_arch,
+            "Expected `{expected_arch}` architecture, got `{actual_arch}`."
+        );
```
polarathene, line 54: Your `verify_sanity_gguf()` method was migrated to this one.

mistralrs-core/src/models/quantized_phi3.rs
```diff
@@ -160,28 +159,78 @@ fn precomput_freqs_cis(
...
+            "attention.head_count",
+            "attention.head_count_kv",
+            "block_count",
+            "embedding_length",
+            "feed_forward_length",
+            "rope.dimension_count",
+            "attention.layer_norm_rms_epsilon",
+            "context_length",
+        ];
+        c.has_required_keys(&required)?;
```
polarathene, line 192: I'm not sure how useful this is. Feel free to drop it, it's only purpose was to fail early with all non-optional keys logged if not found, rather than one at a time (_probably not a real concern to warrant this UX?_).

mistralrs-core/src/models/quantized_phi2.rs
```diff
@@ -143,27 +143,76 @@ fn layer_norm(w: QTensor, b: QTensor, eps: f64) -> Result<LayerNorm> {
...
+            head_count: c.get_value::<u32>("attention.head_count")? as usize,
+            head_count_kv: c.get_value::<u32>("attention.head_count_kv")? as usize,
+            block_count: c.get_value::<u32>("block_count")? as usize,
+            embedding_length: c.get_value::<u32>("embedding_length")? as usize,
+            rope_dim: c.get_value::<u32>("rope.dimension_count")? as usize,
+            ln_eps: c.get_value::<f32>("attention.layer_norm_rms_epsilon")? as f64,
+            max_seq_len: c
+                .get_value::<u64>("context_length")
+                .ok()
+                .unwrap_or(MAX_SEQ_LEN as u64) as usize,
```
polarathene, line 188: This PR did not carry of your upcast, but I assume for `u64` that'll be available upstream from `candle` at least, so I'm not sure if this regressed in support.

I noticed from the spec that other values are also u64 with v3, so these `PropsGGUF` may need to be  updated. I've not change their expected type or names (_`layer_norm_rms_epsilon` has several different struct names it gets assigned to depending on model_).

polarathene (commented):
Added context.

The additional commits pushed provide some more in the commit messages when relevant.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -136,6 +93,54 @@ pub fn convert_ggml_to_hf_tokenizer(content: &Content) -> Result<ConversionResul
...
+    tokenizer.with_decoder(decoders::sequence::Sequence::new(vec![
+        DecoderWrapper::Replace(Replace::new("▁", " ").map_err(anyhow::Error::msg)?),
+        DecoderWrapper::ByteFallback(ByteFallback::new()),
+        DecoderWrapper::Fuse(Fuse::new()),
+        DecoderWrapper::Strip(Strip::new(' ', 1, 0)),
+    ]));
+    tokenizer.with_normalizer(normalizers::Sequence::new(vec![
+        NormalizerWrapper::Prepend(Prepend::new("▁".to_string())),
+        NormalizerWrapper::Replace(Replace::new(" ", "▁").map_err(anyhow::Error::msg)?),
+    ]));
```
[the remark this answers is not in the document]
EricLBuehler: Yes, if you could add that it would be great.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
EricLBuehler: Great, thanks. Hopefully we can confirm this in the future.

mistralrs-core/src/utils/gguf_metadata.rs
EricLBuehler: Sounds great.

mistralrs-core/src/models/quantized_phi3.rs
EricLBuehler: No, I like that. In the case of an incompatible GGUF file somehow, it would be better to immediately tell the user what is missing rather than forcing them to run 10 times.

mistralrs-core/src/models/quantized_phi2.rs
EricLBuehler: Yes, I think we should always use u64 in PropsGGUF (we cast to usize anyway) because the Candle upcasting should cover it. I added that code before the upcast PR was merged but never removed it.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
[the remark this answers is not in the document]
EricLBuehler: Ah, ok. Sounds good.

EricLBuehler (commented):
@polarathene thank you for the comments. I think this is basically ready to merge. I think the change you suggested with the DecoderWrapper/Normalizer would be helpful for readability: so if you could add it, that would be great.

polarathene:
One of the tests are failing. I assume it's temporary and may have been from a rate limit.

```
---- sampler::tests::test_gumbel_speculative stdout ----
thread 'sampler::tests::test_gumbel_speculative' panicked at mistralrs-core\src\sampler.rs:423:60:
called `Result::unwrap()` on an `Err` value: IoError(Os { code: 5, kind: PermissionDenied, message: "Access is denied." })
```

https://github.com/EricLBuehler/mistral.rs/blob/69f86268f70f83ddb99a77e01a8cab0c298faff7/mistralrs-core/src/sampler.rs#L445-L461

https://github.com/EricLBuehler/mistral.rs/blob/69f86268f70f83ddb99a77e01a8cab0c298faff7/mistralrs-core/src/sampler.rs#L415-L425

I'm not familiar with what the rate limit might be (_total requests within a duration, total bandwidth within a duration, etc_), I guess it'd be optimal to have any models pulled as very small ones that are able to meet the needs of the test, even if they're not suitable for real world usage (_perhaps custom trained model with considerably less parameters to keep file size down?_).

Provided the model doesn't change much, you could easily leverage `actions/cache` in the CI to restore it locally to avoid making an external request each CI run?

---

> so if you could add it, that would be great.

Alright, will add that and push shortly 👍 

> I think we should always use u64 in PropsGGUF (we cast to usize anyway)

I'll leave as-is with this PR, I prefer those kind of changes to be a bit more isolated for `git blame` traceability without all the added noise from unrelated changes.

I could open a TODO issue for that if you'd like, since you're going to update this file next. Or leave with the TODO comment reminders I have in the PR 😅

EricLBuehler:
> Provided the model doesn't change much, you could easily leverage actions/cache in the CI to restore it locally to avoid making an external request each CI run?

Thanks, I'll look into using a cache. The Windows test is a bit flaky for this reason.

> I could open a TODO issue for that if you'd like, since you're going to update this file next. Or leave with the TODO comment reminders I have in the PR 😅

Sounds good. I'll probably resolve it quickly, as I will bundle the changes in the next PR.

polarathene:
I am not sure about this new test failure.

It's with the decoder again, but since your calling a non-deterministic input from a REST API web request, it's a bit tricky to know when the test is actually passing correctly?

Perhaps my change to the workaround types has a flaw :(

---

**UPDATE:**

I can see that every space character is being replaced with ~~`_`~~ (similar but not same, actually this: `▁`) in the GGUF tokenizer (_there is almost 16k of these_), but it looks like there is other differences too.

After converting those to spaces, it looks to be a similar issue. The GGUF tokenizer was prepending `▁` to the string.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -136,6 +91,160 @@ pub fn convert_ggml_to_hf_tokenizer(content: &Content) -> Result<ConversionResul
...
+    let normalizer = Normalizer::Sequence(vec![
+        Normalizer::Prepend("▁"),
+        Normalizer::Replace(" ", "▁"),
+    ]);
+
+    let mut tokenizer: Tokenizer = TokenizerX::try_builder()
+        .with_model(model)
+        .with_decoder(decoder)
+        .with_normalizer(normalizer)
+        .build()?;
```
polarathene, line 138: This is the gist of the recent commits pushed. I tried to use the upstream `TokenizerBuilder` but it was problematic.

I've opened an issue for that upstream, but in the meantime have implemented `TokenizerX` struct with `buildstructor` to a similar builder API as upstream without the issues.

---

Is there a reference for this decoder + normalizer config for LLaMa?

I found several github discussions about the quirks with the [`Prepend` + `Replace` requirement](https://github.com/ggerganov/llama.cpp/pull/2398#issuecomment-1718177991) (_looks like `_` was the special character but it's not_), and the `Strip` is presumably to [remove the initial space inserted from the encode](https://github.com/ggerganov/llama.cpp/issues/167#issuecomment-1473369760)?:

> _replace occurrences of `▁` (Lower One Eighth Block) in a newly allocated string._

> _`SentencePiece` always encodes first token with whitespace even if you ask to prepend `<bos>` token._

I assume there is a spec somewhere to reference which could be added as a comment for additional context?

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -136,6 +91,160 @@ pub fn convert_ggml_to_hf_tokenizer(content: &Content) -> Result<ConversionResul
...
+            tokenizer.with_decoder(d);
+        }
+        if let Some(normalizer) = with_normalizer {
+            let n = NormalizerWrapper::try_from(normalizer)?;
+            tokenizer.with_normalizer(n);
+        }
+
+        Ok(tokenizer)
+    }
+}
```
polarathene, line 178: Just commenting since the diff is collapsed, this is the builder workaround.

Just add equivalent builder methods when needed, should be pretty straight-forward.

Similar for the local enum `Decoder` + `Normalizer` definitions below if you need the other upstream variants 👍

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -197,50 +304,61 @@ mod tests {
...
-        let gguf_decoded = gguf_tokenizer
-            .decode(gguf_tokenized.get_ids(), true)
-            .map_err(anyhow::Error::msg)?;
+        // With special tokens added
+        // SKIPPED:
+        // - Bugged the GGUF tokenizer does not prepend `<s> `
+        // - Due to HF tokenizer using BPE (tokenizer.json) while GGUF tokenizer uses Unigram (metadata)?
+        /*
+        let hf_decoded = codec_roundtrip(&hf_tokenizer, passage.as_str(), true)?;
+        let gguf_decoded = codec_roundtrip(&gguf_tokenizer, passage.as_str(), true)?;
```
polarathene, line 358: Looking at the [GGUF metadata spec for the tokenizer props](https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#ggml), `llama` is apparently from `tokenizer.model`, and `gpt2` for `tokenizer.json`.

I assume the test can be fixed by selecting a GGUF model that has `gpt` used instead, or a HF model that provides `tokenizer.model` (like that Hermes mistral model).

Many of the originals on HF require an account though. `Phi3` doesn't, it has both `tokenizer.model` and `tokenizer.json`, I haven't checked what the GGUF metadata is:
- Base: https://huggingface.co/microsoft/Phi-3-mini-4k-instruct
- GGUF: https://huggingface.co/microsoft/Phi-3-mini-4k-instruct-gguf

polarathene (commented):
Context for recent commits.

Should be good to merge now?

mistralrs-core/src/pipeline/gguf_tokenizer.rs
EricLBuehler: I built that section of code by following a `tokenizer.json` for an HF sentencepiece model. I am not sure that there is a reference for these configs, but you can see that [Google's implementation](https://github.com/google/sentencepiece?tab=readme-ov-file#subword-regularization-and-bpe-dropout) also prepends this `_`.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
EricLBuehler: Great, thanks!

mistralrs-core/src/pipeline/gguf_tokenizer.rs
EricLBuehler: Hmm, perhaps we can switch to phi3 later? I think it is likely that this is the reason the test is broken, that the tokenizer type is `llama` whereas the equivalent is `gpt2.`

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -197,50 +304,61 @@ mod tests {
         }
     }
 
-    #[allow(dead_code)]
     fn get_test_passage() -> String {
+        // TODO: Why is it necessary to depend on this for a multi-line test string?
```
EricLBuehler, line 308: Perhaps we can hardcode this? It should be easy.

EricLBuehler (commented):
Thank you for the changes and comments! Can you please hardcode the test passage so that we have a consistent input? It does not need to be excessively long, just enough to test the capabilities of the tokenizer.

polarathene:
> Can you please hardcode the test passage so that we have a consistent input?

Sure.

I haven't gone over the contents of [this `llama.cpp` tokenizer test](https://github.com/ggerganov/llama.cpp/blob/7027b27d765db95d4ac6b569d976e387a8715881/tests/test-tokenizer-random.py#L99-L161), but it's probably a good reference (_lists some edge cases they encountered too where some models fail_).

For now I'll just go with a single input based on some content from there.

mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -197,13 +304,35 @@ mod tests {
...
-        let passage = reqwest::blocking::get("https://loripsum.net/api")
-            .expect("Failed to download sample text")
-            .bytes()
-            .expect("Failed to get bytes");
-        String::from_utf8(passage.to_vec()).expect("Failed to convert sample text to string.")
+        let passage = "Hello, world! \n🚀 (normal) 😶‍🌫️ (compound emoji, zwj sequence) ✅ (emoji as single token)\n你好世界！\nNǐ hǎo shìjiè!";
+
+        passage.to_owned()
+    }
+
```
polarathene, line 313: I think the emoji one is a bit interesting for test coverage. Especially the compound/ZWJ type.

👨‍👩‍👧‍👦 for example is 25 bytes, trimming some from the start or end can either produce a different emoji or corrupt output.

polarathene:
@EricLBuehler [your recent merge](https://github.com/EricLBuehler/mistral.rs/commit/5a7ebb7402d50e033fb53abc9651df0472ab909d#diff-b649f8fb6ba89c102b7e4b60a080d446da76b80428618c168a623799eb3aa6b2) was based off your earlier PR work that you since reverted, you merged that and now there's a bunch of conflicts for changes unrelated to Phi3 vision 😓 

I'm not even able to resolve those via the web UI this time apparently (_not that I particularly would want to_):

> image

Potentially, revert the merge, and fix the history there? 

[That PR was 184 commits long](https://github.com/EricLBuehler/mistral.rs/pull/351) though, it looks there were other changes like dropping some docs (Adding models, CLI docs), so I don't know what else was done that was/wasn't intentional 🤷‍♂️

EricLBuehler:
Ok, sounds good. Is this PR ready for merge?

polarathene:
> Is this PR ready for merge?

Yes?

You asked for the test to be updated and I addressed that roughly 11 hours ago. But now there is the conflicts.

EricLBuehler:
Ok, thanks for confirming. I'll undo the merge of that PR so that this can merge cleanly. Sorry for the inconvenience!

mistralrs-core/src/pipeline/gguf_tokenizer.rs
EricLBuehler: Nice!

EricLBuehler (approved):

EricLBuehler:
Thank you!

polarathene:
Just to confirm, was the removal of those mentioned doc pages intentional? Seemed like useful information, just a bit outdated?

EricLBuehler:
Sorry, which docs pages were removed?

polarathene:
> which docs pages were removed?

I referenced them at the end of the [comment above](https://github.com/EricLBuehler/mistral.rs/pull/389#issuecomment-2155630358). "Adding Models" / "CLI".

I then said if that was unintentional that there may have been other mishaps, but there was quite a bit of changes in the PR you merged.

EricLBuehler:
Oh I see. I removed the "Adding models" in favor of a comment in the README contributing section because of the rate of development regarding that section, and the "CLI" document was prone to being outdated. If you know of a better way to autogenerate the CLI docs that would be great, I think we *could* do it in CI.

polarathene:
> If you know of a better way to autogenerate the CLI docs that would be great, I think we _could_ do it in CI.

Yeah, you could probably have a template file to produce the markdown content? Or switch to a docs generator that can import content from separate sources.

Then for the CI part, you would just get the output of the CLI help and if it doesn't match the importable snippet, it'd update that and generate the page again.

Another way with CI is to have that sort of behaviour triggered on an automated schedule rather than commit triggers. That can then create a PR with the update, I've not done that before myself but I'm aware of something like that 🤔

EricLBuehler:
Ok, I'll look into that. But before starting work on this, do you see static CLI docs as a useful tool when users can simply download and run with --help? That was my rationale.

polarathene:
> do you see static CLI docs as a useful tool when users can simply download and run with --help? That was my rationale.

It was useful as an insight when I first came across `mistral.rs` and was looking through the README. I think that was more of an issue with the README at the time being a bit confusing with instructions, and I knew I'd have to build the CLI.

If there's no additional context though, it's probably not that meaningful :)

I had a look over the deleted document, it doesn't appear to provide any value beyond an overview to reference without the CLI available. **I wouldn't worry about restoring it** 👍

EricLBuehler:
Great. I did a refactor of the README yesterday, do you see any parts that are confusing or hard to find?

PR #397, opened 2024-06-05
Implement gpt2 (BPE) GGUF tokenizer conversion

EricLBuehler:


mistralrs-core/src/pipeline/gguf_tokenizer.rs
```diff
@@ -113,6 +113,34 @@ pub fn convert_ggml_to_hf_tokenizer(content: &Content) -> Result<ConversionResul
...
+
+            let bpe = BpeBuilder::new()
+                .vocab_and_merges(vocab, merges)
+                .build()
+                .map_err(anyhow::Error::msg)?;
+            let mut tokenizer = Tokenizer::new(ModelWrapper::BPE(bpe));
+            tokenizer.with_decoder(decoders::byte_level::ByteLevel::new(true, true, true));
+
+            tokenizer.add_special_tokens(&[AddedToken::from(tokens[bos as usize].clone(), true)]);
+            tokenizer.add_special_tokens(&[AddedToken::from(tokens[eos as usize].clone(), true)]);
```
polarathene, line 140: Just curious, BPE has support for [setting `unk` in it's builder variant](https://docs.rs/tokenizers/latest/tokenizers/models/bpe/struct.BpeBuilder.html#method.unk_token), is it not relevant for some reason? (_I know very little about these things_)

EricLBuehler: @polarathene the GGUF file I am testing with (QuantFactory/Meta-Llama-3-8B-Instruct-GGUF) does not have a unk metadata, so I left it out here.

polarathene: Sure, but what about when they do? I assume that's possible since the tokenizer builder for BPE does support setting `unk`? There is no check for this, so if there was it'd just ignore it and introduce a bug?

EricLBuehler: Yes, I wasn't sure if this is guaranteed to be not provided, but just in case I added [8d4dba5](https://github.com/EricLBuehler/mistral.rs/pull/397/commits/8d4dba52462c9e8002767c654946e3f043a2872e) and [763241e](https://github.com/EricLBuehler/mistral.rs/pull/397/commits/763241e16ad10379eff8a4e9f03bcf48ad2fd85e).

mistralrs-core/src/gguf/gguf_tokenizer.rs
```diff
@@ -140,15 +183,43 @@ fn unigram_tokenizer(p: &PropsGGUF) -> Result<(Tokenizer, TokenizerKind, Vec<Str
...
+        .merges
+        .as_ref()
+        .ok_or(anyhow::Error::msg("BPE tokenizer must include merges"))?
+        .iter()
+        .map(|merge| {
+            let split: (&str, &str) = merge
+                .splitn(2, ' ')
+                .collect_tuple()
+                .expect("Failed to convert split into 2-tuple");
+            (split.0.to_string(), split.1.to_string())
```
polarathene, line 202: This splits a string like "I like rust" into `("I", "like rust")`?

I haven't seen any examples where you'd have space in the input, but I assume this is referencing an existing implementation somewhere already 😅 (_not doubting your work, just curious_)

I came across [this article](https://martinlwx.github.io/en/the-bpe-tokenizer/) that mentions white-space splitting at the end, noting it is not suitable for languages like chinese.

mistralrs-core/src/gguf/gguf_tokenizer.rs
```diff
@@ -140,15 +183,43 @@ fn unigram_tokenizer(p: &PropsGGUF) -> Result<(Tokenizer, TokenizerKind, Vec<Str
...
+        vocab.insert(token.clone(), i as u32);
     }
 
-    Ok((tokenizer, TokenizerKind::Unigram, special_tokens))
+    let bpe = BpeBuilder::new()
+        .vocab_and_merges(vocab, merges)
+        .build()
+        .map_err(anyhow::Error::msg)?;
+    let mut tokenizer = Tokenizer::new(ModelWrapper::BPE(bpe));
+    tokenizer.with_decoder(decoders::byte_level::ByteLevel::new(true, true, true));
```
polarathene, line 221: While not necessary, this could be handled with the `TokenizerX::builder()`, which you could then call `.with_model(bpe)`.

The `ByteLevel` decoder would need to be added to the decoder enum below with:

```rs
enum Decoder<'a> {
    ByteLevel(bool, bool, bool),
```

```rs
    fn try_from(variant: Decoder) -> Result<Self, Self::Error> {
        let value: DecoderWrapper = match variant {
            Decoder::ByteLevel(add_prefix_space, trim_offsets, use_regex) => {
                ByteLevel::new(add_prefix_space, trim_offsets, use_regex).into()
            }
```

Then you would have:

```suggestion
    let mut tokenizer = TokenizerX::builder()
        .with_model(bpe);
        .with_decoder(Decoder::ByteLevel(true, true, true));
```

That would make it a bit more aligned with the unigram construction.

mistralrs-core/src/gguf/gguf_tokenizer.rs
```diff
@@ -79,26 +87,61 @@ pub fn convert_gguf_to_hf_tokenizer(content: &Content) -> Result<GgufTokenizerCo
...
+            eos: special_tokens[1].clone(),
+            unk: Some(special_tokens[2].clone()),
+        }
+    } else {
+        AddedTokensCollection {
+            bos: special_tokens[0].clone(),
+            eos: special_tokens[1].clone(),
+            unk: None,
+        }
+    }
```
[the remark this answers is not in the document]
polarathene: Actually we could also get these tokens via [`tokenizer.id_to_token(u32)`](https://docs.rs/tokenizers/latest/tokenizers/tokenizer/struct.Tokenizer.html#method.id_to_token) in the main function body where they're needed. That'll provide the expected `Option<String>` for each already for us 👍

Just a little tricky if you have default u32 values changing like for `unk` in Unigram. `&mut PropsGGUF` could technically accommodate though.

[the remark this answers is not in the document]
EricLBuehler: They are not optional here, only `unk`. I do like your method of adding the special tokens though, I have added that.

mistralrs-core/src/gguf/gguf_tokenizer.rs
EricLBuehler: No problem, happy to explain. When you have a `merges` field, from all the examples I have looked at, they are encoded together, separated by a space because the BPE tokenizer inserts spaces, so there is no space token. The key point is that each element of the table is a merge *pair*. This code would split the merge representation `"he llo"` into the pair `("he", "llo")`.

I looked at the article you linked, and it also mentioned the pair construction of merges.

mistralrs-core/src/gguf/gguf_tokenizer.rs
EricLBuehler: Thanks for the idea, I have added this.

mistralrs-core/src/gguf/gguf_tokenizer.rs
```diff
@@ -140,15 +179,49 @@ fn unigram_tokenizer(p: &PropsGGUF) -> Result<(Tokenizer, TokenizerKind, Vec<Str
...
-        let token = p.tokens[token_id as usize].as_str();
+    let special_tokens = add_special_tokens(p, &mut tokenizer, bos, eos, Some(unk));
 
-        special_tokens.push(token.to_owned());
-        tokenizer.add_special_tokens(&[AddedToken::from(token.to_owned(), true)]);
+    Ok((tokenizer, TokenizerKind::Unigram, special_tokens))
+}
+
+fn bpe_tokenizer(p: &PropsGGUF) -> Result<(Tokenizer, TokenizerKind, AddedTokensCollection)> {
+    let merges = p
```
polarathene, line 188: Some context to add if you think it might be helpful for those like myself that aren't as familiar but wanted to understand what was going on here 😅 

```suggestion
    // BPE merges have each string item as a space-delimited pair:
    // https://github.com/EricLBuehler/mistral.rs/pull/397#discussion_r1631988370
    let merges = p
```

EricLBuehler: Yes, thank you!

PR #401, opened 2024-06-07
docs: Resolve CI lints on docs

polarathene:
I'm not familiar with the syntax, I've assumed markdown.

This PR should address the issues the action keeps annotating about unrelated to actual PR changes.

### DIFF
--- a/mistralrs-core/src/pipeline/mod.rs
+++ b/mistralrs-core/src/pipeline/mod.rs
@@ -1,5 +1,5 @@
 mod cache_manager;
-mod chat_template;
+pub mod chat_template;
 mod ggml;
 mod gguf;
 mod gguf_tokenizer;
@@ -56,12 +56,14 @@ pub trait ModelPaths {
     /// Model weights files (multiple files supported).
     fn get_weight_filenames(&self) -> &[PathBuf];
 
-    /// Retrieve the PretrainedConfig file.
-    /// See: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/configuration#transformers.PretrainedConfig
+    /// Retrieve the [`PretrainedConfig`] file.
+    ///
+    /// [`PretrainedConfig`]: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/configuration#transformers.PretrainedConfig
     fn get_config_filename(&self) -> &PathBuf;
 
-    /// A serialised `tokenizers.Tokenizer` HuggingFace object.
-    /// See: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/tokenizer
+    /// A serialised [`tokenizers.Tokenizer`] HuggingFace object.
+    ///
+    /// [`tokenizers.Tokenizer`]: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/tokenizer
     fn get_tokenizer_filename(&self) -> &PathBuf;
 
     /// Content expected to deserialize to [`ChatTemplate`].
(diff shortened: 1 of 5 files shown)


mistralrs-core/src/pipeline/mod.rs
```diff
@@ -18,7 +18,7 @@ mod sampling_pipeline;
 use crate::lora::{LoraConfig, Ordering};
 use crate::DeviceMapMetadata;
 use candle_core::quantized::GgmlDType;
-use chat_template::ChatTemplate;
+pub use chat_template::ChatTemplate;
```
polarathene, line 21: Not sure if this should be public, but seems like might have been required for the docs reference to it. The struct itself is annotated as `pub` and not `pub(crate)` so I think that's ok?

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -56,12 +56,12 @@ pub trait ModelPaths {
     /// Model weights files (multiple files supported).
     fn get_weight_filenames(&self) -> &[PathBuf];
 
-    /// Retrieve the PretrainedConfig file.
-    /// See: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/configuration#transformers.PretrainedConfig
+    /// Retrieve the [PretrainedConfig] file.
+    /// [PretrainedConfig]: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/configuration#transformers.PretrainedConfig
```
polarathene, line 60: Should this be backtick wrapped like other values?

```suggestion
    /// Retrieve the [`PretrainedConfig`] file.
    /// [`PretrainedConfig`]: https://huggingface.co/docs/transformers/v4.40.2/en/main_classes/configuration#transformers.PretrainedConfig
```

mistralrs-core/src/pipeline/speculative.rs
```diff
@@ -117,7 +117,7 @@ impl Loader for SpeculativeLoader {
     }
 }
 
-/// Speculative decoding pipeline: https://arxiv.org/pdf/2211.17192
+/// Speculative decoding pipeline: <https://arxiv.org/pdf/2211.17192>
```
polarathene, line 120: Presumably this will render as a raw URL.

mistralrs-server/src/main.rs
```diff
@@ -89,8 +89,8 @@ struct Args {
     chat_template: Option<String>,
 
     /// Source of the token for authentication.
-    /// Can be in the formats: "literal:<value>", "env:<value>", "path:<value>", "cache" to use a cached token or "none" to use no token.
-    /// Defaults to using a cached token.
+    /// Can be in the formats: `literal:<value>`, `env:<value>`, `path:<value>`, `cache` to use a cached token, or `none` to use no token.
+    /// Defaults to `cache`.
```
polarathene, line 93: Switched to backticks instead of double quote wrapping as the CI check considered `<value>` as HTML tag 🤷‍♂️ 

This isn't actually used to generate docs for the crate, only for CLI help. I noticed `mistralrs-server` is locked down heavily on the docs generated, and assume there is a way to exclude it from showing in the generated docs if there's no intended lib usage via the crate.

polarathene (commented):
Context

polarathene (commented):
Context

mistralrs-core/src/pipeline/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod cache_manager;
-mod chat_template;
+pub mod chat_template;
```
polarathene, line 2: Not sure if this module should be made public, but seems like might have been required for the docs reference to it?

EricLBuehler: Yeah, makes sense. I actually want to refactor the macros to remove the `#[macro_export]` usage later.

EricLBuehler (approved):

PR #412, opened 2024-06-09
Remove causal masks cache

EricLBuehler:


gregszumel:
Looks good to me, I cannot reproduce the problem on this PR.

For context - I observed that memory was not being fully freed in inference, so a long request might take up a bunch of memory and not properly release it, even when shorter requests come through. This fixes that -- thanks @EricLBuehler!

EricLBuehler:
@gregszumel thank you for confirming!

PR #421, opened 2024-06-11
Switch to minijinja's pycompat mode

mitsuhiko:
This switches to MiniJinja's new pycompat code to avoid the manual string manipulation.

See https://docs.rs/minijinja-contrib/latest/minijinja_contrib/pycompat/fn.unknown_method_callback.html


EricLBuehler (approved):
Thank you!

PR #426, opened 2024-06-12
Fix Python deps, base64 impl, add examples

EricLBuehler:
- Add examples of using base64 encoded inputs
- Add credits to READMEs


EricLBuehler:
Note: this PR bumped the versionto 0.1.18 as it introduces a build patch

PR #446, opened 2024-06-18
Fix clippy shadowing

EricLBuehler:
Should reduce memory usage and hopefully increase speed...


EricLBuehler:
@chenwanqq I cannot measure any T/s speedup. I think that because we are using a GPU and the commands are async, the Rust drop code runs quickly enough that there is no difference.

I implemented this change only for the `models/quantized_llama.rs` code.

chenwanqq:
> @chenwanqq I cannot measure any T/s speedup. I think that because we are using a GPU and the commands are async, the Rust drop code runs quickly enough that there is no difference.
> 
> I implemented this change only for the `models/quantized_llama.rs` code.

I think the problem might not be about speed, but about peak memory usage.🧐 For instance, whether it can run a model within limited memory space or how many tokens it can process for a given model.

Issue #395, opened 2024-06-05
Cross GPU device mapping feature


joshpopelka20:
I'm working with a long context model (gradientai/Llama-3-8B-Instruct-262k) that exceeds the memory of a single A100 GPU. While the model weights are loaded, when I try to run inference, I get CUDA Out of Memory exception.  

Requesting a new feature to allow users to use cross GPU device mapping.

b0xtch:
Related issue: https://github.com/huggingface/candle/issues/2007

There was an attempt to do tensor parallelism:

https://github.com/EricLBuehler/mistral.rs/pull/72

EricLBuehler:
Hi @joshpopelka20 and @b0xtch! I just merged #462 which adds cross-GPU device mapping support (including for Python). I plan on implementing tensor parallelism, too, in the future.

joshpopelka20:
@EricLBuehler thanks for adding this feature. 

When using the pypi package, I'm getting this error:
```
        = note: /usr/bin/ld: /tmp/pip-install-8y1wtzkm/mistralrs-cuda_36180abfef7b4f0687d7842e9b298d9e/target/release/build/mistralrs-core-f5a6ed7a31cbdb62/out/libmistralcuda.a(nonzero_bitwise-b50867152df76f01.o): relocation R_X86_64_32 against symbol `_Z17transform_indicesPKjjS0_jPj' can not be used when making a shared object; recompile with -fPIC
                /usr/bin/ld: final link failed: Nonrepresentable section on output
                collect2: error: ld returned 1 exit status
```

I'm passing the CUDA_NVCC_FLAGS flag so not sure why it's saying to "recompile with -fPIC". These are the commands I'm using:

> env["CUDA_NVCC_FLAGS"] = "-fPIE"
> result = subprocess.run(['pip', 'install', 'mistralrs-cuda'], env=env)

EricLBuehler:
> I'm passing the CUDA_NVCC_FLAGS flag so not sure why it's saying to "recompile with -fPIC". These are the commands I'm using:

Can you try:
```
env["CUDA_NVCC_FLAGS"] = "-fPIC"
result = subprocess.run(['pip', 'install', 'mistralrs-cuda'], env=env)
```
The -fPIC requirement may stem from your Linux distribution (some require -fPIE, I'll add this to the README)?

joshpopelka20gmail:
[...]

PR #462, opened 2024-06-22
Add support for cross-gpu device mapping

EricLBuehler:
Refs #395.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Mistal.rs supports several model categories:
 - Quantized model support: 2-bit, 3-bit, 4-bit, 5-bit, 6-bit and 8-bit for faster inference and optimized memory usage.
 - Continuous batching.
 - Prefix caching.
-- Device mapping: load and run some layers on the device and the rest on the CPU.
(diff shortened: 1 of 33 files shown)


chenwanqq:
https://github.com/EricLBuehler/mistral.rs/blob/e04f8400c0a5b3c5d8a2a68912b273cc973b53e8/mistralrs-core/src/models/llama.rs#L302-L303
I think maybe this line copy the entire weights..
It does cost serious memory usage problem, can not even start a 7b llama model in 4090 GPU...

EricLBuehler:
@chenwanqq can you please raise an issue about this?

chenwanqq:
> @chenwanqq can you please raise an issue about this?

The problem is that I can not track this problem again.. after using memory usage to check:
```rust
println!("vb.device: {:?}", vb.device());
        println!("memory usage: {:?}", memory_track.get_memory_available(vb.device()));
        let vb = vb.set_dtype(mapper.get_min_dtype()?);
        println!("memory usage: {:?}", memory_track.get_memory_available(vb.device()));
```
It is same before and after. And I did not encounter oom problem today...

Issue #448, opened 2024-06-18
Accelerate quantization with Marlin kernel or HQQ


EricLBuehler:
The [Marlin INT4xFP16](https://github.com/IST-DASLab/marlin) CUDA matmul kernel can achieve ~4x speed improvement over CUTLASS matmul.

See also: [hqq](https://github.com/mobiusml/hqq/) as a quantization method which supports Marlin and other optimized kernels, without calibration data, and quantization in minutes.

EricLBuehler:
Closing as this has been implemented!

Issue #418, opened 2024-06-11
Is there any plans to support AWQ and GPTQ in the future?


franklucky001:
support quantization of awq and gptq

EricLBuehler:
Hi @franklucky001! I do plan on supporting other quantization method such as AWQ and GPTQ in the future. I would probably start with GPTQ.

EricLBuehler:
@franklucky001 we are going to merge GPTQ soon!

EricLBuehler:
@franklucky001 GPTQ and Marlin support have been merged already.

BuildBackBuehler:
So does AWQ work? For some reason I could've sworn it was a form of GPTQ so I figure there's a possibility. Wish I had an AWQ model downloaded to try

PR #467, opened 2024-06-23
Implement GPTQ quantization

EricLBuehler:
This PR adds GPTQ quantization ([paper here](https://arxiv.org/abs/2210.17323)) support.

Refs: #418, #448.


EricLBuehler:
cargo run --features cuda -- -i plain -m kaitchup/Phi-3-mini-4k-instruct-gptq-4bit -a phi3

BuildBackBuehler:
Broke my heart when I went to try Mistral_Large 2-bit EQAT (AutoGPTQ) on my M1 and only then saw no Mac support 😭. Wondering when might that come around? If adding that support for MPS/Metal was not too high-level expert knowledge prereq'd of a task I wouldn't mind taking a swing at it 😂

EricLBuehler:
@BuildBackBuehler If you could add this, it would be amazing!

I haven't seen GPTQ kernels on Mac though, if you can find any it shouldn't be too hard to add it and I would appreciate it if you take a shot!

Issue #263, opened 2024-05-03
distribute with cargo dist?


ashleygwilliams:
love to see more rust in the AI space. i work on a tool called cargo-dist that can help package up pre-built binaries and build installers so it's easier for folks to get up and running with your project (without having to install rust or sys deps manually). let me know if you'd be interested at all, would be happy to work with you to get it set up! 

https://github.com/axodotdev/cargo-dist

EricLBuehler:
Hi @ashleygwilliams! I think that this would be a great addition. I think an installer would work best here due to the wide variety of `--features` and also the Python API. If you could submit a PR adding this, it would be great!

EricLBuehler:
Hello @ashleygwilliams! I am very interested in using cargo dist to improve ease of installation, have you made progress with this?

PR #480, opened 2024-06-25
release using cargo-dist

kranurag7:
Fixes: https://github.com/EricLBuehler/mistral.rs/issues/263

As of now, musl and windows based build are disabled because they were [breaking](https://github.com/kranurag7/mistral.rs/actions/runs/9666023904), I can take a look at fixing those later. 

I created a release on my fork here: https://github.com/kranurag7/mistral.rs/releases/tag/v0.1.24

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ installers = []
 targets = ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu"]
 # Publish jobs to run in CI
 pr-run-mode = "plan"
-# Whether to auto-include files like READMEs, LICENSEs, and CHANGELOGs (default true)
+# Whether to auto-include files like READMEs, LICENSE, and CHANGELOGs (default true)
 auto-includes = false
 
 [profile.profiling]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,29 @@ image = "0.25.1"
 reqwest = { version = "0.12.4", features = ["blocking"] }
 base64 = "0.22.1"
 
+# Config for 'cargo dist'
+[workspace.metadata.dist]
+# The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
+cargo-dist-version = "0.16.0"
+# CI backends to support
+ci = "github"
+# The installers to generate for each app
+installers = []
+# Target platforms to build apps for (Rust target-triple syntax)
+targets = ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu"]
+# Publish jobs to run in CI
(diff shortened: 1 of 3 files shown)


EricLBuehler:
Hi @kranurag7! Thanks for adding this. I noticed you added a few new CI steps: build-global-artifacts, host, announce. I was wondering if you could give a quick overview of what you added and generally how it works? In particular, we already use GH releases to create tags and track releases, so I'm not sure if that would be affected by the new `announce` step?

> As of now, musl and windows based build are disabled because they were [breaking](https://github.com/kranurag7/mistral.rs/actions/runs/9666023904), I can take a look at fixing those later.

Sounds good.

kranurag7:
The build jobs builds the artifacts and then we create a release in GitHub. This means you'll get your release assets uploaded and a table which will ease the downloading of tarballs for users. 

See https://github.com/kranurag7/mistral.rs/releases/tag/v0.1.24 

This is what is done in the last `announce` step. It as the last stage which gets executed after the build is finished. 

From what I see, you push a tag to build the image via two workflow. 
- [build_cpu.yaml](https://github.com/EricLBuehler/mistral.rs/blob/master/.github/workflows/build_cpu.yaml) and [build_cuda_all.yaml](https://github.com/EricLBuehler/mistral.rs/blob/master/.github/workflows/build_cuda_all.yaml)

This workflow extends that, now when you push the tag, you'll have your binaries present as well on the releases.

EricLBuehler (approved):
Thank you!

ashleygwilliams:
so glad to see this! sorry i dropped the ball on actually making the PR.

in case it's not clear, the tool `cargo-dist` fully generates the workflow. to update the workflow, you'll use the `cargo-dist` tool and edit config in your `Cargo.toml`. to update- you can run `cargo dist selfupdate` which will fetch the new version of `cargo-dist` and also run `cargo dist init` which will ask you some questions and re-gen your workflows. we do not recommend you hand edit the workflow and cargo-dist will complain if you do (you can tell it to stop, but again, we don't recommend that). feel free to ask me if you have any questions or hop in our discord, https://discord.gg/qzzYcWPh

Cargo.toml
```diff
@@ -40,9 +40,29 @@ image = "0.25.1"
...
 base64 = "0.22.1"
 
+# Config for 'cargo dist'
+[workspace.metadata.dist]
+# The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
+cargo-dist-version = "0.16.0"
+# CI backends to support
+ci = "github"
+# The installers to generate for each app
+installers = []
```
ashleygwilliams, line 50: for ease of installing, you may want to add some installers :) this will let people install the pre-built binaries instead of having to compile on their machine with `cargo install`

Issue #465, opened 2024-06-23
What are the differences between the llama block in this repo and the implementation in candle-transformer?


chenwanqq:
Hello, as we discussed before, I am integrating LLaVA into this repo, but I encountered an extremely difficult problem that is hard to debug.

In LLaVA, the image features (obtained through CLIP) and text features need to be concatenated and fed into the LLaMA model. However, during the processing of the llama block, all activation values became NaN. I used the following code for debugging:

```rust
fn forward(
        &self,
        x: &Tensor,
        attention_mask: &Option<Tensor>,
        seqlen_offsets: &[usize],
        start_offsets_kernel: Tensor,
        block_idx: usize,
        kv_cache: &mut crate::pipeline::LayerCaches,
    ) -> Result<Tensor> {
        let residual = x;
        let max_value = x.max(0)?.max(0)?.max(0)?;
        let min_value = x.min(0)?.min(0)?.min(0)?;
        println!("before rms_1 max_value: {} min_value: {}",max_value,min_value);
        let x = self.rms_1.forward(x)?;
        println!("after rms_1 x: {}",x);
        let x = (self.attn.forward(
       //....
```

I found that after passing through two blocks, the max_value and min_value became infinity and negative infinity, respectively, which caused subsequent issues.

Here are some key points:

(1) If I only input text, this problem does not occur.

(2) This problem does not occur in my implementation based on candle-transformers. I measured the maximum and minimum values of the features before entering the block, and both implementations are around plus or minus 30, which seems to make no difference.

After further reading the code, I found that the llama used in this repo adopts "fused_rope". I'm not sure what this is. Could you please explain the difference between this and the original version of rope?

chenwanqq:
By the way my working branch is [chenwanqq/mistral.rs](https://github.com/chenwanqq/mistral.rs/tree/llava). However it's a mess and there are tons of local path and println! for debugging.

EricLBuehler:
Hello @chenwanqq! Our llama block in this repo is not different than the candle-transformers one in design. The only difference is that we, as you pointed out, use the fused RoPE kernel instead of the Candle RoPE. Our fused RoPE is faster and does all the operations for multiple batches in one kernel.

Does this behavior occur after the first block runs (so in the second block) or at some other block? That could be a clue to which part of the code is not behaving correctly.

chenwanqq:
[...]

PR #476, opened 2024-06-24
AnyMoE: Build an MoE model from anything, quickly

EricLBuehler:
This PR implements AnyMoE, a method to build a flexible MoE model from any combination of fine-tuned expert models. Please see [the paper](https://arxiv.org/abs/2405.19076) for reference, although currently there are a few implementation-level differences.

The technique implemented here is similar to the gating method found in models such as [Cephalo Vision 3x8b beta](https://huggingface.co/lamm-mit/Cephalo-Idefics2-vision-3x8b-beta) or Mixtral: it uses a gating layer to select experts in the MLP layer. This PR implements a built-in pretraining strategy with the option to save the resulting gating layer weights, powered by Candle's autograd infrastructure.

By allowing users to create an MoE model quickly by using the expert weights and a small pretraining dataset, AnyMoE lowers the barrier to entry for users who want to create and deploy tailored MoE models for their application.

## Status
- [x] AnyMoEPipeline
  - [x] Add most of it
  - [x] Pretraining infrastructure
  - [x] Connect to the rest of mistral.rs
    - [x] Python API
    - [x] CLI
    - [x] Server

PR #484, opened 2024-06-27
Add LLaVA Support

chenwanqq:
## Introduction

This implementation is based on my work for [candle](https://github.com/huggingface/candle). However, it incorporates some notable differences:

* I have completely removed support for the model format used in the liuhaotian/llava (original) repo. Instead, I adopted the model format from the llava-hf repo, providing a more standardized approach to utilizing the model, particularly the tokenizer.
* Similar to the Python transformer library, I have segmented the original project into LLaVA (1.5) and LLaVANext (1.6). This reduction in the use of if-else statements simplifies the code.
* I have revamped the process of concatenating text and image features. The updated method aligns closely with phi3v's implementation in this repo, ensuring a more uniform coding style.

Still Working!

## Some Notes
* The fused RoPE might cause issues, possibly due to precision-related matters. Please refer to this https://github.com/EricLBuehler/mistral.rs/issues/465. Therefore, I have implemented an alternative version in the llava-llm folder.
~~* Certain modifications regarding cross-GPU mapping support [[link]](https://github.com/EricLBuehler/mistral.rs/pull/462#issuecomment-2194217523) might result in significant memory usage problems.~~ I can not reproduce this problem right now

## Status
- [x] Implement model
	- [x] LLaVANext (1.6)
	- [x] LLaVA 1.5
		- [x] Input processor
		- [x] Model structure
- [x] Support for different LLM Backends
	- [x] LLaMA-like models (vicuna, Nous-Hermes-2-Yi-34B)
	- [x] Mistral
- [x] Cleanup of some debug code
- [x] Documentation
- [x] Examples


EricLBuehler:
Hi @chenwanqq! Please let me know when this is ready for review.

chenwanqq:
> Hi @chenwanqq! Please let me know when this is ready for review.

Hhh I think it's ready. Maybe some docs and examples are a little bit hurry but I think we can review it🤣

EricLBuehler:
Great, I'll take a look. I think that the recent merge of #476 caused some merge conflicts here, can you please take a look?

chenwanqq:
> Great, I'll take a look. I think that the recent merge of #476 caused some merge conflicts here, can you please take a look?
> 
> Edit: Seems like they are trivial to fix.

The addition of AnyMOE (which modifies the dependencies of NormalModel) affects the LLaVA llm backend. However, the current solution based on copied code is just a temporary one, so I have removed the NormalModel trait from LLaVALLM altogether. Once the fusedRoPE is fixed, we should change LLaVA to rely on an ordinary model as soon as possible.

chenwanqq:
And LLaVANext is just LLaVA1.5 plus some multiple resolution trick(anyres). So they share some infrastuctres. If this causes any confusion, we can modify it.

mistralrs-core/src/vision_models/llava/llava15.rs
```diff
@@ -0,0 +1,333 @@
...
+        result = self.llm.embed(&result)?; //[seq_len,hidden_size]
+        let image_features = self.encode_images(&images.to_dtype(self.dtype)?)?; //[num of images,patch_size*patch_size,hidden_size]
+        let num_of_images = image_features.shape().dims()[0];
+        let mut image_features_vec = Vec::new();
+        for i in 0..num_of_images {
+            image_features_vec.push(image_features.get(i)?.unsqueeze(0)?);
+        }
+        for (i, image_index) in image_indexes.iter().enumerate() {
+            let image_id = image_ids[i];
+            result = result.slice_assign(
```
EricLBuehler, line 198: What were your thoughts on our slice assign changes on the Candle fork 😃? I have a Candle PR open.

mistralrs-core/src/vision_models/llava/llava15.rs
```diff
@@ -0,0 +1,333 @@
...
+                context_lens,
+                position_ids,
+            )
+        }
+    }
+}
+
+impl IsqModel for Model {
+    fn get_tensors(&mut self) -> (Vec<(&mut QMatMul, Option<usize>)>, &dyn DeviceMapper) {
+        // I don't really get this part
```
EricLBuehler, line 253: Basically, we are getting the MatMul tensors which can be quantized. When you do self.llm.get_tensors, you are only going to ISQ the text model, which should be fine.

mistralrs-core/src/vision_models/llava/llava15.rs
```diff
@@ -0,0 +1,333 @@
...
+        self.config.text_config.max_length
+    }
+
+    fn has_conv2d(&self) -> bool {
+        true
+    }
+}
+
+impl AnyMoeBaseModelMixin for Model {
+    //untested
```
EricLBuehler, line 301: That's fine, given it forwards the methods to tested models, if you could remove this that would be great :)

mistralrs-core/src/vision_models/llava/llava_inputs_processor.rs
```diff
@@ -0,0 +1,328 @@
...
+        MessagesAction::FlattenOnlyText
+    }
+}
+
+impl LLaVAProcessor {
+    pub fn new(config: &str) -> Self {
+        let model_config =
+            serde_json::from_str::<LLaVAConfig>(config).expect("Failed to parse model config.");
+        let image_tag_splitter =
+            Regex::new(r"<\|image_\d+\|>").expect("Failed to compile split regex.");
```
EricLBuehler, line 46: Is this the right image token? From the HF model card I thought it was `<image>`.

mistralrs-core/src/vision_models/llava/llava_llm/llama.rs
```diff
@@ -0,0 +1,609 @@
+//LLaMA without fused RoPE
```
EricLBuehler, line 1: Just to confirm, this is the only reason for not using `models/llama.rs`?

mistralrs-core/src/vision_models/llava/llava_llm/mistral.rs
```diff
@@ -0,0 +1,646 @@
...
+
+        let mut q = q
+            .reshape((b_sz, q_len, self.num_heads, self.head_dim))?
+            .transpose(1, 2)?
+            .contiguous()?;
+        let mut k = k
+            .reshape((b_sz, q_len, self.num_kv_heads, self.head_dim))?
+            .transpose(1, 2)?
+            .contiguous()?;
+        q = OrdinaryRoPE::forward(&q, seqlen_offsets[0], rope_parameter.0, rope_parameter.1)?;
```
EricLBuehler, line 176: Just to confirm, this is the only reason for not using `models/mistral.rs`?

mistralrs-core/src/vision_models/llava/llava_next.rs
```diff
@@ -0,0 +1,426 @@
...
+                context_lens,
+                position_ids,
+            )
+        }
+    }
+}
+
+impl IsqModel for Model {
+    fn get_tensors(&mut self) -> (Vec<(&mut QMatMul, Option<usize>)>, &dyn DeviceMapper) {
+        // I don't really get this part
```
EricLBuehler, line 334: See comment above re. ISQ

mistralrs-core/src/vision_models/llava/llava_next.rs
```diff
@@ -0,0 +1,426 @@
...
+        self.config.text_config.max_length
+    }
+
+    fn has_conv2d(&self) -> bool {
+        true
+    }
+}
+
+impl AnyMoeBaseModelMixin for Model {
+    //untested
```
EricLBuehler, line 394: Same comment as above, if you could remove this that would be great!

mistralrs-core/src/vision_models/llava/llava_next_inputs_processor.rs
```diff
@@ -0,0 +1,386 @@
...
+        MessagesAction::FlattenOnlyText
+    }
+}
+
+impl LLaVANextProcessor {
+    pub fn new(config: &str) -> Self {
+        let model_config =
+            serde_json::from_str::<LLaVANextConfig>(config).expect("Failed to parse model config.");
+        let image_tag_splitter =
+            Regex::new(r"<\|image_\d+\|>").expect("Failed to compile split regex.");
```
EricLBuehler, line 49: I think this should be `<image>`?

mistralrs/Cargo.toml
```diff
@@ -70,7 +70,11 @@ name = "idefics2"
 required-features = []
 
 [[example]]
-name = "anymoe"
```
EricLBuehler, line 73: I think you might have removed `anymoe` and replaced it with `llava` here?

mistralrs/examples/llava/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let (tx, mut rx) = channel(10_000);
+    let request = Request::Normal(NormalRequest {
+        messages: RequestMessage::VisionChat {
+            images: vec![DynamicImage::new(1280, 720, ColorType::Rgb8)],
+            messages: vec![IndexMap::from([
+                ("role".to_string(), Either::Left("user".to_string())),
+                (
+                    "content".to_string(),
+                    Either::Left("<|image_1|>what is this image show?".to_string()),
```
EricLBuehler, line 50: Same comment as above, I think this should be `<image>` as per the HF model card.

mistralrs/examples/llava_next/main.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    let (tx, mut rx) = channel(10_000);
+    let request = Request::Normal(NormalRequest {
+        messages: RequestMessage::VisionChat {
+            images: vec![DynamicImage::new(1280, 720, ColorType::Rgb8)],
+            messages: vec![IndexMap::from([
+                ("role".to_string(), Either::Left("user".to_string())),
+                (
+                    "content".to_string(),
+                    Either::Left("<|image_1|>what is this image show?".to_string()),
```
EricLBuehler, line 51: Same comment as above, I think this should be `<image>` as per the HF model card.

chenwanqq:
Seem quite a lot for me to check!🧐 I will see it later, and just to explain something that:
(1) As I mentioned, I did not use the original image tag processing logic of llava, but instead adopted a processing logic similar to phi3v. This is due to compatibility considerations with the input of this repo. For example, the original logic of llava does not calculate num_img_tokens before encode_image, etc.; Additionally, phi3v's logic is better for handling multiple images. Since <image> in llava is just a placeholder, if we set up the chat_template correctly (the same as the backend llm), the effects of the two implementations will be the same. I think we should add some special notifications?
(2) llava supports multiple images, but this is only practically meaningful in version 1.5. In version 1.6, the introduction of multi-resolution processing has led to longer feature representations for each image, but the context length is only 4K, so multiple images will be truncated.
(3) Yes, it is only the rope problem that I copy the code of LLM.

chenwanqq:
Seem quite a lot for me to check!🧐 I will see it later, and just to explain something that:
(1) As I mentioned, I did not use the original image tag processing logic of llava, but instead adopted a processing logic similar to phi3v. This is due to compatibility considerations with the input of this repo. For example, the original logic of llava does not calculate num_img_tokens before encode_image, etc.; Additionally, phi3v's logic is better for handling multiple images. Since <image> in llava is just a placeholder, if we set up the chat_template correctly (the same as the backend llm), the effects of the two implementations will be the same. I think we should add some special notifications?
(2) llava supports multiple images, but this is only practically meaningful in version 1.5. In version 1.6, the introduction of multi-resolution processing has led to longer feature representations for each image, but the context length is only 4K, so multiple images will be truncated.

mistralrs-core/src/pipeline/vision.rs
```diff
@@ -213,6 +216,7 @@ impl Loader for VisionLoader {
             .get_gen_conf_filename()
             .map(|f| serde_json::from_str(&fs::read_to_string(f).unwrap()).unwrap());
         let chat_template = get_chat_template(paths, &self.chat_template, None);
+        println!("Chat template: {:?}", chat_template);
```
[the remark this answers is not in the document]
chenwanqq: uhh... Ok... it is just a left debug code....

EricLBuehler:
> As I mentioned, I did not use the original image tag processing logic of llava, but instead adopted a processing logic similar to phi3v. This is due to compatibility considerations with the input of this repo. For example, the original logic of llava does not calculate num_img_tokens before encode_image, etc.; Additionally, phi3v's logic is better for handling multiple images. Since  in llava is just a placeholder, if we set up the chat_template correctly (the same as the backend llm), the effects of the two implementations will be the same. I think we should add some special notifications?

I see. I think we should use the Llava `<image>` token to avoid confusion, regardless that it is a placeholder. Should be an easy change, most of the comments I made are about that.

> ) llava supports multiple images, but this is only practically meaningful in version 1.5. In version 1.6, the introduction of multi-resolution processing has led to longer feature representations for each image, but the context length is only 4K, so multiple images will be truncated.

That makes sense. Given that we support multiple images, if we could document that it would be great! 😃

mistralrs-core/src/pipeline/vision.rs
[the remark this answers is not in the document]
EricLBuehler: Thanks!

docs/LLaVA.md
```diff
@@ -0,0 +1,216 @@
...
+
+```py
+from mistralrs import Runner, Which, ChatCompletionRequest, VisionArchitecture
+
+runner = Runner(
+    which=Which.VisionPlain(
+        model_id="llava-hf/llava-v1.6-mistral-7b-hf",
+        tokenizer_json=None,
+        repeat_last_n=64,
+        arch=VisionArchitecture.Idefics2,
```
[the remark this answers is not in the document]
chenwanqq: typo, fixed

docs/LLaVA.md
```diff
@@ -0,0 +1,216 @@
...
+                "content": [
+                    {
+                        "type": "image_url",
+                        "image_url": {
+                            "url": "https://d2r55xnwy6nx47.cloudfront.net/uploads/2018/02/Ants_Lede1300.jpg"
+                        },
+                    },
+                    {
+                        "type": "text",
+                        "text": "What is shown in this image?",
```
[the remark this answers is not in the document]
chenwanqq: I refactor it to use <image> tag. we can see it through

mistralrs-core/src/vision_models/llava/llava15.rs
chenwanqq: removed

mistralrs-core/src/vision_models/llava/llava15.rs
chenwanqq: removed

mistralrs/Cargo.toml
chenwanqq: ok fixed it

chenwanqq:
I've fix the image tag problem and anymoe problem accordingly. May we can have more review?

EricLBuehler:
Hi @chenwanqq! I made some edits for some small updates to the model support matrix and model architecture docs.

This looks good, but it crashes when I run:
```
cargo run --release --features cuda -- --port 1234 --isq Q4K vision-plain -m llava-hf/llava-v1.6-mistral-7b-hf -a llava_next
```
And
```
python3 examples/server/llava_next.py
```
With
```
prompt step - Model failed with error: WithBacktrace { inner: DeviceMismatchBinaryOp { lhs: Cuda { gpu_id: 0 }, rhs: Cpu, op: "conv2d" }, backtrace: Backtrace [{ fn: "candle_core::error::Error::bt" }, { fn: "candle_core::storage::Storage::same_device" }, { fn: "candle_core::storage::Storage::conv2d" }, { fn: "candle_core::conv::<impl candle_core::tensor::Tensor>::conv2d_single_group" }, { fn: "candle_core::conv::<impl candle_core::tensor::Tensor>::conv2d" }, { fn: "<candle_nn::conv::Conv2d as candle_core::Module>::forward" }, { fn: "<mistralrs_core::vision_models::clip::ClipVisionEmbeddings as candle_core::Module>::forward" }, { fn: "mistralrs_core::vision_models::clip::ClipVisionTransformer::forward_get_hidden_states" }, { fn: "mistralrs_core::vision_models::llava::llava_next::Model::encode_images" }, { fn: "<mistralrs_core::vision_models::llava::llava_next::Model as mistralrs_core::pipeline::VisionModel>::forward" }, { fn: "<mistralrs_core::pipeline::vision::VisionPipeline as mistralrs_core::pipeline::Pipeline>::forward_inputs" }, { fn: "mistralrs_core::pipeline::Pipeline::step::{{closure}}" }, { fn: "mistralrs_core::engine::Engine::run::{{closure}}" }, { fn: "tokio::runtime::runtime::Runtime::block_on" }, { fn: "std::sys_common::backtrace::__rust_begin_short_backtrace" }, { fn: "core::ops::function::FnOnce::call_once{{vtable.shim}}" }, { fn: "std::sys::pal::unix::thread::Thread::new::thread_start" }, { fn: "start_thread", file: "./nptl/pthread_create.c", line: 442 }, { fn: "__GI___clone3", file: "./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S", line: 81 }] }
````
It seems to come from the CLIP embeddings. Note that for ISQ models, the VBs are on the CPU and are cast using `mapper.set_nm_device`. Pass `loading_isq` as false if this is not going to be touched by ISQ, and otherwise, be sure to pass the correct parameter from the normal loading metadata.

I would reccomend that you check out the Idefics 2 model, as I do `.set_device(text_model.device().clone())` several times, which will set it to the primary device. This may be what you want. Please let me know if you have any questions!

When I run without ISQ, it successfully produces the correct output though - great work here!

**Comment by chenwanqq** [comment]
> Hi @chenwanqq! I made some edits for some small updates to the model support matrix and model architecture docs.
> 
> This looks good, but it crashes when I run:
> 
> ```
> cargo run --release --features cuda -- --port 1234 --isq Q4K vision-plain -m llava-hf/llava-v1.6-mistral-7b-hf -a llava_next
> ```
> 
> And
> 
> ```
> python3 examples/server/llava_next.py
> ```
> 
> With
> 
> ```
> prompt step - Model failed with error: WithBacktrace { inner: DeviceMismatchBinaryOp { lhs: Cuda { gpu_id: 0 }, rhs: Cpu, op: "conv2d" }, backtrace: Backtrace [{ fn: "candle_core::error::Error::bt" }, { fn: "candle_core::storage::Storage::same_device" }, { fn: "candle_core::storage::Storage::conv2d" }, { fn: "candle_core::conv::<impl candle_core::tensor::Tensor>::conv2d_single_group" }, { fn: "candle_core::conv::<impl candle_core::tensor::Tensor>::conv2d" }, { fn: "<candle_nn::conv::Conv2d as candle_core::Module>::forward" }, { fn: "<mistralrs_core::vision_models::clip::ClipVisionEmbeddings as candle_core::Module>::forward" }, { fn: "mistralrs_core::vision_models::clip::ClipVisionTransformer::forward_get_hidden_states" }, { fn: "mistralrs_core::vision_models::llava::llava_next::Model::encode_images" }, { fn: "<mistralrs_core::vision_models::llava::llava_next::Model as mistralrs_core::pipeline::VisionModel>::forward" }, { fn: "<mistralrs_core::pipeline::vision::VisionPipeline as mistralrs_core::pipeline::Pipeline>::forward_inputs" }, { fn: "mistralrs_core::pipeline::Pipeline::step::{{closure}}" }, { fn: "mistralrs_core::engine::Engine::run::{{closure}}" }, { fn: "tokio::runtime::runtime::Runtime::block_on" }, { fn: "std::sys_common::backtrace::__rust_begin_short_backtrace" }, { fn: "core::ops::function::FnOnce::call_once{{vtable.shim}}" }, { fn: "std::sys::pal::unix::thread::Thread::new::thread_start" }, { fn: "start_thread", file: "./nptl/pthread_create.c", line: 442 }, { fn: "__GI___clone3", file: "./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S", line: 81 }] }
> ```
> 
> It seems to come from the CLIP embeddings. Note that for ISQ models, the VBs are on the CPU and are cast using `mapper.set_nm_device`. Pass `loading_isq` as false if this is not going to be touched by ISQ, and otherwise, be sure to pass the correct parameter from the normal loading metadata.
> 
> I would reccomend that you check out the Idefics 2 model, as I do `.set_device(text_model.device().clone())` several times, which will set it to the primary device. This may be what you want. Please let me know if you have any questions!
> 
> When I run without ISQ, it successfully produces the correct output though - great work here!

That's exactly the problem! With the new commit, I believe I fixed it (tested in llava-v1.6-vicuna-7b-hf, llava-v1.6-mistral-7b-hf, and llava-1.5-7b-hf). However, when I deal with llama-like models (such as Vicuna), there is a device mismatch problem with rmsnorm:
```
prompt step - Model failed with error: DeviceMismatchBinaryOp { lhs: Cuda { gpu_id: 0 }, rhs: Cpu, op: "rms-norm" }
^C
```
When I checked the code, I found that the Llama and Mistral code have slight differences in initialization:
https://github.com/EricLBuehler/mistral.rs/blob/c8b71e9b5894b4f43c58ab600a5144a70d52a4b1/mistralrs-core/src/models/llama.rs#L291-L299
https://github.com/EricLBuehler/mistral.rs/blob/c8b71e9b5894b4f43c58ab600a5144a70d52a4b1/mistralrs-core/src/models/mistral.rs#L288-L296
I change it as Mistral way in my copied version of backend llama llm. I don't have Llama weights right now (and it's not very easy for me to download them). I think maybe you can check if the Llama text model has this issue.

EricLBuehler (approved):
Looks good! Thank you for adding this. I'll need to check the RoPE instability in a future PR but I think the current solution + LlavaModel is good enough.

PR #498, opened 2024-06-30
Warn instead of throw error in chat template

EricLBuehler:


step21:
If this is what I think it is, I would be in favour of this being merged. Like f.e. every time an image does not load or something else in the chat is wrongly formatted, it takes some time to load especially as in-situ-quantization only uses 1 core. (why is that anyway, is that a mistral.rs choice or is that set somewhere else?)

PR #523, opened 2024-07-02
Add judgment for whether model_id is a local path to speed up local model loading

chenwanqq:
In the original logic of marcos.os, regardless of whether model_id is a valid ID on Hugging Face, a request to Hugging Face is made, and if the request fails, it then attempts to load from the local environment. However, if the network environment cannot connect to Hugging Face normally, this step becomes very slow, with each API call attempt taking tens of seconds. Therefore, I have added a workaround here: first, it checks whether model_id is an absolute path. If it is, it directly loads from the local environment. This simple modification directly reduces the time spent on load_model_from_hf from 1 to 2 minutes to just a few microseconds 🤣
Edit: Now I change to check if model_id exists as a local path

### DIFF
--- a/mistralrs-core/src/pipeline/macros.rs
+++ b/mistralrs-core/src/pipeline/macros.rs
@@ -2,14 +2,85 @@
 #[macro_export]
 macro_rules! api_dir_list {
     ($api:expr, $model_id:expr) => {
-        $api.info()
-            .map(|repo| {
-                repo.siblings
-                    .iter()
-                    .map(|x| x.rfilename.clone())
-                    .collect::<Vec<String>>()
-            })
-            .unwrap_or_else(|e| {
+        if std::path::Path::new($model_id).is_absolute() {
+            let listing = std::fs::read_dir($model_id);
+            if listing.is_err() {
+                panic!("Cannot list directory {:?}", $model_id)
+            }
+            let listing = listing.unwrap();
+            listing
+                .into_iter()
+                .map(|s| {
+                    s.unwrap()
+                        .path()
+                        .file_name()
+                        .unwrap() // Should never terminate in `..`
+                        .to_str()
+                        .expect("Could not convert to str")
+                        .to_string()
+                })
+                .collect::<Vec<String>>()
+                .into_iter()
+        } else {
+            $api.info()
+                .map(|repo| {
+                    repo.siblings
+                        .iter()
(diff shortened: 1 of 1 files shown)


mistralrs-core/src/pipeline/macros.rs
```diff
@@ -2,14 +2,85 @@
...
     ($api:expr, $model_id:expr) => {
-        $api.info()
-            .map(|repo| {
-                repo.siblings
-                    .iter()
-                    .map(|x| x.rfilename.clone())
-                    .collect::<Vec<String>>()
-            })
-            .unwrap_or_else(|e| {
+        if std::path::Path::new($model_id).is_absolute() {
```
EricLBuehler, line 5: Can we just check if the path exists? For example, someone may pass `../models/gemma2`.

mistralrs-core/src/pipeline/macros.rs
```diff
@@ -2,14 +2,85 @@
...
+                .into_iter()
+        }
+    };
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! api_get_file {
+    ($api:expr, $file:expr, $model_id:expr) => {
+        if std::path::Path::new($model_id).is_absolute() {
```
EricLBuehler, line 75: Same here, if we could check if the file exists that may be more flexible.

EricLBuehler (commented):
Thank you for adding this! I think that if we check that the file exists, this may be more flexible. Is there a reason why you chose to use `is_aboslute` rather than `exists`?

mistralrs-core/src/pipeline/macros.rs
chenwanqq: Sure, I think maybe this new change is better and more simple🤣

mistralrs-core/src/pipeline/macros.rs
chenwanqq: same above

chenwanqq:
> Thank you for adding this! I think that if we check that the file exists, this may be more flexible. Is there a reason why you chose to use `is_aboslute` rather than `exists`?

Hhh just not sure whether it would break some old logics. And If it is not, we can just check if model_id exists as a local path

EricLBuehler (approved):
Thank you! We keep the logic for file path checking in these macros, so this should be fine.

PR #552, opened 2024-07-05
Optimize multi-batch and inference performance with PagedAttention

EricLBuehler:
Based on work in [candle-vllm](https://github.com/EricLBuehler/candle-vllm). Credit to guoqingbao for his great contributions enabling this!


EricLBuehler:
RUST_BACKTRACE=1 cargo run --release --features cuda -- -i gguf -t mistralai/Mistral-7B-Instruct-v0.1 -m TheBloke/Mistral-7B-Instruct-v0.1-GGUF -f mistral-7b-instruct-v0.1.Q4_K_M.gguf

EricLBuehler:
🚀 This PR indicates about a 13% performance boost! However, it is not yet producing correct output.

The current state is that it runs and the prefill step produces the correct token, but the rest of the decoding steps produce seemingly random tokens. I checked that the `slot_mappings`, `block_tables`, and `context_lens`, `blocks_to_swap_(in|out)` and `blocks_to_copy` are the same for mistral.rs and candle-vllm in the prompt and decoding steps.

@guoqingbao, would you perhaps be able to take a look? The codebase is similar to the candle-vllm one, requests flow from the `Engine` through a `Processor` to apply chat templating, then they are scheduled and prepared with an inputs processor (see `TextInputsProcessor`, `get_(completion|prompt)_input`). Currently I implemented PagedAttention usage only for GGUF Llama, which can be found at `mistralrs-core/src/models/quantized_llama.rs`.

I would appreciate it if you could take a look and see if you could find what is wrong? I can give you collaborator access if you want to push a fix.

guoqingbao:
> 🚀 This PR indicates about a 13% performance boost! However, it is not yet producing correct output.
> 
> The current state is that it runs and the prefill step produces the correct token, but the rest of the decoding steps produce seemingly random tokens. I checked that the `slot_mappings`, `block_tables`, and `context_lens`, `blocks_to_swap_(in|out)` and `blocks_to_copy` are the same for mistral.rs and candle-vllm in the prompt and decoding steps.
> 
> @guoqingbao, would you perhaps be able to take a look? The codebase is similar to the candle-vllm one, requests flow from the `Engine` through a `Processor` to apply chat templating, then they are scheduled and prepared with an inputs processor (see `TextInputsProcessor`, `get_(completion|prompt)_input`). Currently I implemented PagedAttention usage only for GGUF Llama, which can be found at `mistralrs-core/src/models/quantized_llama.rs`.
> 
> I would appreciate it if you could take a look and see if you could find what is wrong? I can give you collaborator access if you want to push a fix.

Thanks Eric for having me. I will try to fix them tomorrow.

EricLBuehler:
> Thanks Eric for having me. I will try to fix them tomorrow.

Thank you! Let me know if you have any trouble.

---
---

More info: Using our fused RoPE the error persists.

guoqingbao:
> > Thanks Eric for having me. I will try to fix them tomorrow.
> 
> Thank you! Let me know if you have any trouble.
> 
> More info: Using our fused RoPE the error persists.

Hi Eric, I've manged to resolve this issue. Do you want me submit a merge request or submit fix to the branch named "pagedattention_2"?.

EricLBuehler:
Yes, that would be amazing!

guoqingbao:
> Yes, that would be amazing! Thank you!

Submitted a patch through #563 

This is blazingly fast indeed (under quantization + paged-attention) :)

EricLBuehler:
@guoqingbao it works for Llama 2, but strangely not for Mistral? Not sure if you saw this before when working on candle-vllm, it would be greatly appreciated if you could take another look!

EricLBuehler:
More info: I tested the RoPE change merged #563 in a PR #565, and it made the output incorrect for Mistral and Llama 2.

EricLBuehler:
@guoqingbao fixed it!

guoqingbao:
> @guoqingbao just fixed it! The problem was the transpose, the RoPE was incorrect but when you added the transpose it fixed it :)

Great! We need to take care the usage of transpose before Rope and after Paged attention. Some of the transpose ops are intended used before Rope and others are used to restore Paged attention and profiling outputs, remember the output dim orders of Paged attention and profiling are different.

mistralrs-core/src/pipeline/inputs_processor.rs
```diff
@@ -148,6 +262,47 @@ pub mod text_models_inputs_processor {
...
+            if let Some(ref paged_attn_metadata) = paged_attn_metadata {
+                let table = paged_attn_metadata.block_tables.get(seq.id()).unwrap();
+
+                let table = table
+                    .iter()
+                    .map(|block| block.deref_mut().block_id)
+                    .collect::<Vec<_>>();
+
+                let block_number = if start_pos / paged_attn_metadata.block_size >= table.len() {
+                    table.last().unwrap() //position exceed! use last position; TODO (bug fix)
```
EricLBuehler, line 275: @guoqingbao I was wondering what you meant with this comment?

guoqingbao: I encounter the problem of start_pos / paged_attn_metadata.block_size larger than the size of block table in candle-vllm. It seems that the calculation of block table size has some bugs. That's why I padded the prompt tokens into odd number.

EricLBuehler: We don't do any padding of the prompt to an odd number, do you think this is a problem here? Perhaps a solution is to make sure we allocate one extra physical token block?

guoqingbao: From my experience in candle-vllm, the program will be crashed in some cases if no padding to odd number. I think the root cause lies in the allocation of block tables. Perhaps extra block table can address the problem. I haven't figure out how to solve it permanently in an elegant way.

EricLBuehler:
- [ ] Prefix caching (long term; langs: CUDA + Rust)
- [ ] Gemma 2 softcapping in paged attention (near term; langs: CUDA + Rust)
- [ ] GGML support (near term; langs: Rust)
- [ ] Vision support (near term; langs: Rust)
- [ ] AnyMoE support (near term; langs: Rust)

mistralrs-core/src/pipeline/inputs_processor.rs
EricLBuehler: Yeah, this is tricky. I took a look at the code which manages the logical and physical token blocks (the BlockEngine :) ), it appears to do what we want: add the token to a non-full block or append a new one and start filling that up.

Specifically, let's say block_size = 32, and we add 32 prompt tokens: we will have 1 logical token block, so when `BlockEngine::allocate` is called, we make one physical token block. So `tables.len == 1`. But `start_pos == 32-1 == 31`, so `31 / 32 == 0 < 1`. 

This looks like an edge case, I can't think of any case where this would be true. Perhaps I am not understanding the error case correctly? Would it be correct to assume that this occurs any time `start_pos` is even if there is an even number of prompt tokens (this check is not in the decoding preparation, so maybe that is the case)?

guoqingbao: This problem occurs especially when the start_pos is divisible by the block size (e.g., 32). Normally, when the start_pos is divisible by the block size (say 64 / 32), the number of block tables should be at least 3, but that is not the case. So that I padded it to the odd number to ensure extra block table in the corner case. Perhaps we are not taking account the start token when calculating the number of block tables, for example, a request of 100 tokens would occupy 101 token space in the block tables.

guoqingbao: > This problem occurs especially when the start_pos is divisible by the block size (e.g., 32). Normally, when the start_pos is divisible by the block size (say 64 / 32), the number of block tables should be at least 3, but that is not the case. So that I padded it to the odd number to ensure extra block table in the corner case. Perhaps we are not taking account the start token when calculating the number of block tables, for example, a request of 100 tokens would occupy 101 token space in the block tables before decoding.

EricLBuehler:
cargo run --release --features cuda -- -i --pa-gpu-mem 4096 --pa-blk-size 32 gguf -t mistralai/Mistral-7B-Instruct-v0.1 -m TheBloke/Mistral-7B-Instruct-v0.1-GGUF -f mistral-7b-instruct-v0.1.Q4_K_M.gguf

EricLBuehler:
@guoqingbao, if you could skim the code and give a small review, that would be much appreciated!

mistralrs-core/src/pipeline/inputs_processor.rs
EricLBuehler: Ah ok. I reproduced the error :)

guoqingbao:
> @guoqingbao, if you could skim the code and give a small review, that would be much appreciated!

Sorry for the jet lag. Have you found the root cause of the block table issue? That is very tricky.

EricLBuehler:
> Sorry for the jet lag. Have you found the root cause of the block table issue? That is very tricky.

Yep, I fixed it! We weren't allocating the correct number of physical token blocks. I also added a PR to candle-vllm for the fix, in case you want to take a look.

mistralrs-core/src/pipeline/inputs_processor.rs
guoqingbao: > Ah ok. I reproduced the error :)

Have you checked the allocation of block tables on the basis of the number of input tokens (prompt)? I think the problem is there, you may pad the input tokens to the size of divisible by the block size, it will be easier to reproduce and debug. :)

guoqingbao:
> > Sorry for the jet lag. Have you found the root cause of the block table issue? That is very tricky.
> 
> Yep, I fixed it! We weren't allocating the correct number of physical token blocks. I also added a PR to candle-vllm for the fix, in case you want to take a look.

Thanks, that's great! I will take a look and perhaps the padding logic can be removed.

EricLBuehler:
> > > Sorry for the jet lag. Have you found the root cause of the block table issue? That is very tricky.
> > 
> > 
> > Yep, I fixed it! We weren't allocating the correct number of physical token blocks. I also added a PR to candle-vllm for the fix, in case you want to take a look.
> 
> Thanks, that's great! I will take a look and perhaps the padding logic can be removed.

Alright, I fixed it :) We just needed a function to check if the logical token block is empty. If that is the case, we want to allocate the physical token block. I'll update the candle-vllm PR.

PR #572, opened 2024-07-14
Add the OpenVLA model

EricLBuehler:
https://huggingface.co/openvla/openvla-7b


haixuanTao:
Looking forward to this!

PR #626, opened 2024-07-25
Sampling on the GPU for as long as possible

EricLBuehler:
Currently, we apply all sampling:
- Sequentially
- On the CPU

This is super slow. This PR is going to refactor the sampling system to do as much sampling work on the GPU, in parallel, as much as possible until we need to copy the final token & logprobs to the CPU. Only then is the final GPU <> CPU sync done.


EricLBuehler:
Pending some resolution of huggingface/candle#2361, otherwise we still have to do a huge GPU <> CPU sync early.

Issue #635, opened 2024-07-27
Add DRY repetition penalty


p-e-w:
DRY is a modern repetition penalty that is superior to the standard frequency and presence penalties at preventing repetition, while having virtually none of their negative effects on language quality.

Original implementation with detailed explanation: https://github.com/oobabooga/text-generation-webui/pull/5677

C++ implementation in llama.cpp (pending merge): https://github.com/ggerganov/llama.cpp/pull/6839

Independent C++ implementation in Kobold (merged): https://github.com/LostRuins/koboldcpp/pull/982

EricLBuehler:
@p-e-w I just added a PR for this. If you could give it a review that would be amazing.

PR #637, opened 2024-07-27
Implement DRY penalty

EricLBuehler:
@p-e-w, could you please give the implementation a quick check? I'm not sure if you are familiar with Rust, but I ported the algorithm from the oobabooga implemenation you linked.

Refs #635.


p-e-w:
Thank you for implementing this so quickly! I have submitted my review in the form of a pull request into this branch: #645.

EricLBuehler:
Hey @p-e-w! Do you think this is ready to merge?

p-e-w:
@EricLBuehler

Give me a few days to test and verify this, I will let you know once I'm sure. Looks fine at first glance though!

mistralrs-core/src/sampler.rs
```diff
@@ -12,9 +12,12 @@ use pyo3::pyclass;
 
 use rand::distributions::{Distribution, WeightedIndex};
 use rand_isaac::Isaac64Rng;
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use serde::{Deserialize, Serialize};
 use tokenizers::Tokenizer;
 
+const SEQUENCE_BREAKERS: &[&str] = &["\n", ":", "\\", "*"];
```
polarathene, line 19: I assume `Vec<String>` is only used elsewhere to avoid lifetime annotation with the struct? (_I haven't gone through the entire PR_)

This suggestion might not make much of a difference, but you could initialize this once here with `OnceLock` / `OnceCell` (Rust 1.70), or with Rust 1.80 there is also `LazyLock` now available:

```suggestion
use std::sync::LazyLock;
static SEQUENCE_BREAKERS: LazyLock<Vec<String>> = LazyLock::new(|| {
    ["\n", ":", "\\", "*"].map(String::from).to_vec()
});
```

You then have it initialized once on first access and can deref for the inner type to access the `Vec<String>` with `*SEQUENCE_BREAKERS`.

Then later if you do need to store a copy, you can use `unwrap_or(*SEQUENCE_BREAKERS.clone())`, or just access the static variable since it's already `Vec<String>`.

mistralrs-core/src/sampler.rs
```diff
@@ -52,10 +56,92 @@ impl Default for SamplingParams {
...
+                    .sequence_breakers
+                    .into_iter()
+                    .map(|breaker| {
+                        tokenizer
+                            // Prefix with 'a' to get the correct encoding of the token at the end of a text.
+                            //
+                            // FIXME: This is a hack. See https://github.com/LostRuins/koboldcpp/pull/982
+                            //        for the correct solution which covers multi-token sequence breakers
+                            //        and ambiguous encodings.
+                            .encode(format!("a{breaker}"), true)
```
polarathene, line 124: `format!` is slow IIRC, if that actually matters here for performance you'll find `.concat()` on an array works better:

```suggestion
                            .encode(["a", breaker].concat()), true)
```

polarathene (commented):
Just some optional suggestions, I've not really done a proper review.

EricLBuehler:
@polarathene thanks for the review - nice to see you back! I've merged them, thanks for the suggestions!

p-e-w:
@EricLBuehler 

I'm having a hard time testing this properly because of #666.

mistralrs-core/src/sampler.rs
```diff
@@ -488,7 +488,7 @@ impl Sampler {
             let match_indices = toks
                 .par_iter()
                 .enumerate()
-                .take(toks.len() - 1)
+                .take(toks.len().saturating_sub(1))
```
p-e-w, line 491: Just ran into this problem before I noticed your commit. But why can `toks.len()` be 0 here in the first place? That doesn't make sense to me. If I enter a prompt (e.g. "test"), then `apply_dry_penalty` is called with an empty context. Why?

EricLBuehler: So this can only be caused when sampling the token resulting from prompt processing. We discard all prompt tokens from the penalty context:

> https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/pipeline/sampling.rs#L290

Does DRY require those tokens to be included?

p-e-w: I'm afraid I don't understand. Why is sampling needed during prompt processing?

Sure, the transformer will by construction always output logits for each token position. But samplers are only getting involved once a token is actually drawn from the resulting distribution. And that only happens when new tokens are generated, right? In which case the context should always be non-empty after a prompt has been entered.

EricLBuehler: > Why is sampling needed during prompt processing?

After the model processes the prompt, it produces a token distribution as a result and we sample that. The `Sampler::apply_penalties` method has inherent support for the case where the generated tokens do not exist yet, as we just iterate over that context. Perhaps we need a case here to handle this?

p-e-w: I feel bad for taking so much of your time with this, but I just don't get it. Why does sampling ever happen with an empty context? The only purpose of sampling (i.e., *drawing* from the probability distribution, rather than merely generating the distribution) is to generate a new token, no?

Let's say the user enters the prompt "Hello", and then runs generation. Why is `Sampler::apply_penalties` called with an empty context, rather than with context "Hello"? The program doesn't need to sample during prompt processing (since no tokens are being generated at previous positions), so why is this happening?

p-e-w: Thank you, I understand now. So this is intended to be a feature?

It seems that this breaks many assumptions made by frontends (and users) about how things work. The most problematic effect is that it makes sampling a stateful affair, where the original prompt range is somehow "remembered" during generation, and the context given to the samplers doesn't match the input given to the transformer.

In a chat interface, this means repetition penalties cannot take into account previous messages while generating a new one... which is the exact opposite of what we want, since models tend to repeat previous messages. But it also seems wrong philosophically. If you cancel the generation process midway, and then restart it from that position, you might get *different* tokens than you would have if you had allowed it to complete, because now the prompt includes the partially generated output.

I don't believe other loaders do this, and IMO this mechanism should indeed be removed completely, for all samplers.

EricLBuehler: Thanks for pointing this out, I agree that the statefulness makes this incorrect!

> I don't believe other loaders do this, and IMO this mechanism should indeed be removed completely, for all samplers.

Sounds good! I've removed this functionality completely now.

EricLBuehler:
@p-e-w, does this look correct?

`cargo run --features cuda -- -i --isq q4k plain -m microsoft/Phi-3-mini-128k-instruct -a phi3`
```
2024-08-20T13:42:48.556788Z  INFO mistralrs_server: avx: true, neon: false, simd128: false, f16c: true
2024-08-20T13:42:48.556837Z  INFO mistralrs_server: Sampling method: penalties -> temperature -> topk -> topp -> minp -> multinomial
2024-08-20T13:42:48.556848Z  INFO mistralrs_server: Model kind is: normal (no quant, no adapters)
2024-08-20T13:42:48.556957Z  INFO mistralrs_core::pipeline::normal: Loading `tokenizer.json` at `microsoft/Phi-3-mini-128k-instruct`
2024-08-20T13:42:48.556993Z  INFO mistralrs_core::pipeline::normal: Loading `config.json` at `microsoft/Phi-3-mini-128k-instruct`
2024-08-20T13:42:48.722918Z  INFO mistralrs_core::pipeline::paths: Found model weight filenames ["model-00001-of-00002.safetensors", "model-00002-of-00002.safetensors"]
2024-08-20T13:42:48.767432Z  INFO mistralrs_core::pipeline::normal: Loading `generation_config.json` at `microsoft/Phi-3-mini-128k-instruct`
2024-08-20T13:42:48.873727Z  INFO mistralrs_core::pipeline::normal: Loading `tokenizer_config.json` at `microsoft/Phi-3-mini-128k-instruct`
2024-08-20T13:42:48.874018Z  INFO mistralrs_core::pipeline::normal: Loading model `microsoft/Phi-3-mini-128k-instruct` on cuda[0].
2024-08-20T13:42:48.935684Z  INFO mistralrs_core::utils::normal: Detected minimum CUDA compute capability 8.9
2024-08-20T13:42:49.023637Z  INFO mistralrs_core::utils::normal: DType selected is BF16.
2024-08-20T13:42:49.023743Z  INFO mistralrs_core::pipeline::normal: Model config: Config { vocab_size: 32064, hidden_act: Silu, hidden_size: 3072, intermediate_size: 8192, num_hidden_layers: 32, num_attention_heads: 32, num_key_value_heads: 32, rms_norm_eps: 1e-5, rope_theta: 10000.0, bos_token_id: Some(1), eos_token_id: Some(32000), rope_scaling: Some({"type": Phi3RopeScaling(Right("longrope")), "long_factor": Phi3RopeScaling(Left([1.0700000524520874, 1.1200000047683716, 1.149999976158142, 1.4199999570846558, 1.569999933242798, 1.7999999523162842, 2.129999876022339, 2.129999876022339, 3.009999990463257, 5.910000324249268, 6.950000286102295, 9.070000648498535, 9.93000030517578, 10.710000038146973, 11.130000114440918, 14.609999656677246, 15.409998893737791, 19.809999465942383, 37.279998779296875, 38.279998779296875, 38.599998474121094, 40.12000274658203, 46.20000457763672, 50.94000625610352, 53.66000747680664, 54.9373893737793, 56.89738845825195, 57.28738784790039, 59.98738479614258, 60.86738586425781, 60.88738632202149, 61.71739196777344, 62.91739273071289, 62.957393646240234, 63.41739273071289, 63.8173942565918, 63.83739471435547, 63.89739608764648, 63.93739700317383, 64.06739807128906, 64.11434936523438, 64.12435150146484, 64.15435028076172, 64.19435119628906, 64.24435424804688, 64.57435607910156, 64.69000244140625, 64.76000213623047])), "short_factor": Phi3RopeScaling(Left([1.1, 1.1, 1.1, 1.3000000000000005, 1.3500000000000003, 1.3500000000000003, 1.4000000000000004, 1.5500000000000005, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.000000000000001, 2.0500000000000007, 2.0500000000000007, 2.0500000000000007, 2.0500000000000007, 2.0500000000000007, 2.0500000000000007, 2.1000000000000005, 2.1000000000000005, 2.1500000000000004, 2.25, 2.25, 2.25, 2.25, 2.25, 2.3999999999999995, 2.4499999999999993, 2.499999999999999, 2.6999999999999984, 2.6999999999999984, 2.7499999999999982, 2.799999999999998, 2.8999999999999977, 3.049999999999997]))}), max_position_embeddings: 131072, use_flash_attn: false, sliding_window: Some(262144), original_max_position_embeddings: 4096, quantization_config: None }
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 67/67 [00:01<00:00, 77.49it/s]
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 128/128 [00:02<00:00, 23.66it/s]
2024-08-20T13:42:52.401371Z  INFO mistralrs_core::pipeline::isq: Applying in-situ quantization into Q4K to 129 tensors.
2024-08-20T13:42:52.403947Z  INFO mistralrs_core::pipeline::isq: Applying ISQ on 22 threads.
2024-08-20T13:43:02.482772Z  INFO mistralrs_core::pipeline::isq: Applied in-situ quantization into Q4K to 129 tensors out of 129 total tensors. Took 10.08s
2024-08-20T13:43:02.483047Z  INFO mistralrs_core::paged_attention: Allocating 2379 MB for PagedAttention KV cache
2024-08-20T13:43:02.483083Z  INFO mistralrs_core::paged_attention: Using PagedAttention with block size 32 and 198 GPU blocks: available context length is 6336 tokens
2024-08-20T13:43:02.570913Z  INFO mistralrs_core::pipeline::chat_template: bos_toks = "<s>", eos_toks = "<|endoftext|>", "<|assistant|>", "<|end|>", unk_tok = <unk>
2024-08-20T13:43:02.575197Z  INFO mistralrs_server: Model loaded.
2024-08-20T13:43:02.577222Z  INFO mistralrs_core: Enabling GEMM reduced precision in BF16.
2024-08-20T13:43:02.842321Z  INFO mistralrs_core: Enabling GEMM reduced precision in F16.
2024-08-20T13:43:02.844652Z  INFO mistralrs_core::cublaslt: Initialized cuBLASlt handle
2024-08-20T13:43:02.844801Z  INFO mistralrs_server::interactive_mode: Starting interactive loop with sampling params: SamplingParams { temperature: Some(0.1), top_k: Some(32), top_p: Some(0.1), min_p: Some(0.05), top_n_logprobs: 0, frequency_penalty: Some(0.1), presence_penalty: Some(0.1), stop_toks: None, max_len: Some(4096), logits_bias: None, n_choices: 1, dry_params: Some(DrySamplingParams { sequence_breakers: ["\n", ":", "\\", "*"], multiplier: 1.0, base: 1.75, allowed_length: 2 }) }
> hi
Hello! How can I assist you today?
> what is graphene
Graphene is a single layer of carbon atoms arranged in a two-dimensional honeycomb lattice. It is renowned for its exceptional strength, flexibility, electrical conductivity, and transparency. Discovered in 2004 by Andre Geim and Konstantin Novoselov, who later won the Nobel Prize in Physics for their work, graphene has potential applications in various fields such as electronics, energy storage, and materials science.
> write an essay
Title: The Revolutionary Potential of Graphene


Introduction:

Graphene, a material composed of a single atomic layer of sp2-bonded carbon atoms, has emerged as one of the most promising materials of the 21st century. Its discovery has sparked a revolution in material science, with potential applications that span across multiple industries. This essay explores the unique properties of graphene and its transformative potential in revolutionizing technology and industry.

Body:
1. Historical Context:
   - Discuss the discovery of graphite and its limitations.
   – Highlight the groundbreaking work of Andre Geib and Konstantin Novosolov.
   
2. Unique Properties:
    - Elaborate on the exceptional mechanical strength, which is over 100 times stronger than steel.
     - Discus the remarkable electrical and thermal conductivity that surpasses copper.
      - Explain the transparence and flexibility that make it suitable for transparent conductive films.
      
3. Applications:
     a. Electronics:
        - Describe how graphene's high electron mobility could lead to faster and more efficient transistors.
        – Explore potential uses in flexible displays and touchscreens.
        
     b. Energy Storage:
       - Discover how graphite's surface area can enhance battery performance in energy storage devices.
          - Predict future advancements in supercapacitors and lithium-ion batteries.
          
     c. Material Science:
      – Examine how grapheme's strength could lead the development of ultra-lightweight materials for aerospace applications.
            - Consider its role in improving the durability and performance of sports equipment.
            
4. Challenges and Future Prospects:
  - Address the current challenges in mass production and integration into existing technologies.
  – Speculate on future research directions that could overcome these obstacles.
- Conclude by emphasizing the transformative impact that graphene could have on technology and society at large, urging continued investment and research in this field.
>
```

mistralrs-core/src/sampler.rs
```diff
@@ -379,8 +488,71 @@ impl Sampler {
...
+                    match_lengths.insert(next_token, match_length);
+                }
+            }
+
+            // Actually apply penalties
+            for (tok, match_len) in match_lengths {
+                if match_len >= params.allowed_length {
+                    let penalty = params.multiplier
+                        * params.base.powf((match_len - params.allowed_length) as f32);
+                    logits[tok as usize] = penalty;
```
p-e-w, line 551:
```suggestion
                    logits[tok as usize] -= penalty;
```

@EricLBuehler 

:rocket: :exclamation: :exclamation: :exclamation: :exclamation: :exclamation: :rocket: 

Finally, I have found the devil, after looking under every rock! I must have read past that line half a dozen times while debugging, thinking "it can't be there", before noticing the problem. :stuck_out_tongue: 

I'm still not 100% certain the implementation is correct, but the artifacts are all gone now, and both with and without DRY the output looks OK.

EricLBuehler: @p-e-w thanks for finding this!! This sounds good, please feel free to approve the PR or do any additional reviews. I'll do some more testing on my end and hopefully we can merge this today.

mistralrs-core/src/sampler.rs
```diff
@@ -12,9 +12,14 @@ use pyo3::pyclass;
 
 use rand::distributions::{Distribution, WeightedIndex};
 use rand_isaac::Isaac64Rng;
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use serde::{Deserialize, Serialize};
+use std::sync::LazyLock;
 use tokenizers::Tokenizer;
 
+static SEQUENCE_BREAKERS: LazyLock<Vec<String>> =
+    LazyLock::new(|| ["\n", ":", "\\", "*"].map(String::from).to_vec());
```
p-e-w, line 21:
```suggestion
    LazyLock::new(|| ["\n", ":", "\"", "*"].map(String::from).to_vec());
```

Typo probably.

mistralrs-core/src/sampler.rs
```diff
@@ -52,10 +58,91 @@ impl Default for SamplingParams {
...
+            sequence_breakers: sequence_breakers.unwrap_or(SEQUENCE_BREAKERS.clone()),
+            multiplier,
+        })
+    }
+}
+
+impl Default for DrySamplingParams {
+    fn default() -> Self {
+        Self {
+            multiplier: 1.0,
```
p-e-w, line 93:
```suggestion
            multiplier: 0.0,
```

DRY should be disabled by default, because it can cause artifacts with highly-regular output like code.

p-e-w (approved):
@EricLBuehler 

* [X] Verified that the code is equivalent to the original Python implementation.
* [X] Instrumented `apply_dry_penalty` and checked that it produces the expected penalties.
* [X] Tested with repetitive output and validated that this version of DRY actually prevents repetition.

As far as I'm concerned, this is now ready to be merged (after applying the two changes above).

One thing you might consider is to disable DRY completely if `multiplier` is 0, which is what other implementations are doing. Currently, matching is still performed in this case, but has no effect because the resulting penalty is zero. That's a lot of unnecessary work that could be skipped by just not invoking `apply_dry_penalty` in the first place (and the same optimization could be applied for `apply_freq_presc_penalty` I think).

mistralrs-core/src/sampler.rs
EricLBuehler: Thanks!

mistralrs-core/src/sampler.rs
EricLBuehler: Sounds good. Thanks for pointing this out.

EricLBuehler:
@p-e-w @polarathene thank you for your reviews! I'll merge this PR as it looks good and generation is great with it!

PR #640, opened 2024-07-28
Allow setting PagedAttention KV cache allocation from context size

EricLBuehler:
Refs #622,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2319,7 +2319,6 @@ dependencies = [
  "candle-core",
  "clap",
  "cli-table",
- "either",
  "mistralrs-core",
  "serde",
  "serde_json",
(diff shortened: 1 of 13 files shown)


EricLBuehler:
@oldgithubman @mcm007 this implements allocating KV cache allocation for a specific number of tokens, when using CUDA.

@oldgithubman:
> The problem with dynamically-allocated cache (assuming no limiter functionality), is usually, if I'm targeting a certain performance level (for example, RAM), I don't want the cache to grow beyond RAM and even using all my RAM is bad. So, I think a limiter is still useful. Again, it could just be a matter of adapting to a new workflow though. Being able to prevent cache from using swap is very important though, since this adds wear and tear on expensive nvme's

I will add this in a future PR!

PR #649, opened 2024-07-29
Implement tool calling

EricLBuehler:


EricLBuehler:
@mitsuhiko I'm having some issues with the Jinja. The Jinja chat template ([found here](https://huggingface.co/meta-llama/Meta-Llama-3.1-8B-Instruct/blob/main/tokenizer_config.json#L2053)) fails with the error `invalid operation: cannot convert map to bool (in chat_template:57)`.

I would really appreciate it if you could check this out.

I located the section of code which I think is the problem below:
```
{{- bos_token }}{%- if custom_tools is defined %}{%- set tools = custom_tools %}
```

To reproduce this, please git clone mistral.rs, and then run:
(with an Nvidia GPU)
```
git switch tool_calling
cargo run --release --features cuda -- --port 1234 plain -m meta-llama/Meta-Llama-3.1-8B-Instruct -a llama
```
(with a Mac)
```
git switch tool_calling
cargo run --release --features metal -- --port 1234 plain -m meta-llama/Meta-Llama-3.1-8B-Instruct -a llama
```

And then, after it is done loading, in another terminal:
```
python3 examples/server/tool_calling_llama_31.py
```

EricLBuehler:
@mitsuhiko ok, looks like `tojson` needs to take kwargs - in particular, `indent`. Would it be OK if I open a PR?

mitsuhiko:
> @mitsuhiko ok, looks like `tojson` needs to take kwargs - in particular, `indent`. Would it be OK if I open a PR?

Yes, happy to accept patches there.  Also just an issue is fine that describes the exact issue.

EricLBuehler:
@mitsuhiko thanks for merging that!

mitsuhiko:
The changes will go out in minijinja 2.1.1

EricLBuehler:
> The changes will go out in minijinja 2.1.1

Ah ok. I'll put out a release before then, but I'll include them in the next one!

PR #684, opened 2024-08-14
Initial KV RingAttention code

joshpopelka20:
This is the start of the RingAttention code. The changes so far have been to create multiple KV caches (if multiple num_devices) and to try to create separate chunks.


mistralrs-core/src/models/llama.rs
```diff
@@ -421,18 +423,35 @@ impl Llama {
...
-            x.dtype(),
+            // x.dtype(),
+            chunks[0].dtype(),
             self.blocks[0].attn.num_attention_heads,
         )?;
         for (block_idx, block) in self.blocks.iter().enumerate() {
-            x = self.mapper.map(x, block_idx)?;
+            // x = self.mapper.map(x, block_idx)?;
+            // x = self.mapper.map(&chunks[0], block_idx)?;
+            x = self.mapper.map(chunks[0].copy().unwrap(), block_idx)?;
```
EricLBuehler, line 454: I think this is the bit which is causing the issue. It looks like we aren't using the values from the last block as the inputs are always from the embeddings.

joshpopelka20: So that's going to be a problem. From the algorithm in the paper, I need to use the local block on each host. So I need to iterate through the layers and then through the local blocks.

image

Any suggestions on how to redesign it?

EricLBuehler (commented):
Thank you for adding the PR!

So, I think I see what the issue, I have attached a comment below.

I read [this article](https://medium.com/@iamtanujsharma/breaking-the-boundaries-understanding-context-window-limitations-and-the-idea-of-ring-attention-170e522d44b2) which seems to indicate we want to load weights on the devices and then at runtime, use [`Tensor::chunk`](https://docs.rs/candle-core/0.6.0/candle_core/struct.Tensor.html#method.chunk) to process the input sequence between multiple devices. I assume that then most of the work will happen in the attention code. Is this correct?

joshpopelka20:
I didn't plan to use the chunk method. I plan on using IndexOp. I think I had some issues with the chunk method, but don't remember why. 

I kept the forward block mostly the same. I just moved the feedforward (MLP layer) after attention is run on all chunks. Not 100% sure it'll work, but I'm trying to follow the algorithm as close as possible.

EricLBuehler:
To use [`Tensor::chunk`](https://docs.rs/candle-core/0.6.0/candle_core/struct.Tensor.html#method.chunk) maybe we can split the input ids with dim = D::Minus1 into the number of devices chunks.

After that, we can just follow the algorithm for the rest. I would also recommend you split this into another model implementation, perhaps `llama_ring_attention.rs` and associated NormalModelLoader as this seems to be quite invasive.

joshpopelka20:
>  Essentially you need to split Block into a few steps.

Can you give me more details on the steps? That's where I'm confused.

>  I would also recommend you split this into another model implementation, perhaps llama_ring_attention.rs and associated NormalModelLoader as this seems to be quite invasive.

I can do this

joshpopelka20:
I did a simple test to the mapper function that you pointed out above. I've confirmed that it is the problem.

Here is the code snippet of the test:
image

The model now gives the correct output. 

When you have the time, please let me know what should be done to create a more robust solution. I'm not sure how to proceed with it.

Issue #630, opened 2024-07-25
Streamed inference not as smooth (fast?) as with e.g. Ollama - Llama 3.1


ChristianWeyer:
## Describe the bug
Have a look :-)

https://github.com/user-attachments/assets/321dbb21-2403-4330-9ce1-091902298888

## Latest commit or version
0.22
MBP M3 Max

ChristianWeyer:
Sure!

Ollama has `--verbose`:

```
❯ ollama run llama3.1:8b-instruct-fp16 --verbose
>>> tell me a joke
Here's one:

What do you call a fake noodle?

An impasta.

total duration:       1.29921225s
load duration:        34.187542ms
prompt eval count:    15 token(s)
prompt eval duration: 483.086ms
prompt eval rate:     31.05 tokens/s
eval count:           18 token(s)
eval duration:        781.205ms
eval rate:            23.04 tokens/s
```


Is there anything similar for mistral.ai @EricLBuehler ?

EricLBuehler:
Yes, mistral.rs has --throughput before the model selector (plain). It can be used with the server.

ChristianWeyer:
Is there a trick to see the throughput values in interactive mode? Or does it not work with -i?

```
❯ cargo run --release --features metal -- -i --throughput plain -m meta-llama/Meta-Llama-3.1-8B-Instruct -a llama
    Finished `release` profile [optimized] target(s) in 0.43s
     Running `target/release/mistralrs-server -i --throughput plain -m meta-llama/Meta-Llama-3.1-8B-Instruct -a llama`
2024-07-25T18:40:01.534562Z  INFO mistralrs_server: avx: false, neon: true, simd128: false, f16c: false
2024-07-25T18:40:01.534616Z  INFO mistralrs_server: Sampling method: penalties -> temperature -> topk -> topp -> minp -> multinomial
2024-07-25T18:40:01.534652Z  INFO mistralrs_server: Model kind is: normal (no quant, no adapters)
2024-07-25T18:40:01.535339Z  INFO mistralrs_core::pipeline::normal: Loading `tokenizer.json` at `meta-llama/Meta-Llama-3.1-8B-Instruct`
2024-07-25T18:40:01.535540Z  INFO mistralrs_core::pipeline::normal: Loading `config.json` at `meta-llama/Meta-Llama-3.1-8B-Instruct`
[...]

PR #685, opened 2024-08-15
Add missing error case in automatic dtype selection feature

ac3xx:
Fixes https://github.com/EricLBuehler/mistral.rs/issues/630#issuecomment-2288551704


EricLBuehler (approved):
Thank you very much @ac3xx!

PR #702, opened 2024-08-21
Add custom logits processor API

EricLBuehler:
New sampling order:

1) Apply penalties from sampling_params
2) ⭐ Apply these custom logits processors sequentially
3) Apply temperature and softmax
4) Sample the next token (topk, topp, minp, etc)

@Dan-wanna-M would this API work? You can find an example of using it [here](https://github.com/EricLBuehler/mistral.rs/blob/9078786881b229f1d4f740b9cc570ee46be4e860/mistralrs/examples/custom_logits_processor/main.rs#L81-L83). Basically, you provide a Vec of `Arc<dyn Fn(&Tensor, &[u32]) -> Result<Tensor> + Send + Sync>`, which are applied. 

Please let me know if this works for you or what changes would be necessary, and I can merge this PR. [kbnf](https://github.com/Dan-wanna-M/kbnf) looks super cool!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2326,6 +2326,7 @@ dependencies = [
  "image",
  "indexmap 2.4.0",
  "mistralrs-core",
+ "rand",
  "reqwest",
  "serde",
  "serde_json",
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Mistal.rs supports several model categories:
     - Please suggest more by raising an issue!
 - Tool calling: [docs](docs/TOOL_CALLING.md)
 - Prompt chunking (only without PagedAttention for now): handle larger prompts where the activation size would cause an OOM by sending chunks
+- Custom logits processor API in Rust: [example](mistralrs/examples/custom_logits_processor/main.rs)
 
 
 This is a demo of interactive mode with streaming running Phi 3 128k mini with quantization via ISQ to Q4K.
--- a/mistralrs-bench/src/main.rs
+++ b/mistralrs-bench/src/main.rs
@@ -80,6 +80,7 @@ fn run_bench(
         adapters: None,
         tools: None,
         tool_choice: None,
+        logits_processors: None,
     });
 
     let mut usages = Vec::new();
@@ -246,6 +247,7 @@ fn warmup_run(mistralrs: Arc<MistralRs>) {
         adapters: None,
         tools: None,
         tool_choice: None,
+        logits_processors: None,
     });
 
     sender
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -650,6 +650,7 @@ impl Engine {
             topk,
             topp,
             minp,
+            request.logits_processors.unwrap_or_default(),
         );
 
         if request.sampling_params.n_choices == 0 {
--- a/mistralrs-core/src/lib.rs
+++ b/mistralrs-core/src/lib.rs
@@ -76,7 +76,7 @@ pub use pipeline::{
 pub use request::{Constraint, MessageContent, NormalRequest, Request, RequestMessage};
 pub use response::Response;
 pub use response::*;
-pub use sampler::{GenericLogitsProcessor, SamplingParams, StopTokens, TopLogprob};
+pub use sampler::{CustomLogitsProcessor, SamplingParams, StopTokens, TopLogprob};
 pub use scheduler::{DefaultSchedulerMethod, SchedulerConfig};
 use serde::Serialize;
 use tokio::runtime::Runtime;
--- a/mistralrs-core/src/lib.rs
+++ b/mistralrs-core/src/lib.rs
@@ -76,7 +76,7 @@ pub use pipeline::{
 pub use request::{Constraint, MessageContent, NormalRequest, Request, RequestMessage};
 pub use response::Response;
 pub use response::*;
-pub use sampler::{SamplingParams, StopTokens, TopLogprob};
+pub use sampler::{GenericLogitsProcessor, SamplingParams, StopTokens, TopLogprob};
 pub use scheduler::{DefaultSchedulerMethod, SchedulerConfig};
 use serde::Serialize;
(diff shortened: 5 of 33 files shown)


Dan-wanna-M:
@EricLBuehler One thing I am pondering is that `kbnf` is stateful, which means to use `Arc<dyn Fn()>` I need to use `Mutex<T>`. This requirement by itself is fine, but I am wondering its implications. For example, if logits processors will be directly shared across threads and executed in parallel, then some very bad stuff probably will happen.

EricLBuehler:
> @EricLBuehler One thing I am pondering is that kbnf is stateful, which means to use Arc<dyn Fn()> I need to use Mutex<T>. This requirement by itself is fine, but I am wondering its implications. For example, if logits processors will be directly shared across threads and executed in parallel, then some very bad stuff probably will happen.

@Dan-wanna-M yes, we parallelize the sampling process. Perhaps we could change the API to accept some generic type which has some method to do the logits processing? Then, you could contain the state in that type in `Mutex`.

EricLBuehler:
@Dan-wanna-M I just updated it so now the API is:
```rust
pub trait CustomLogitsProcessor: Send + Sync {
    /// Logits and sequence context (prompt and generated tokens), returning modified tokens.
    fn apply(&self, logits: &Tensor, context: &[u32]) -> Result<Tensor>;
}
```
In the example, we have a custom struct:
```rust
struct ThresholdLogitsProcessor {
    threshold: f64,
}

impl CustomLogitsProcessor for ThresholdLogitsProcessor {
    fn apply(&self, logits: &Tensor, _context: &[u32]) -> Result<Tensor> {
        // Mask is 1 for true, 0 for false.
        let mask = logits.ge(self.threshold)?;
        logits.broadcast_mul(&mask.to_dtype(logits.dtype())?)
    }
}
```

Would this work if you could put shared state into the custom struct w/ a mutex?

Dan-wanna-M:
> @Dan-wanna-M I just updated it so now the API is:
> 
> ```rust
> pub trait CustomLogitsProcessor: Send + Sync {
>     /// Logits and sequence context (prompt and generated tokens), returning modified tokens.
>     fn apply(&self, logits: &Tensor, context: &[u32]) -> Result<Tensor>;
> }
> ```
> 
> In the example, we have a custom struct:
> 
> ```rust
> struct ThresholdLogitsProcessor {
>     threshold: f64,
> }
> 
> impl CustomLogitsProcessor for ThresholdLogitsProcessor {
>     fn apply(&self, logits: &Tensor, _context: &[u32]) -> Result<Tensor> {
>         // Mask is 1 for true, 0 for false.
>         let mask = logits.ge(self.threshold)?;
>         logits.broadcast_mul(&mask.to_dtype(logits.dtype())?)
>     }
> }
> ```
> 
> Would this work if you could put shared state into the custom struct w/ a mutex?

I spent some time rethinking it and I think as long as one generation only uses one logits processor then it should work(So a batch of N generation will need N separate logits processor)

EricLBuehler:
> > @Dan-wanna-M I just updated it so now the API is:
> > ```rust
> > pub trait CustomLogitsProcessor: Send + Sync {
> >     /// Logits and sequence context (prompt and generated tokens), returning modified tokens.
> >     fn apply(&self, logits: &Tensor, context: &[u32]) -> Result<Tensor>;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In the example, we have a custom struct:
> > ```rust
> > struct ThresholdLogitsProcessor {
> >     threshold: f64,
> > }
> > 
> > impl CustomLogitsProcessor for ThresholdLogitsProcessor {
> >     fn apply(&self, logits: &Tensor, _context: &[u32]) -> Result<Tensor> {
> >         // Mask is 1 for true, 0 for false.
> >         let mask = logits.ge(self.threshold)?;
> >         logits.broadcast_mul(&mask.to_dtype(logits.dtype())?)
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would this work if you could put shared state into the custom struct w/ a mutex?
> 
> I spent some time rethinking it and I think as long as one generation only uses one logits processor then it should work(So a batch of N generation will need N separate logits processor)

Sounds good, and this API implements `CustomLogitsProcessor` for the corresponding closure type! I'll merge this if there are no other problems.

Dan-wanna-M:
> > > @Dan-wanna-M I just updated it so now the API is:
> > > ```rust
> > > pub trait CustomLogitsProcessor: Send + Sync {
> > >     /// Logits and sequence context (prompt and generated tokens), returning modified tokens.
> > >     fn apply(&self, logits: &Tensor, context: &[u32]) -> Result<Tensor>;
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > In the example, we have a custom struct:
> > > ```rust
> > > struct ThresholdLogitsProcessor {
> > >     threshold: f64,
> > > }
> > > 
> > > impl CustomLogitsProcessor for ThresholdLogitsProcessor {
> > >     fn apply(&self, logits: &Tensor, _context: &[u32]) -> Result<Tensor> {
> > >         // Mask is 1 for true, 0 for false.
> > >         let mask = logits.ge(self.threshold)?;
> > >         logits.broadcast_mul(&mask.to_dtype(logits.dtype())?)
> > >     }
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Would this work if you could put shared state into the custom struct w/ a mutex?
> > 
> > 
> > I spent some time rethinking it and I think as long as one generation only uses one logits processor then it should work(So a batch of N generation will need N separate logits processor)
> 
> Sounds good, and this API implements `CustomLogitsProcessor` for the corresponding closure type! I'll merge this if there are no other problems.

Yes it should be mergeable now

EricLBuehler:
Sounds good @Dan-wanna-M! Just merged it.

PR #727, opened 2024-08-30
feat(vision): add support for proper file and data image URLs

Schuwi:
Currently when using vision models, sending image data as base64 in proper URL format (i.e. `"url": "data:image/png;base64,iVBORw0K..."` in JSON request) fails with a 500 error response on the API. I encountered this problem when using [LibreChat](https://github.com/danny-avila/LibreChat) with mistral.rs.

This PR fixes this issue by implementing proper URL parsing.

Backwards-compatibility is maintained by parsing naked file paths (e.g. `"url": "some/file/path.png"`) and naked base64 input (e.g. `"url": "iVBORw0K..."`) the same as before.

The new parser logic uses the `url` and `data-url` crates.
Unit tests have also been added.


EricLBuehler (commented):
Hi @Schuwi! Thanks for the PR, it looks really great.

One thing I noticed is that there is no update to the code in `mistralrs-pyo3`, which also parses URLs ([here](https://github.com/EricLBuehler/mistral.rs/blob/cfc0fa7ffaed0ca71cf86caff4dbd950d08626d8/mistralrs-pyo3/src/lib.rs#L729-L754)). The only problem I see which prevents using the new `parse_image_url` function there is that the pyo3 code isn't async. Perhaps you could duplicate the function there but without async?

EricLBuehler (approved):
Thank you!

PR #771, opened 2024-09-13
Fix Metal build error with seed

EricLBuehler:
Refs EricLBuehler/candle#25.


EricLBuehler:
New rev will be [8a99f7c](https://github.com/EricLBuehler/candle/commit/8a99f7cf31a1d8f175281492eaa7026730067d08)

PR #775, opened 2024-09-15
Reduce CPU usage when idle

scottwey:
Currently, the tight loop in `Engine` causes very high single core CPU usage when idle. This is also not great because this is long-running blocking code running inside of an async task, blocking an async worker entirely. On systems with lower core count, this will probably impact performance fairly negatively.

From my testing, this change dramatically drops CPU usage with minimal impact to performance, although I have not had a chance to benchmark properly.

### DIFF
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -103,6 +103,8 @@ impl Engine {
         let rng = Arc::new(std::sync::Mutex::new(Isaac64Rng::seed_from_u64(SEED)));
         let mut last_completion_ids: Vec<usize> = vec![];
         'lp: loop {
+            tokio::time::sleep(std::time::Duration::from_micros(500)).await;
+
             if matches!(
                 ENGINE_INSTRUCTIONS
                     .lock()
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -125,7 +125,6 @@ impl Engine {
                 }
 
                 else => {
-
                     let scheduled = self.scheduler.schedule();
 
                     match scheduled {
@@ -443,6 +442,8 @@ impl Engine {
                         }
                     }
 
+                    // if there are no more pending requests in the scheduler, yield the current task
+                    // this will mark the task as `Pending` for one runtime tick, the loop will resume on the next tick
                     if self.scheduler.waiting_len() == 0 {
                         tokio::task::yield_now().await;
                     }
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -103,8 +103,6 @@ impl Engine {
         let rng = Arc::new(std::sync::Mutex::new(Isaac64Rng::seed_from_u64(SEED)));
         let mut last_completion_ids: Vec<usize> = vec![];
         'lp: loop {
-            tokio::time::sleep(std::time::Duration::from_micros(500)).await;
-
             if matches!(
                 ENGINE_INSTRUCTIONS
                     .lock()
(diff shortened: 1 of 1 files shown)


mistralrs-core/src/engine/mod.rs
```diff
@@ -103,6 +103,8 @@ impl Engine {
         let rng = Arc::new(std::sync::Mutex::new(Isaac64Rng::seed_from_u64(SEED)));
         let mut last_completion_ids: Vec<usize> = vec![];
         'lp: loop {
+            tokio::time::sleep(std::time::Duration::from_micros(500)).await;
```
EricLBuehler, line 106: I like the idea of not sitting in a hot loop, but I feel like always waiting 0.5ms (essentially wasting performance in the decoding stage) isn't good. Could you modify this to only happen when the scheduler has no waiting/running sequences?

scottwey:
@EricLBuehler I switched to using `yield_now` only when nothing is scheduled instead of sleeping indiscriminately. I also did a bit of refactoring to hopefully make things a bit cleaner.

mistralrs-core/src/engine/mod.rs
```diff
@@ -116,331 +116,340 @@ impl Engine {
...
+                                }
                             }
                         }
                     }
-                }
-            }
 
-            self.scheduler.free_finished_sequence_groups();
+                    if self.scheduler.waiting_len() == 0 {
+                        tokio::task::yield_now().await;
```
EricLBuehler, line 447: I like the idea of the `tokio::select!` addition! I'm just a bit confused about how this function would implement what we want (that is, to not sit in a loop). If I understand correctly, it yields to tokio's runtime, which means we wait for the other arm of the `tokio::select!` (the request recieve arm) to match?

Perhaps you could add a comment here explaining what the logic/flow is?

scottwey: You understand the flow correctly. The reason we no longer sit in the loop anymore is mostly because of the `yield_now`, it basically marks the task as `Pending` for one iteration of Tokio's runtime. On the next iteration, `yield_now` will marked `Ready` and the loop can continue. Yielding doesn't take much time, but is enough to lower CPU usage dramatically.

scottwey: @EricLBuehler Thoughts on this? Happy to close this and rethink if this change doesn't make sense to you.

mistralrs-core/src/engine/mod.rs
```diff
@@ -116,331 +116,341 @@ impl Engine {
...
                             }
                         }
                     }
-                }
-            }
 
-            self.scheduler.free_finished_sequence_groups();
+                    // if there are no more pending requests in the scheduler, yield the current task
+                    // this will mark the task as `Pending` for one runtime tick, the loop will resume on the next tick
+                    if self.scheduler.waiting_len() == 0 {
```
EricLBuehler, line 447:
```suggestion
                    if self.scheduler.waiting_len() == 0 && self.scheduler.running_len() == 0 {
```

EricLBuehler (changes requested):
@scottwey I think after this change and if you could do some testing of it, this should be good to merge!

Finally, if you could drop some rough metrics on CPU usage before vs after it would be great too.

scottwey:
@EricLBuehler 
Thanks for the patience here. I have had 0 time for anything besides work.
I finally had the time to leave this running and use it as a backend for a lot of personal inference and I don't see any obvious issues. Let me know if there are any specific things you want me to be on the lookout for.
In terms of CPU usage, it drops from a single core pegged at 100% down to ~1-2%. Happy to gather more rigorous metrics if you would like. Just let me know.

Otherwise, I will get the conflicts sorted and await further review.

EricLBuehler:
@scottwey thanks you for the testing! Great to see the 100% -> ~1% utilization :).

If you could resolve the conflicts, we can absolutely merge.

PR #785, opened 2024-09-22
Easier and simpler Rust-side API

EricLBuehler:


rickbeeloo:
Maybe the implicit default sampling parameters is not ideal:
https://github.com/EricLBuehler/mistral.rs/blob/c2ff4027824f71d59ebcc6a4aad87f099865a348/mistralrs/src/model.rs#L68.

I think often people would want to change those, like temperature or way of sampling. 

Not really sure what would be more intuitive to have:


**1. On RequestBuilder**
```
  let request = RequestBuilder::new()
            .set_constraint(self.yes_no_constraint.clone())
            .add_message(TextMessageRole::User, prompt);
            .set_temperature(0.1)
 ```
 
 **1. Seperate sampling**
 ```
 let sampling = SamplingParams...{temperature=0.01}
   let request = RequestBuilder::new()
            .set_constraint(self.yes_no_constraint.clone())
            .add_message(TextMessageRole::User, prompt);
            .set_sampling(sampling)
 ```
    
    
  **Opinion?**      
Maybe 1 is "clearner" as almost all (except `logits_bias`, `dry_params`, and `stop_toks` are primitives or options of primitive which can easily be a single function. then it will be like:
```
  let request = RequestBuilder::new()
            .set_constraint(self.yes_no_constraint.clone())
            .add_message(TextMessageRole::User, prompt);
            .set_temperature(0.1)
            .set_max_len(Some(10))
            .frequency_penalty(Some(0.4))
            ...
 ```
 But for 2 it might be clearer that "Sampling" is changed, or we should prefix in option 1 like `set_sample_temperature()`, `set_sample_top_k()`.

Not sure, what do you think? @EricLBuehler. I could PR but guess this is very minimal changes to what you already did

EricLBuehler:
Hi @rickbeeloo, thanks for pointing this out! Yes, we should absolutely add configurability for this. I just opened #790 which is a combination of options 1 and 2, if you could review that would be great!

PR #787, opened 2024-09-22
Deduplicate RoPE caches

EricLBuehler:
Reduces memory consumption for Phi 3.5 mini, ISQ=Q4K from 4000MB to 2500MB.


rickbeeloo:
👀👀

PR #790, opened 2024-09-23
Rust API for sampling

EricLBuehler:


mistralrs-core/src/sampler.rs
```diff
@@ -63,6 +63,29 @@ impl Default for SamplingParams {
     }
 }
 
+impl SamplingParams {
+    /// This sets up the parameters so that there is:
+    /// - No temperature, topk, topp, minp
+    /// - No penalties, stop tokens, or logit bias
+    /// - No maximum length
+    pub fn deterministic() -> Self {
```
rickbeeloo, line 71: This has the same settings as default: https://github.com/EricLBuehler/mistral.rs/blob/c2ff4027824f71d59ebcc6a4aad87f099865a348/mistralrs-core/src/sampler.rs#L48C4-L63C6.

I think it's confusing to have both `::default` and `::deterministic` if they are the same

EricLBuehler: That makes sense, the Defualt impl was a bit unclear. I removed it.

EricLBuehler:
> Based on your comment line it seems that the tags should be different for llava and phiv but they are the same in the code:

Thanks for finding this! Yes, it was incorrect - it is fixed now.

rickbeeloo:
Perfect, this is easy to use, thanks!

PR #815, opened 2024-10-02
Integrate KBNF grammar

EricLBuehler:


examples/server/grammar.py
EricLBuehler, line 1: Are the grammars here correct? Perhaps we can have a simpler example.

mistralrs-core/src/pipeline/sampling.rs
```diff
@@ -324,76 +325,141 @@ pub async fn sample_sequence(
...
+                    if add_to_trie {
+                        let SequenceRecognizer::Kbnf(ref mut kbnf) = seq.recognizer else {
+                            unreachable!()
+                        };
+                        kbnf.add_token(second_sampled.token)?;
+                    }
+                    Ok(second_sampled)
+                }
+                KbnfGrammarBias::FinishedGeneration => {
+                    todo!()
```
EricLBuehler, line 458: TODO: Need to mark the sequence as done.

mistralrs-core/src/kbnf.rs
EricLBuehler, line 1: I was wondering if this file generally looks correct? It is paired with the sampling routines in `sampling.rs` and initialized in `engine/mod.rs`.

EricLBuehler (commented):
@Dan-wanna-M I was wondering if you could perhaps review some of this code and take a look at some of the review comments I made?

examples/server/grammar.py
```diff
@@ -0,0 +1,99 @@
+from openai import OpenAI
+
+client = OpenAI(api_key="foobar", base_url="http://localhost:1234/v1/")
+
+JSON_KBNF = '''
+(* JSON Grammar *)
+
+(* JSON text must contain a single JSON value *)
+start       = value ;
```
Dan-wanna-M, line 9: Should be `start ::= value;`

examples/server/grammar.py
```diff
@@ -0,0 +1,99 @@
...
+value      ::= object 
+           | array 
+           | string 
+           | number 
+           | "true" 
+           | "false" 
+           | "null" ;
+
+(* A JSON object is a collection of key/value pairs enclosed in curly braces *)
+object     ::= "{" [ members ] "}" ;
```
Dan-wanna-M, line 21: You can use both `[members]` and `members?` to express optional nonterminal

examples/server/grammar.py
```diff
@@ -0,0 +1,99 @@
...
+members    ::= pair { "," pair } ;
```
Dan-wanna-M, line 22: You can use both `{members}` and `members*` to express 0 or infinite repetition.

examples/server/grammar.py
```diff
@@ -0,0 +1,99 @@
...
+pair       ::= string ":" value ;
+
+(* A JSON array is an ordered list of values enclosed in square brackets *)
+array      ::= "[" [ elements ] "]" ;
+elements   ::= value { "," value } ;
+
+(* A JSON string is a sequence of Unicode characters enclosed in double quotes *)
+string     ::= "\"" { character } "\"" ;
```
Dan-wanna-M, line 30: You can use regex in kbnf, see https://docs.rs/kbnf/latest/kbnf/#regular-expression.

examples/server/grammar.py
Dan-wanna-M: They look mostly correct to me; I have added some comments that can make the grammar more familiar and/or simpler to general users. In terms of simpler example, I think we can use something within regular grammar's capacity but can be more clearly written in kbnf. I think markdown list and the kbnf for a concrete json schema both qualify. Which one do you prefer, or you would like some examples in other areas?

examples/server/grammar.py
```diff
@@ -0,0 +1,99 @@
+from openai import OpenAI
+
+client = OpenAI(api_key="foobar", base_url="http://localhost:1234/v1/")
+
+JSON_KBNF = '''
```
Dan-wanna-M, line 5: You may also want to take a look at https://github.com/Dan-wanna-M/formatron/blob/master/src/formatron/formats/json.py where I put a json grammar definition. It also contains whitespace pattern that makes llm generate better json.

mistralrs-core/src/kbnf.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// The token sampled with the bias can be added to the sequence.
+    Resample { new_logits: Tensor },
+    /// Generation was finished, the token can be added to the sequence,
+    /// but no more generation is necessary.
+    FinishedGeneration,
+}
+
+impl KbnfGrammar {
+    pub fn new(grammar: &str, tokenizer: &Tokenizer) -> Result<Self> {
+        let tokenizer_vocab = tokenizer.get_vocab(true);
```
Dan-wanna-M, line 28: I think we still need to "unpreprocess" the vocabulary or the control ASCII characters(like `\n`) and all non-ASCII characters will not be recognized correctly.

mistralrs-core/src/kbnf.rs
```diff
@@ -0,0 +1,103 @@
...
+                Ok(KbnfGrammarBias::Accepted)
+            }
+            Err(AcceptTokenError::Rejected) => {
+                if add_to_trie {
+                    self.engine.compute_allowed_token_ids();
+                }
+                let mut bias = vec![0f32; self.vocab_size];
+                match self.engine.mask_logits(&mut bias) {
+                    Ok(()) => {
+                        let new_logits = (logits.to_device(&Device::Cpu)?.to_dtype(DType::F32)?
```
Dan-wanna-M, line 66: It might not be a good idea to move logits back to CPU due to potential CUDA synchronization issues. I think we can benchmark this to check latency. If it does bring too much latency, we can follow a strategy like [this file](https://github.com/Dan-wanna-M/kbnf/blob/master/python/kbnf/engine.py) which essentially only moves the indices tensor(tends to be smaller than the whole logits) to GPU and operate there.

mistralrs-core/src/kbnf.rs
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }
+            Ok(AcceptTokenResult::Finished) | Err(AcceptTokenError::Finished) => {
+                Ok(KbnfGrammarBias::FinishedGeneration)
+            }
+            Err(AcceptTokenError::UnknownTokenID) => candle_core::bail!("Unknown token ID {tok}"),
+        }
+    }
+
+    /// Add a token, also to the trie.
```
Dan-wanna-M, line 83: technically speaking the cache is a hashmap, not a trie.

Dan-wanna-M (changes requested):
@EricLBuehler I have reviewed the PR and suggested some changes.

PR #873, opened 2024-10-21
fix Half-Quadratic Quantization and Dequantization on CPU

haricot:
This confirms that `test_bitpack` is running solely on non-CPU hardware. To address this, we could implement a fix by ensuring contiguous data slices.

### DIFF
--- a/mistralrs-quant/src/hqq/hqq_cpu.rs
+++ b/mistralrs-quant/src/hqq/hqq_cpu.rs
@@ -65,8 +65,9 @@ pub(crate) struct Dequant4Bit {
 }
 
 impl Dequant4Bit {
-    fn dequantize<T: WithDType>(&self, w: &[u8], s: &[T], z: &[T]) -> Vec<T> {
-        let mut out = Vec::with_capacity(w.len());
+    fn dequantize<T: WithDType + Default>(&self, w: &[u8], s: &[T], z: &[T]) -> Vec<T> {
+        let output_size = w.len() * 2;
+        let mut out = vec![T::default(); output_size];
         for (i, w) in w.iter().enumerate() {
             let j = i % self.w;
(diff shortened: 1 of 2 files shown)


EricLBuehler (changes requested):
Hi @haricot! Thanks for the PR. Can you please update it so it also tests 8 bit quantization? Thanks!

EricLBuehler:
@haricot were you planning on implementing HQQ for non-CUDA devices in this PR? The name seems to indicate so, I was just wondering!

haricot:
Indeed the title is not correct. It is rather: fix Half-Quadratic Quantization and Dequantization on CPU, because I am not sure I can optimize it.
My first goal was to make the quantization work on my device. In fact, I could not quantize the models, I got OOM.

With your models quantized on GPU and 8 VRAM, it worked correctly with this command:
`cargo run -r --features cuda -- --pa-gpu-mem-usage 0.5 -i plain -m '/path/' --from-uqff /model/llm-hqq4.uqff
`

With the model quantized on CPU, the inferences on CPU or GPU, I got inconsistent texts. After correct it , I realized that the inferences on CPU with the quantized models were not optimal. I tried to profile to see which step should be optimized and implement a SIMD version of the dequantize function but it didn't turn out to be enough, it seems necessary to add data alignment with SIMD.
I wonder if it is only about optimizing the dequantize function or adding a KV cache aligned system or something else?

About Half-Quadratic Quantization on GPU and possible offload CPU to avoid OOM:
I would like also to propose a method when a max limit of available VRAM is reached, the tensor is quantized on CPU. This is relevant?

EricLBuehler (approved):
Hi @haricot! All tests pass on CPU (the target of this PR) and the changes look good. Merging now, thanks for the contribution.

PR #874, opened 2024-10-22
add marlin as cargo feature

haricot:
This ensures broader accessibility to Mistral.rs across different hardware configurations by enabling its use even on systems with a CUDA compute capability less than 8.0.


haricot:
Closed in favor #878

PR #895, opened 2024-10-29
Update Dockerfile

Reckon-11:
Combined Commands:
Combined multiple COPY and RUN commands where possible to reduce layer count.

Reduced Layer Creation:
Combined the installation and cleanup commands in each stage to minimize layers, which helps reduce image size.

Runtime Stage:
Installed only the essential runtime dependencies (libomp-dev, ca-certificates, libssl-dev, curl) and minimized the final image by using debian:bookworm-slim in the runtime stage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,41 @@
+# Stage 1: Build environment
 FROM rust:latest AS builder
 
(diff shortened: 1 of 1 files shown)


EricLBuehler (approved):
Thank you!

ShivamSphn (commented):
good

PR #899, opened 2024-11-06
use llguidance library for constraints (including json schemas)

### LINKED ISSUE #754: AICI -> llguidance?
I noticed you guys forked a bunch of controller code from AICI for your constraints. I think you might be interested in https://github.com/microsoft/llguidance - it implements a more general constraint parser and is significantly faster. It can be also easily consumed as a Rust crate.

Grammars for llguidance can be generated from the Python Guidance https://github.com/guidance-ai/guidance (in particular there is code there that translates a JSON schema into Guidance and then llguidance; it shouldn't be too hard to port that into Rust but I have not gotten around to do that yet)

I also did a quick PoC in TypeScript at https://github.com/mmoskal/guidance-ts

I expect more client libraries to emerge.

Of course it's easy to just write a grammar with a single regex if this is what you're after (llguidance is using https://github.com/microsoft/derivre for regexes/lexers).

LLguidance is [deployed as preview](https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/discover-the-new-multi-lingual-high-quality-phi-3-5-slms/ba-p/4225280) in Azure for Phi-3.5-mini

mmoskal:
Fixes #754 

This imports the `llguidance` library for dealing with constraints (regexp, json schema, CFGs in lark syntax) and removes the `aici` submodule which was implementing regexps and CFGs (in yacc syntax) before. The code for both is related (I wrote the original versions of both).

This is WIP - putting it out there for comments about how likely it is to be merged.

TODO
- [x] remove testing scriptgs (`scripst/*.sh`)
- [x] remove `aici` folder
- [x] update python APIs to accept "json_schema"
- [x] update examples


mmoskal:
I probably won't have time for it this week, but hopefully the one after that.

EricLBuehler:
@mmoskal friendly ping :) Let me know if I can help!

mmoskal:
It seems I won't have much time in the next two-three weeks. If anyone else is willing to take on the TODOs that would be great!

I'll be extending the llguidance APIs to be able to perform multithreaded mask computation in background (one of the things I will be busy with). Enabling this will get the constraint computation out of the hot path of sampling - but that should be a one-line change here.

EricLBuehler:
@mmoskal that sounds exciting! I tried to recreate this PR to quickly merge it and get something working but it seems many of the APIs have changed since the code here was written. Looking forward to the enhancements and hopefully getting this merged!

mmoskal:
@EricLBuehler should be good for review!

examples/server/lark.py
EricLBuehler, line 1: Wow, this is really cool! I also saw a JSON yacc grammar [here](https://github.com/lark-parser/lark/blob/master/examples/json_parser.py#L13-L33) which might be nice to include as a seperate example, too?

mistralrs-core/src/request.rs
```diff
@@ -13,11 +13,15 @@ use crate::{
...
+
 #[derive(Clone)]
-/// Control the constraint with Regex or Yacc.
+/// Control the constraint with llguidance.
 pub enum Constraint {
     Regex(String),
-    Yacc(String),
+    Lark(String),
+    JsonSchema(serde_json::Value),
+    Llguidance(LlguidanceGrammar),
```
EricLBuehler, line 24: Can we export the `llguidance` and (if necessary) `toktrie_hf_tokenizers` crates from`mistralrs-core`? I think this will help users constructing a `LlguidanceGrammar`.

EricLBuehler (changes requested):
Thanks, this looks really great! I was wondering if you could please add more examples, namely:

- Server: add an example of JSON schema and maybe a llguidance grammar
- Python: similarly, add examples of Lark, JSON schema, and maybe a llguidance grammar
- Rust: similarly, add examples of Lark, JSON schema, and maybe a llguidance grammar

Otherwise, the code looks great. I left a few comments, and it looks like formatting and clippy are failing in CI.

mistralrs-core/src/request.rs
mmoskal: Only `llguidance` I think.

examples/server/lark.py
mmoskal: Added, but I'm not sure this is the right place to point people to - getting a grammar for specific JSON is somewhat tricky (for example, you need to care about quoting in strings).

I have also added example in Llguidance format that combine Lark with JSON Schema - that's a more robust way of doing this.

mmoskal:
Added samples in rust, python and for server; fixed clippy and fmt. Should be good to go!

EricLBuehler (approved):
Thank you so much! This is really amazing and looks great.

I'll keep my eye on llguidance's development for new features!

mmoskal:
Thanks for merging!

There is a few areas where the integration can be improved, I filed #963, #964, #965 to track.

PR #914, opened 2024-11-17
Metal: Use mtl resource shared to avoid one copy

EricLBuehler:


EricLBuehler:
Small optimization. Docs: https://developer.apple.com/documentation/metal/mtlstoragemode/shared

We were using: https://developer.apple.com/documentation/metal/mtlstoragemode/managed

PR #957, opened 2024-12-01
set minimum rustc version to 1.82

mmoskal:
The `is_none_or` function was introduced in 1.82. 

BTW, maybe we need 1.83?


EricLBuehler (approved):
Thank you!

EricLBuehler:
Is there something which is requiring 1.83? 

It looks like this change of using `is_none_or` was introduced by using Clippy for 1.83!

Issue #764, opened 2024-09-10
v0.3.0 missing binaries 


aretrace:
no pre-built `mistralrs-server` binaries under assets in [0.3.0 github release](https://github.com/EricLBuehler/mistral.rs/releases/tag/v0.3.0)
very sad 😢

EricLBuehler:
@aretrace I will include the binaries in the 0.3.2 release in a few days!

aretrace:
@EricLBuehler how were you generating binaries for previous releases? Do you think using something like GH Actions would work to automate the process?

BafS:
I was also confused about this, I didn't find binaries for version 0.3.2

PR #973, opened 2024-12-05
Remove outdated binaries mention in the readme

BafS:
Remove outdated binaries mention, as they are unfortunately not published anymore.

See https://github.com/EricLBuehler/mistral.rs/issues/764


EricLBuehler (approved):
Thanks! We may investigate binaries in the next release now that the PyPI wheel generation method is stable!

PR #974, opened 2024-12-06
Improve error handling

cdoko:
Currently, when the `bos_token_id` or `eos_token_id` field is missing from `generation_config.json`, the error message is "Error(`missing field `bos_token_id`", line: 4, column: 1)", which does not provide clear context about the source of the issue.

This change indicates the problem is in the `generation_config.json` file.


EricLBuehler (approved):
Thank you!

PR #979, opened 2024-12-08
Add None check to prevent panic in evict_all_to_cpu in prefix_cacher.rs

cdoko:
This check is already present in the `evict_to_cpu` function and was missing in `evict_all_to_cpu`.

Note: I'm not entirely sure if the comment is relevant here, I just copied it from the `evict_to_cpu` function - still getting familiar with the code


EricLBuehler (approved):
Thank you! Yes, this comment is relevant here (the "normal cache" is the preallocated, faster/newer cache implementation than the original "full cache", supporting it here would take a bit of work and probably come with PagedAttention support)

PR #982, opened 2024-12-10
Fail fast on TcpListener bind errors

cdoko:
Moves the creation of the `TcpListener` to the top of the `main` function to check for errors when binding to the address before loading the model. This avoids loading the model, which can take several minutes, when a bind error occurs.


EricLBuehler (approved):
Thank you!

PR #994, opened 2024-12-16
Partial Fix for Sliding Window Attention

cdoko:
Currently, in models like Phi3 and Mistral, with paged attention disabled, sliding window attention fails when the input sequence length exceeds the sliding window size, due to a "start > dim_len" error caused by the mask tensor being narrowed using its original length. 

An initial fix was updating the mask length after the first narrow operation, but this brings up another error where the mask's dimension was one higher than the corresponding dimension in the key and value tensors, leading to a shape mismatch in the `naive_sdpa` function.

The issue was resolved by removing the concatenation of the mask tensor, resulting in coherent outputs for non-flash attention, although flash attention remains non-functional. 

Any feedback or review of this change would be greatly appreciated!


EricLBuehler (approved):
Thank you!

PR #997, opened 2024-12-19
Add `--cpu` flag to `mistralrs-server`

cdoko:
Convenience flag to allow users to run the model purely on the CPU. If you have any feedback or would like to request changes, please let me know!

### DIFF
--- a/mistralrs-server/src/main.rs
+++ b/mistralrs-server/src/main.rs
@@ -473,8 +473,7 @@ async fn main() -> Result<()> {
         .with_opt_log(args.log)
         .with_truncate_sequence(args.truncate_sequence)
         .with_no_kv_cache(args.no_kv_cache)
-        .with_prefix_cache_n(args.prefix_cache_n)
-        .with_gemm_full_precision_f16(args.cpu);
+        .with_prefix_cache_n(args.prefix_cache_n);
 
     if args.interactive_mode {
         interactive_mode(builder.build(), args.throughput_log).await;
--- a/mistralrs-server/src/main.rs
+++ b/mistralrs-server/src/main.rs
@@ -149,6 +149,10 @@ struct Args {
     /// Number of tokens to batch the prompt step into. This can help with OOM errors when in the prompt step, but reduces performance.
     #[arg(long = "prompt-batchsize")]
     prompt_batchsize: Option<usize>,
+
+    /// Use CPU only
+    #[arg(long)]
+    cpu: bool,
 }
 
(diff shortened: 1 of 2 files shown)


mistralrs-server/src/main.rs
```diff
@@ -464,7 +473,8 @@ async fn main() -> Result<()> {
         .with_opt_log(args.log)
         .with_truncate_sequence(args.truncate_sequence)
         .with_no_kv_cache(args.no_kv_cache)
-        .with_prefix_cache_n(args.prefix_cache_n);
+        .with_prefix_cache_n(args.prefix_cache_n)
+        .with_gemm_full_precision_f16(args.cpu);
```
EricLBuehler, line 477: It seems this method hasn't been documented well - do you think you could update the docs for `with_gemm_full_precision_f16`?

Something like:
"This setting is only applicable on CUDA. If set to false or not specified, this setting enables f16/bf16 reduced precision matmul for GPUs which support it. If set to true, this setting has no effect"

EricLBuehler (changes requested):

EricLBuehler:
The PR looks great, my only comment was enhancing the docs for one method you used. I think that perhaps you should remove its usage here, as the default is what is intended to be used?

EricLBuehler:
@cdoko 

> By setting gemm_full_precision_f16 to true when --cpu is used, I aimed to avoid unnecessary GEMM configuration. Please let me know if I'm misunderstanding the code.

I think it'd be best to not include this - since the automatic behavior is probably sufficient for now, I think it would be fine.

> On a separate note, I had a question regarding the PagedAttention configuration. When using device mapping, I see a message saying "Device mapping or device topology and PagedAttention are incompatible, disabling PagedAttention." Is PagedAttention inherently incompatible with multi-GPU setups, or is it just not yet implemented in Mistral.rs? If it's the latter, what would it take to support PagedAttention with device mapping or multi-GPU setups in the future?

This simply hasn't been implemented yet. If you'd like to, I think all that would be necessary is to modify how [the Paged Attention KV cache](https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/paged_attention/cache_engine.rs) is allocated by providing the device mapping. I'd be happy to accept such a change!

EricLBuehler (changes requested):
Thank you so much! I noticed that the formatting check failed, can you please run `cargo fmt --all`?

EricLBuehler (approved):
Thank you!

PR #1001, opened 2024-12-21
Metal PagedAttention support

EricLBuehler:


guoqingbao:
Perhaps we need to retain the **vec** type implemented in the CUDA version for the paged attention kernel. While I understand that the current implementation is revised to use primitive types (f32, bf16, and f16) to simplify FMA calculations, this approach misaligns the workflow with the CUDA codebase (e.g., the **k_vecs** issue).

CUDA vec type:

``` c++
// A vector type to store Q, K, V elements.
template<typename T, int VEC_SIZE>
struct Vec {};

// A vector type to store FP32 accumulators.
template<typename T>
struct FloatVec {};
```

CUDA dot kernel based on vec type:

``` c++
// Q*K^T operation.
template<int THREAD_GROUP_SIZE, typename Vec, int N>
inline __device__ float qk_dot_(const Vec (&q)[N], const Vec (&k)[N]) {
  using A_vec = typename FloatVec<Vec>::Type;
  // Compute the parallel products for Q*K^T (treat vector lanes separately).
  A_vec qk_vec = mul<A_vec, Vec, Vec>(q[0], k[0]);
#pragma unroll
  for (int ii = 1; ii < N; ++ii) {
    qk_vec = fma(q[ii], k[ii], qk_vec);
  }

  // Finalize the reduction across lanes.
  float qk = sum(qk_vec);
#pragma unroll
  for (int mask = THREAD_GROUP_SIZE / 2; mask >= 1; mask /= 2) {
    qk += VLLM_SHFL_XOR_SYNC(qk, mask);
  }
  return qk;
}
```

guoqingbao:
> > Perhaps we need to retain the vec type implemented in the CUDA version for the paged attention kernel. While I understand that the current implementation is revised to use primitive types (f32, bf16, and f16) to simplify FMA calculations, this approach misaligns the workflow with the CUDA codebase (e.g., the k_vecs issue).
> 
> Thanks for pointing this out. It turns out, the reason why it wasn't working was that the q_vecs were not stored in the threadgroup memory (equiv. of CUDA `__shared__`), I changed this and it works now!
> 
> However, we can certainly benefit from using the Vec types for performance reasons. I'll take a look at adding those.

That's great! Maybe there are some CUDA equivalent fma implementations in metal for performance benefits. Porting complex CUDA kernel like paged attention to metal with few lines of tweaks, it works, which is beyond my mind 🧠!

guoqingbao:
Hi Eric @EricLBuehler I have made further improvements but don't have permission to push to this branch.

PR #1003, opened 2024-12-25
Fix bug for kernel cache & Phi3 inference

guoqingbao:
The metal kernels need to be loaded with load_source in each of the function callls in the current implementation because there is no global kernel cache, this PR fix the problem.

### DIFF
--- a/mistralrs-core/src/lib.rs
+++ b/mistralrs-core/src/lib.rs
@@ -1,5 +1,4 @@
 #![deny(clippy::cast_possible_truncation, clippy::cast_precision_loss)]
-#![feature(is_none_or)]
 use candle_core::Device;
 use cublaslt::setup_cublas_lt_wrapper;
 use engine::Engine;
--- a/mistralrs-core/src/lib.rs
+++ b/mistralrs-core/src/lib.rs
@@ -1,5 +1,5 @@
 #![deny(clippy::cast_possible_truncation, clippy::cast_precision_loss)]
-
+#![feature(is_none_or)]
 use candle_core::Device;
 use cublaslt::setup_cublas_lt_wrapper;
 use engine::Engine;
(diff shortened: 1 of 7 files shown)


mistralrs-core/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![deny(clippy::cast_possible_truncation, clippy::cast_precision_loss)]
-
+#![feature(is_none_or)]
```
EricLBuehler, line 2: Can you please remove this addition? It requires the nightly compiler, not sure if this was a leftover.

mistralrs-paged-attn/src/metal/kernels/mod.rs
```diff
@@ -51,13 +51,19 @@ pub struct Kernels {
...
+static INIT: Once = Once::new();
+pub(crate) static mut G_KERNEL: Option<Kernels> = None;
 
 impl Kernels {
+    pub fn default() -> &'static Option<Self> {
+        unsafe {
+            INIT.call_once(|| {
+                G_KERNEL = Some(Self::new());
+            });
+            &G_KERNEL
```
EricLBuehler, line 63: On Rust 1.83, I'm getting: `warning: creating a shared reference to mutable static is discouraged`.

Perhaps we can make it a Mutex?

EricLBuehler (changes requested):
Thanks for the PR! I made a few comments.

mistralrs-core/src/lib.rs
guoqingbao: > Can you please remove this addition? It requires the nightly compiler, not sure if this was a leftover.

Yes, I will remove this. I use nightly compiler because program compiled by the stable one has conflicts with lldb debugger on my Mac (unable to show concrete values for vec variables).

mistralrs-paged-attn/src/metal/kernels/mod.rs
guoqingbao: Warning fixed with once_cell.

EricLBuehler (approved):
Thank you!

PR #1009, opened 2024-12-27
Support BF16 kvcache, rope and attentions for inference of GGUF/GGML models

guoqingbao:
The current approach for inference of GGUF/GGML models requires the input data type to be `F32` due to limitations in kernel implementations for quantized matmul, which necessitate an `F32` input and quantized weights. This limitation causes side effects, making attention mechanisms (including SDPA, paged attention, and flash attention) and the KV cache also to be `F32`.  

This PR addresses the issue by allowing the data type to be specified for GGUF/GGML models and casting the `F32` results of quantized matmuls to `BF16`. This casting applies to subsequent attention operations, rotary embeddings, and KV cache storage. As a result, KV cache usage is reduced by half, and inference performance is improved due to the use of `BF16` for matmuls and attentions.

### Example

On Apple Silicon (`Metal`), run the following example with throughput logging enabled, paged attention settings (limiting KV cache to a maximum of 4GB), and data type set to `BF16` (for KV cache, rotary embeddings, and attentions):

```bash
cargo build --release --features metal
./target/release/mistralrs-server -i --throughput --paged-attn --pa-gpu-mem 4096 gguf --dtype bf16 -m /Users/Downloads/ -f Phi-3.5-mini-instruct-Q4_K_M.gguf
```

You may also run without paged attention:
```bash
./target/release/mistralrs-server -i --throughput gguf --dtype bf16 -m /Users/Downloads/ -f Phi-3.5-mini-instruct-Q4_K_M.gguf
```

### Notes

- For quantized GGUF/GGML models, you may specify the data type as either `f32` or `bf16`. 
- The use of `f16` is not recommended due to its lower precision, which can adversely impact the precision of quantized inference.

### DIFF
--- a/mistralrs-core/src/model_loader.rs
+++ b/mistralrs-core/src/model_loader.rs
@@ -89,8 +89,9 @@ pub fn get_model_dtype(model: &ModelSelected) -> anyhow::Result<ModelDType> {
         | ModelSelected::GGML { dtype, .. }
         | ModelSelected::GGUF { dtype, .. }
         | ModelSelected::XLoraGGUF { dtype, .. }
-        | ModelSelected::XLoraGGML { dtype, .. } => Ok(*dtype),
-        ModelSelected::LoraGGUF { .. } | ModelSelected::LoraGGML { .. } => Ok(ModelDType::Auto),
+        | ModelSelected::XLoraGGML { dtype, .. }
+        | ModelSelected::LoraGGUF { dtype, .. }
+        | ModelSelected::LoraGGML { dtype, .. } => Ok(*dtype),
         ModelSelected::Toml { file } => {
             let selector: TomlSelector = toml::from_str(
                 &fs::read_to_string(file.clone())
@@ -276,6 +277,7 @@ fn loader_from_model_selected(args: LoaderBuilder) -> anyhow::Result<Box<dyn Loa
             adapters_model_id,
             order,
             topology,
+            ..
         } => GGUFLoaderBuilder::new(
             args.chat_template,
             tok_model_id,
@@ -363,6 +365,7 @@ fn loader_from_model_selected(args: LoaderBuilder) -> anyhow::Result<Box<dyn Loa
             order,
             gqa,
             topology,
+            ..
         } => GGMLLoaderBuilder::new(
             GGMLSpecificConfig {
                 gqa,
--- a/mistralrs-core/src/model_loader.rs
+++ b/mistralrs-core/src/model_loader.rs
@@ -87,11 +87,10 @@ pub fn get_model_dtype(model: &ModelSelected) -> anyhow::Result<ModelDType> {
         | ModelSelected::VisionPlain { dtype, .. }
         | ModelSelected::DiffusionPlain { dtype, .. }
         | ModelSelected::GGML { dtype, .. }
-        | ModelSelected::GGUF { dtype, .. } => Ok(*dtype),
-        ModelSelected::LoraGGUF { .. }
-        | ModelSelected::LoraGGML { .. }
-        | ModelSelected::XLoraGGUF { .. }
-        | ModelSelected::XLoraGGML { .. } => Ok(ModelDType::Auto),
+        | ModelSelected::GGUF { dtype, .. }
+        | ModelSelected::XLoraGGUF { dtype, .. }
+        | ModelSelected::XLoraGGML { dtype, .. } => Ok(*dtype),
+        ModelSelected::LoraGGUF { .. } | ModelSelected::LoraGGML { .. } => Ok(ModelDType::Auto),
         ModelSelected::Toml { file } => {
             let selector: TomlSelector = toml::from_str(
                 &fs::read_to_string(file.clone())
@@ -245,6 +244,7 @@ fn loader_from_model_selected(args: LoaderBuilder) -> anyhow::Result<Box<dyn Loa
             order,
             tgt_non_granular_index,
             topology,
+            ..
         } => GGUFLoaderBuilder::new(
             args.chat_template,
             tok_model_id,
@@ -330,6 +330,7 @@ fn loader_from_model_selected(args: LoaderBuilder) -> anyhow::Result<Box<dyn Loa
             tgt_non_granular_index,
             gqa,
             topology,
+            ..
         } => GGMLLoaderBuilder::new(
             GGMLSpecificConfig {
                 gqa,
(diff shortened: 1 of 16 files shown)


mistralrs-core/src/model_loader.rs
```diff
@@ -85,10 +85,10 @@ pub fn get_model_dtype(model: &ModelSelected) -> anyhow::Result<ModelDType> {
...
         | ModelSelected::XLora { dtype, .. }
         | ModelSelected::VisionPlain { dtype, .. }
-        | ModelSelected::DiffusionPlain { dtype, .. } => Ok(*dtype),
-        ModelSelected::GGUF { .. }
-        | ModelSelected::LoraGGUF { .. }
-        | ModelSelected::GGML { .. }
+        | ModelSelected::DiffusionPlain { dtype, .. }
+        | ModelSelected::GGML { dtype, .. }
+        | ModelSelected::GGUF { dtype, .. } => Ok(*dtype),
+        ModelSelected::LoraGGUF { .. }
```
EricLBuehler, line 91: It appears that you also updated the adapter ggml/gguf models to support setting the type - but I think that is not refleced here?

mistralrs-core/src/toml_selector.rs
```diff
@@ -366,10 +374,10 @@ pub fn get_toml_selected_model_dtype(model: &TomlSelector) -> ModelDType {
...
         | TomlModelSelected::Lora { dtype, .. }
         | TomlModelSelected::XLora { dtype, .. }
-        | TomlModelSelected::VisionPlain { dtype, .. } => dtype,
-        TomlModelSelected::GGUF { .. }
-        | TomlModelSelected::LoraGGUF { .. }
-        | TomlModelSelected::GGML { .. }
+        | TomlModelSelected::VisionPlain { dtype, .. }
+        | TomlModelSelected::GGUF { dtype, .. }
+        | TomlModelSelected::GGML { dtype, .. } => dtype,
+        TomlModelSelected::LoraGGUF { .. }
```
EricLBuehler, line 380: Same comment as above:
> It appears that you also updated the adapter ggml/gguf models to support setting the type - but I think that is not refleced here?

mistralrs-core/src/pipeline/paths.rs
```diff
@@ -381,12 +382,17 @@ pub(crate) fn get_chat_template(
     } else {
         panic!("Expected chat template file to end with .json, or you can specify a tokenizer model ID to load the chat template there. If you are running a GGUF model, it probably does not contain a chat template.");
     };
-
     let mut template: ChatTemplate = match chat_template_ovrd {
         Some(chat_template) => {
             // In this case the override chat template is being used. The user must add the bos/eos/unk toks themselves.
             info!("Using literal chat template.");
             let mut template = ChatTemplate::default();
+            if chat_template.contains("<|end|>") {
```
EricLBuehler, line 390: I think we should avoid such hardcoded cases generally, but was there a bug being fixed here?

mistralrs-server/src/main.rs
```diff
@@ -143,7 +143,7 @@ struct Args {
     no_paged_attn: bool,
 
     /// Enable PagedAttention on Metal. Because PagedAttention is already enabled on CUDA, this is only applicable on Metal.
-    #[arg(long = "no-paged-attn", default_value_t = false)]
+    #[arg(long = "paged-attn", default_value_t = false)]
```
EricLBuehler, line 146: Thanks for catching this!

EricLBuehler (changes requested):
Thanks for the PR! I made a few comments.

mistralrs-core/src/pipeline/paths.rs
guoqingbao: > I think we should avoid such hardcoded cases generally, but was there a bug being fixed here?

Yes, it fixes a bug related to running Phi3 GGUF models, where the end-of-sequence token ("<|end|>") is not properly parsed. If there is no external `chat_template.json` or `config.json`, we need to parse both the BOS and EOS tokens from the GGUF file (the literal chat template). I think there's a more elegant way to handle this situation, and I'm willing to replace the hardcoded one.

mistralrs-core/src/model_loader.rs
guoqingbao: > It appears that you also updated the adapter ggml/gguf models to support setting the type - but I think that is not refleced here?

I’ve made revisions to pass the dtype to the Xlora GGUF/GGML models. However, I’m not sure how this applies to the LoRA model, as I didn’t find an explicit definition for a quantized model in LoRA. Therefore, I’ve left it in its original version.

mistralrs-core/src/toml_selector.rs
guoqingbao: > Same comment as above:
> 
> > It appears that you also updated the adapter ggml/gguf models to support setting the type - but I think that is not refleced here?

Refer to previous fix.

mistralrs-core/src/pipeline/paths.rs
guoqingbao: I think it's better to raise an issue #1012

mistralrs-core/src/model_loader.rs
EricLBuehler: Sorry for the confusion - the X-LoRA and LoRA models are the same implementation, so the updates apply to "both".

mistralrs-core/src/toml_selector.rs
EricLBuehler: > Sorry for the confusion - the X-LoRA and LoRA models are the same implementation, so the updates apply to "both".

mistralrs-core/src/pipeline/paths.rs
EricLBuehler: Sounds good. I'll take a look at that issue, we can make a fix outside of this PR!

mistralrs-core/src/model_loader.rs
guoqingbao: > Sorry for the confusion - the X-LoRA and LoRA models are the same implementation, so the updates apply to "both".

This have been resolved.

mistralrs-core/src/toml_selector.rs
guoqingbao: > > Sorry for the confusion - the X-LoRA and LoRA models are the same implementation, so the updates apply to "both".

This have been resolved.

EricLBuehler (approved):
Thank you!

PR #1011, opened 2024-12-28
Support device mapping for Paged Attention

cdoko:
Added support for device mapping in Paged Attention by passing the device list from the mapper to the cache engine. Manual moving was required for certain unmapped tensors in the paged attention forward pass. I have tested the device mapping support on several text models and it appears to be functional. 

Memory allocation is currently calculated as if all memory is available on a single device, resulting in that memory being split across devices; ideally, we should calculate available memory per GPU. If this PR is fine, I will work on this next.

Additionally, this feature currently only supports GPU devices due to an error when attempting to use reshape_and_cache() with non-cuda tensors.  

Please let me know if you'd like me to revise anything!

### DIFF
--- a/mistralrs-core/src/models/gemma.rs
+++ b/mistralrs-core/src/models/gemma.rs
@@ -283,7 +283,8 @@ impl Attention {
             // Optimization for seqlen = 1, avoid transpose and just modify reshape dims
             v.reshape((b_sz, self.num_kv_heads, q_len, self.head_dim))?
         };
-
+        
+        let start_offsets_kernel = start_offsets_kernel.to_device(q.device())?;
         self.rotary_emb
             .forward(seqlen_offsets, &start_offsets_kernel, &mut q, &mut k, b_sz)?;
 
--- a/mistralrs-core/src/models/gemma.rs
+++ b/mistralrs-core/src/models/gemma.rs
@@ -284,7 +284,6 @@ impl Attention {
             v.reshape((b_sz, self.num_kv_heads, q_len, self.head_dim))?
         };
 
-        let start_offsets_kernel = start_offsets_kernel.to_device(q.device())?;
         self.rotary_emb
             .forward(seqlen_offsets, &start_offsets_kernel, &mut q, &mut k, b_sz)?;
 
--- a/mistralrs-core/src/models/gemma.rs
+++ b/mistralrs-core/src/models/gemma.rs
@@ -283,7 +283,7 @@ impl Attention {
             // Optimization for seqlen = 1, avoid transpose and just modify reshape dims
             v.reshape((b_sz, self.num_kv_heads, q_len, self.head_dim))?
         };
-        
+
         let start_offsets_kernel = start_offsets_kernel.to_device(q.device())?;
         self.rotary_emb
             .forward(seqlen_offsets, &start_offsets_kernel, &mut q, &mut k, b_sz)?;
--- a/mistralrs-core/src/models/gemma.rs
+++ b/mistralrs-core/src/models/gemma.rs
@@ -320,12 +320,11 @@ impl Attention {
                     None,
                 )?,
                 None => {
-                    let mut input_metadata = PagedAttentionInputMetadata {
-                        block_tables: None,
-                        context_lens: None,
-                        max_context_len: None,
-                        slot_mappings: Tensor::new(&[0f32], q.device())?,
-                    };
+                    // If we don't have metadata, we are most likely generating an imatrix so we don't want to populate that.
+                    // Generating the dummy metadata with the assumption that we are not generating text (only processing prompts).
+                    let mut input_metadata = PagedAttentionInputMetadata::dummy(q.device())?;
+                    // Sanity check.
+                    assert!(attention_mask.is_some());
                     paged_attn.forward(
                         &q,
                         &k,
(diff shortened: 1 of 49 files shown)


mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -353,10 +353,11 @@ impl Loader for GGUFLoader {
                 self.get_id(),
                 device.device_pretty_repr()
             );
-        } else if paged_attn_config.is_some() {
-            warn!("Device mapping or device topology and PagedAttention are incompatible, disabling PagedAttention.");
-            paged_attn_config = None;
         }
+        // } else if paged_attn_config.is_some() {
```
EricLBuehler, line 357: Is there a reason why this was commented out instead of removed? It seems that the change adds full support.

mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -463,7 +482,13 @@ impl Loader for GGUFLoader {
                 model_config,
                 device,
             )?;
-            let cache_engine = CacheEngine::new(model_config, &cache_config, DType::F32, device)?;
+            let cache_engine = CacheEngine::new(
+                model_config,
+                &cache_config,
+                DType::F32,
```
EricLBuehler, line 488: This won't work with #1009, can you please add a comment (`TODO` perhaps) to remind me/others to update this?

mistralrs-core/src/paged_attention/layers/paged_attention.rs
```diff
@@ -70,6 +70,34 @@ impl PagedAttention {
             input_metadata.slot_mappings.clone()
         };
 
+        // When device mapping, these Tensors are fixed on the first device, and must be moved to the same device as q,k,v
```
EricLBuehler, line 73: I see you mentioned a performance penalty:
> Currently, device-mapped paged attention is approximately 10% slower compared to single device paged attention. I found the slowdown is at least partially due to the overhead of moving tensors to the device on every layer forward pass.

To avoid this, can you please update `PagedAttentionInputMetadata` to store all tensors as hashmaps of device location to the actual tensor? Do you think this is a good solution?

This takes up more memory on each GPU but requires only one copy (in the inputs processor) and enables us to remove this section. I'm thinking something similar to [this](https://github.com/EricLBuehler/mistral.rs/blob/d28ddf96d9dc79469cd6f1f856ea4d2370819a80/mistralrs-core/src/models/llama.rs#L480-L494) where we create multiple RoPE instantiations on different devices.

Additionally, I just merged #1014. Can you please merge with master to get these new changes, otherwise a conflict will occur with the addition of the changes I requested above.

mistralrs-core/src/models/gemma.rs
```diff
@@ -284,6 +284,7 @@ impl Attention {
             v.reshape((b_sz, self.num_kv_heads, q_len, self.head_dim))?
         };
 
+        let start_offsets_kernel = start_offsets_kernel.to_device(q.device())?;
```
EricLBuehler, line 287: Can we merge these into `RotaryEmbedding` in `layers.rs`?

EricLBuehler (changes requested):
Hi @cdoko! Thanks for the PR.

> Currently, device-mapped paged attention is approximately 10% slower compared to single device paged attention. I found the slowdown is at least partially due to the overhead of moving tensors to the device on every layer forward pass.

With the comments I made, hopefully this should be adressed. 

> However, I have not yet found a way to pass the layer_devices information there.

Please feel free to make whatever changes you find necessary to get this to work!

I think a nice way to do this would be to add an API to the mapper (`device_map.rs`) to extract all the devices which will be mapped to (inlcuding `normal_loading_metadata.real_device`). One of my comments suggests a way to utilize this information, and I think this method would work nicely with that. What do you think?

mistralrs-core/src/pipeline/gguf.rs
EricLBuehler: Thank you!

EricLBuehler:
Thanks for the updates. I think this is close to merge.

> I updated PagedAttentionInputMetadata to store tensors as hashmaps of device location to the actual tensor. The performance penalty has decreased, but there's still a remaining small penalty of a few percent; I even tested a similar optimization to the attention mask in the model forward pass, since it's also copied every layer, but it didn't give any further performance improvement. I suspect this is due to the unavoidable mapping of xs in the layer forwards. Tensor parallelism might help mitigate the remaining penalty and potentially provide a performance gain.

Sounds great! I agree, TP is something we should look into. I think the hard part is integrating it nicely with the existing codebase.

> To implement the updated PagedAttentionInputMetadata, I made some changes to the pipeline trait. Since the struct now uses hashmaps to store tensors, we need to have the mapper available wherever the struct is instantiated. I added the mapper to the MetadataMixin trait, allowing it to be passed from the pipeline to the input processor, and then used to create the PagedAttentionInputMetadata instance. I also noted that most pipelines don't use paged attention, so I used an Option to represent the mapper and passed None for those pipelines.

Sounds good.

> As a side note, I noticed that Qwen2 and Quantized Llama's RotaryEmbedding uses candle directly - was this intentional or should it be using mistralrs instead, which now includes the Tensor::to_device addition? Other models use the mistralrs one

Yes, can you please update it to use the ones in mistralrs?

EricLBuehler (approved):
@cdoko thanks for the PR!

> Is resolving these cache issues the main blocker for getting speculative decoding working?

Yes, it's just that using the Normal cache isn't supported yet. 

> And what about with PA?

I think the main problem is that we need some extensive management of the KV cache (in particular, rolling back the cache) for PA, which I haven't implemented yet.


> If the PR is ok, I'll probably be working on the VRAM calculations for mistralrs-server next, because currently the flags like --pa-gpu-mem assume single device and don't account for multi-device setups.

Sounds great!

Issue #1026, opened 2025-01-04
Text models output gibberish after multiple requests


cdoko:
Commit: `a06f937e916e583811e5bbcce4573ef6a252c7f4`
Build: cuda or none
Paged Attention disabled. 

The issue occurs with any combination of prefix cache and KV cache enabled/disabled, where the model either outputs something based on a combination with previous prompts or is simply gibberish. 

In `engine/mod.rs`, changing the `pre_op` from `Reset` to `Nothing` in some recent commit caused the issue, as the cache was not being cleared before processing new requests. Reverting the `pre_op` to `Reset` properly clears the cache, resolving the problem when prefix caching is disabled. 
https://github.com/EricLBuehler/mistral.rs/blob/a06f937e916e583811e5bbcce4573ef6a252c7f4/mistralrs-core/src/engine/mod.rs#L209-L246

However, with prefix caching enabled, repeating requests with the same prompt still produces unexpected behavior: responses become unrelated or gibberish from the second request onwards, and subsequent responses are identical. This issue occurs as long as the prompt is identical, even when the same request is repeated in separate, non-consecutive requests. The prefix cacher likely matches the prompt but doesn't handle the cache correctly.

grahamking:
I have the same problem, but for me it doesn't seem to be the CacheInstruction.

It was broken by https://github.com/EricLBuehler/mistral.rs/pull/1000
The last good commit is `3a26a46a0`.

I tested by running the same prompt twice with a variant of [the lower level gguf locally example](https://github.com/EricLBuehler/mistral.rs/tree/master/mistralrs/examples/lower_level/gguf_locally). After #1000 is merged the second response is gibberish.

- Without paged attention ([here](https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs/examples/lower_level/gguf_locally/main.rs#L49)) only the very first query after startup is correct, everything else is gibberish. 
- With paged attention it's only repeated prompts that get the nonsense output.

EricLBuehler:
@grahamking @cdoko I merged #1035 which is disables the prefix cacher (until we get it to work!), but this should result in correct outputs.

Does it work for you?

grahamking:
Yes it does! Thanks for the quick fix.

haricot:
I tested on a06f937e 

works for me with:
`cargo run -F "cuda cudnn" -r -- --no-paged-attn -i plain -m meta-llama/Llama-3.2-1B-Instruct`
or
[...]

PR #1027, opened 2025-01-05
Revamp speculative decoding!

EricLBuehler:
Planning to entail:
- PagedAttention support
- Normal & Full cache support
- Cleaner implementation

@cdoko I'm planning on a full rewrite of speculative decoding. If you have the time to help with this, I would be happy to add you as a collaborator!


EricLBuehler:
```
cargo run --release --features ... '--' -i toml -f speculative.toml
```

EricLBuehler:
> Yes, I'd be happy to help!

Thanks! I sent you a collaborator invite.

EricLBuehler:
@cdoko I sent you a collaborator invite, let me know if you have any trouble accessing the repo.

> From my reading, I thought the rejection sampling step would compare the probabilities of the target model and draft model for the token produced by the draft model. However, the code appears to sample all logits from the target model to generate 'gamma' number of tokens, and then compares these sampled tokens to the draft model's output. The acceptance criteria seems to be based on a match between the sampled tokens and the draft model's tokens.

I agree, the code certainly seems incorrect. I think it warrants a rewrite, what are your thoughts?

> Additionally, I'd like to propose adding device mapping support for speculative decoding. However, this would require implementing device mapping support for the regular cache first.

Yes, we should certainly implement that after it is working.

> I've also encountered a potential issue with the prefix cacher, as reported in https://github.com/EricLBuehler/mistral.rs/issues/1026. While testing speculative decoding, I've disabled the prefix cacher and modified the preop as mentioned in the issue.

Yes, PR #1035 disables the prefix cacher and reverts #1019 so it should *work* but no prefix caching until we find a permanent fix.

PR #1054, opened 2025-01-12
Ergonomic automatic device mapping support

EricLBuehler:
Introduces:
- maximum sequence length
- maximum batch size
- (vision models) maximum image size
- (vision models) maximum number of images

### DIFF
--- a/docs/DEVICE_MAPPING.md
+++ b/docs/DEVICE_MAPPING.md
@@ -7,9 +7,14 @@ Automatic device mapping works by prioritizing loading models into GPU memory, a
 Models architectures such as vision models which greatly benefit from GPU acceleration also automatically prioritize keeping those
 components on the GPU.
 
-To control the mapping across devices, you can set the maximum sequence length and maximum batch size which the model should expect. For vision models,
-you can also specify the maximum image size and number of images. These parameters do not translate to hard limits during runtime, they only control
-the mapping.
+To control the mapping across devices, you can set the following maximum parameters which the model should expect in a prompt.
+
+- maximum sequence length
+- maximum batch size
+- (vision models) maximum image size
+- (vision models) maximum number of images
+
+These parameters do not translate to hard limits during runtime, they only control the mapping.
 
 > [!NOTE]
 > The maximum sequence length is also used to ensure that a KV cache will fit for with and without PagedAttention.
--- a/docs/DEVICE_MAPPING.md
+++ b/docs/DEVICE_MAPPING.md
@@ -7,14 +7,9 @@ Automatic device mapping works by prioritizing loading models into GPU memory, a
 Models architectures such as vision models which greatly benefit from GPU acceleration also automatically prioritize keeping those
 components on the GPU.
 
-### When the model does not fit on 1 GPU
-If the model does not fit on 1 GPU, it will be loaded onto any other GPUs detected and any residual parts onto the CPU.
-
-Some memory is reserved for runtime (KV cache allocation, activations, etc), which can be controlled with the `mb_resrv_per_gpu` parameter (CLI/Python) or the `MbReservePerGpu` enum (Rust).
-
-### When the model fits on 1 GPU
-If the whole model fits on one GPU then no addition memory is reserved as described above. Memory can still be reserved (triggering possible CPU offloading)
-with the `mb_resrv_per_gpu` parameters (CLI/Python) or the `MbReservePerGpu` enum (Rust). 
+To control the mapping across devices, you can set the maximum sequence length and maximum batch size which the model should expect. For vision models,
+you can also specify the maximum image size and number of images. These parameters do not translate to hard limits during runtime, they only control
+the mapping.
 
 ---
 
--- a/docs/DEVICE_MAPPING.md
+++ b/docs/DEVICE_MAPPING.md
@@ -7,13 +7,16 @@ Automatic device mapping works by prioritizing loading models into GPU memory, a
 Models architectures such as vision models which greatly benefit from GPU acceleration also automatically prioritize keeping those
 components on the GPU.
 
-## When the model does not fit on 1 GPU
+### When the model does not fit on 1 GPU
 If the model does not fit on 1 GPU, it will be loaded onto any other GPUs detected and any residual parts onto the CPU.
 
-Some memory is reserved for runtime (KV cache allocation, activations, etc), which can be controlled with the `mb_resrv_per_gpu` parameters (CLI/Python) or the `MbReservePerGpu` enum (Rust).
+Some memory is reserved for runtime (KV cache allocation, activations, etc), which can be controlled with the `mb_resrv_per_gpu` parameter (CLI/Python) or the `MbReservePerGpu` enum (Rust).
 
-## When the model fits on 1 GPU
-If the whole model fits on one GPU then no addition memory is reserved as described above.
+### When the model fits on 1 GPU
+If the whole model fits on one GPU then no addition memory is reserved as described above. Memory can still be reserved (triggering possible CPU offloading)
+with the `mb_resrv_per_gpu` parameters (CLI/Python) or the `MbReservePerGpu` enum (Rust). 
+
+---
 
 If you want to manually device map the model (not recommended), please continue reading.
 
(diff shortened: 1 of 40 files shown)


EricLBuehler:
@cdoko this PR improves our automatic device mapping system for those with limited GPU memory.

The idea of this PR is to enable, for example, running a model with large sequence length that does not fit on one GPU. In that case, you need to make space by reserving memory on the GPU for activations and KV cache.

This PR introduces the `mb_resrv_per_device` parameters which do this. I realized that it is impractical to consider only a max seq len as the sole user input, and for vision models it gets even more complicated because we need to handle varying images with differing sizes. My conclusion was that it is best for the user to have full control.

Do you think this is a better solution or do you have a better idea for solving this problem?

vkomenda:
Reserving the same amount of memory on each of the available GPUs might be wasteful? Unless caches and weights are split evenly among them that is.

vkomenda:
> Yes, I think it's best to split the cache and weights evenly across the available GPUs, in proportion to their available memory.

What if the GPUs are different and don't have the same amount of RAM? I think, if it really needs to be a user-specified absolute parameter, maybe it's better to have it in a topology.toml file that would have various details for each of the available GPUs.

mistralrs-core/src/pipeline/loaders/mod.rs
```diff
@@ -357,7 +357,28 @@ impl ModelKind {
...
 }
 
+macro_rules! b_to_mb {
+    ($x:expr) => {
+        $x / (1024 * 1024)
+    };
+}
+
+#[derive(Debug, Clone)]
+pub enum AutoDeviceMapParams {
```
EricLBuehler, line 367: Allow user to have fine control over automatic device mapping. We should choose sensible defaults for the server/Python APIs. I think these should go in the `plain`/`vision-plain` subcommands and in the respective `Which` variants as they are modality specific.

mistralrs-core/src/pipeline/loaders/vision_loaders.rs
```diff
@@ -1089,6 +1106,36 @@ impl IsqModelLoader for VLlamaLoader {
 }
 
 impl DeviceMappedModelLoader for VLlamaLoader {
+    fn max_act_size_elems(&self, config: &str, params: &AutoDeviceMapParams) -> Result<usize> {
```
EricLBuehler, line 1109: This is the implementation for Llama vision. You can see how we compute the max vision attention size in elements and compare that to the text counterpart. If we have the image seq len for the models (i.e. just what's being fed into the attention) as a function of the config or other information, it is easy from there.

I found this equation for the image seq len in the code for the model, not in the image processor. I haven't checked the other models so I'm not sure if it will apply there too, but maybe we can do something similar.

EricLBuehler (commented):
@cdoko @vkomenda 

Thanks for linking the exllamav2 implementation @cdoko! I took a look and agree it is a super clean implementation! I like how they run a forward pass on each module to ensure it fits, but unfortunately, we don't have the introspection abilities to do that.

The mb_resrv_per_gpu solution was imperfect and the plain manual device mapping would likely work better, so I've removed it. As an alternative solution, I introduced the `AutoDeviceMapParams` enum which allows specifying the max seq len, max batch size, (and for vision models) max image size and number of images. Each model loader will then return the maximum runtime activation usage, and we can allocate the KV cache accordingly too (I have not implemented that).

I implemented the max activation size APIs for all the text models and Llama vision, but not for the rest of the vision models. As a test, Llama vision seems to work nicely when offloading using this new technique. 

@cdoko, I was wondering if you could please take a look at the other vision models and perhaps try to get the maximum image seqlen in the respective vision towers as a function of image size and number of images? Once we have that information, we can easily compute the maximum activation size in the attention block (the highest memory usage area), and quickly move forward.

I added some comments for clarity on the current changes, what do you think about this approach?

EricLBuehler:
> Do these introspection abilities come from PyTorch?

Yes, but also from the fact that it's written in Python.

> Should we also consider:
>- factoring in prompt_batchsize, as it significantly impacts memory usage?
>- accounting for KV quantization when it's implemented in the future?

Yes, but I think that we should target merging this PR first and then factoring in prompt batchsize later. KV cache quantization wouldn't have an impact here because we are looking at peat consumption during the attention operation, that would be factored in later when allocating the KV cache.

> For text model defaults, I suggest:

Those defaults seem reasonable. I'll go ahead and add them to the APIs.

> I had a look at one implementation and found it challenging, as I'm unfamiliar with vision architecture. If you're looking to move quickly, I may not be of much help. However, I'll look at LLaVA first, it looks relatively simpler.

I implemented this for all the models but the LLaVA ones - I'll have some time later today to take a look, so no rush.

EricLBuehler:
@cdoko I just added support for the [LLaVA models](https://github.com/EricLBuehler/mistral.rs/pull/1054/commits/8199a6cc53f66a170b371eded939569c99b595ed).

EricLBuehler:
@cdoko I updated all the APIs so everything should be able to compile and work. If you could give a review or test it out that would be great.

docs/DEVICE_MAPPING.md
```diff
@@ -7,6 +7,17 @@ Automatic device mapping works by prioritizing loading models into GPU memory, a
 Models architectures such as vision models which greatly benefit from GPU acceleration also automatically prioritize keeping those
 components on the GPU.
 
+### When the model does not fit on 1 GPU
+If the model does not fit on 1 GPU, it will be loaded onto any other GPUs detected and any residual parts onto the CPU.
+
+Some memory is reserved for runtime (KV cache allocation, activations, etc), which can be controlled with the `mb_resrv_per_gpu` parameter (CLI/Python) or the `MbReservePerGpu` enum (Rust).
```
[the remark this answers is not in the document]
EricLBuehler: Fixed

EricLBuehler:
@cdoko thanks for testing!

1)
 > I believe this is because both get_device_layers() and the paged attention cache use 0.9 capacity as the cap.

I merged #1059 which should fix this, we now consider the memory availability on all devices.

2)
> Loading large models can cause an OOM error on the first device. ... I thought that factoring in the maximum activation size would account for all memory usage, am I missing something? 

I'll test this with `meta-llama/Llama-3.3-70B-Instruct`. It's interesting that the model fails at all during loading, as we do account for the activation size (this should give some buffer during loading anyway).

3)
> When loading a model across devices, --no-paged-attn, the dummy run fails with a CUDA_ERROR_ILLEGAL_ADDRESS error.

This error seems to come from Flash Attention. Let's solve this in a later PR!

EricLBuehler:
@cdoko I've made some changes which should help address the first two issues you mentioned:

1) [609c1e8](https://github.com/EricLBuehler/mistral.rs/pull/1054/commits/609c1e84f1293b027f550654abb0cb6e7f71094f) takes into account KV cache size based on max_seq_len so we shouldn't run out of memory in the way you described. Can you please test it again (and if it fails, please let me know the model and device size)?
2) [85311d6](https://github.com/EricLBuehler/mistral.rs/pull/1054/commits/85311d609db39c6e258189f134c34ba816c6900d) disables PagedAttention if there is a mix of GPU+CPU in the device mapping. CPU PagedAttention support may be a nice low-hanging fruit in the future for a "good first issue".

EricLBuehler:
I tested this PR on a multi-GPU setup with/without PagedAttention and on a system requiring offloading. All tests passed.

EricLBuehler:
@cdoko thanks for testing. Can you please try Llama 3.3 70B with ISQ to see if it is a problem with the GGUF implementation?

- When we are not using `gpu1`, this is probably a problem with part 3 of the algorithm.
- With the OOM on the first device at layer 38, this is strange as you mentioned layers 0-37 are on gpu0 but 38-78 are on gpu1.
- `Error: Num GPU blocks is 0.` Can you please run with debug and `RUST_BACKTRACE=1` to see what line the error originates from

EricLBuehler:
@cdoko I merged #1069 which fixes several issues relating to `QRmsNorm` and `Embedding` being dequantized, but also improves the UX by adding some warnings. Can you please try to load the model again with and without PagedAttention?

EricLBuehler:
@cdoko just curious, what is the GPU VRAM capacity of your setup (2x3090?)?

PR #1066, opened 2025-01-17
Integrate NCCL

EricLBuehler:


guoqingbao:
Hi Eric,
Great work on supporting the DeepSeek R1 model! I’m curious about how mistral.rs is currently handling the loading of the R1 model, given that its size far exceeds the memory capacity of a single GPU. I noticed you’ve introduced a solution called AutoDeviceMapper—does it provide a TensorParallel-like functionality similar to NCCL?

EricLBuehler:
Hi @guoqingbao!

> I noticed you’ve introduced a solution called AutoDeviceMapper—does it provide a TensorParallel-like functionality similar to NCCL?

No, the automatic device mapping system only moves parts of the model to different devices.

The problem with integrating a TensorParallel functionality is that there would be many far-reaching changes that make this difficult to implement, including:
- abstracting the varbuilder (sharded vs unsharded)
- abstracting the retrieval of sharded tensors (we need to make a `Shard` struct)

PR #1071, opened 2025-01-18
Fix paged attention for vision models on multiple devices

cdoko:
The `DeviceMapper` was set to `None` for vision models before, now it's being passed properly. Tested working on Phi3V.


EricLBuehler (approved):
Thank you!

PR #1082, opened 2025-01-22
Allow using library in CurrentThread runtime

sgrebnov:
The `block_in_place` function requires a multi-threaded runtime and will crash when executed in `CurrentThread` mode, which is the default for unit tests. While unit tests can be configured to run in multi-threaded mode, this is sometimes not practical or may require a significantly more complex setup, especially when tests involve tracing collectors.

With this change, the dummy run is skipped in single-threaded environments, simplifying the setup for unit tests.


EricLBuehler (approved):
Thank you, looks great.

Issue #935, opened 2024-11-26
Couldnt run any vision model


GraphicalDot:
## Describe the bug
Hey everyone,

I’m trying to run vision models in Rust on my M4 Pro (48GB RAM). After some research, I found [Mistral.rs](https://github.com/EricLBuehler/mistral.rs/tree/master), which seems like the best library out there for running vision models locally. However, I’ve been running into some serious roadblocks, and I’m hoping someone here can help!
What I Tried

    Running Vision Models Locally: I tried running the following commands:

`cargo run --features metal --release -- -i --isq Q4K vision-plain -m lamm-mit/Cephalo-Llama-3.2-11B-Vision-Instruct-128k -a vllama`

`cargo run --features metal --release -- -i vision-plain -m Qwen/Qwen2-VL-2B-Instruct -a qwen2vl`

Neither of these worked. When I tried to process an image using Qwen2-VL-2B-Instruct, I got the following error:
```

> \image /Users/sauravverma/Desktop/theMeme.png describe the3 image

thread '<unnamed>' panicked at mistralrs-core/src/vision_models/qwen2vl/inputs_processor.rs:265:30:

Preprocessing failed: Msg("Num channels must match number of mean and std.")

```
This means the preprocessing step failed. Not sure how to fix this.

2. Quantization Runtime Issues: The commands above download the entire model and perform runtime quantization. This consumes a huge amount of resources and isn't feasible for my setup.

3. Hosting as a Server: I tried running the model as an HTTP server using mistralrs-server:

`./mistralrs-server gguf -m /Users/sauravverma/.pyano/models/ -f Llama-3.2-11B-Vision-Instruct.Q4_K_M.gguf`

This gave me the following error:

```
thread 'main' panicked at mistralrs-core/src/gguf/content.rs:94:22:

called \Result::unwrap()` on an `Err` value: Unknown GGUF architecture `mllama``
```

However, when I tried running another model:

`./mistralrs-server -p 52554 gguf -m /Users/sauravverma/.pyano/models/ -f MiniCPM-V-2_6-Q6_K_L.gguf`

What I Need Help With

    Fixing the Preprocessing Issue:

        How do I resolve the Num channels must match number of mean and std. error for Qwen2-VL-2B-Instruct?

    Avoiding Runtime Quantization:

        Is there a way to pre-quantize the models or avoid the heavy resource consumption during runtime quantization?

    Using the HTTP Server for Inference:

[...]

PR #1085, opened 2025-01-23
Move supports_attn_softmax logic to build.rs

EricLBuehler:
@sgrebnov this is the solution I mentioned in #935. What do you think?

### DIFF
--- a/mistralrs-core/build.rs
+++ b/mistralrs-core/build.rs
@@ -1,6 +1,8 @@
 #[cfg(feature = "cuda")]
 const CUDA_NVCC_FLAGS: Option<&'static str> = option_env!("CUDA_NVCC_FLAGS");
 
+const SUPPORTS_ATTN_SOFTMAX_FILE: &str = "src/utils/supports_attn_softmax.rs";
+
 fn main() {
     #[cfg(feature = "cuda")]
     {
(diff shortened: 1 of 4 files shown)


sgrebnov:
@EricLBuehler - this looks great! I'm just trying to learn whether `supports_attn_softmax` is required to compile vs in runtime. My previous impression was that it is required in runtime as I was able to build the app with `metal` acceleration enabled using GH macos runner that does not support acceleration and run successfully on another device with metal acceleration supported.

EricLBuehler:
Hi @sgrebnov!

> I'm just trying to learn whether supports_attn_softmax is required to compile vs in runtime.

The reason why we need the check is that the attn_softmax implementation for Metal uses intrinsic types only present in Metal 3.1+. So really, it only depends on the installed Metal framework version, which is why it *should* be fine to check this at build time (unless the Metal framework version changes during runtime?).

I don't love how this method modifies files in the project - do you perhaps have a better way to pass the information to `attention.rs` is?

Jeadie:
At least in our use case, the metal framework version changes at runtime. Why is this? We want to deliver compiled versions of our metal binary to arbitrary metal users. Therefore our compile time version is our CI computers, but at runtime, it will be whichever version the user has locally.

sgrebnov:
@EricLBuehler - I'm learning about `bfloat` and it seems the app will fail if compiled with `supports_attn_softmax ` but run on device with metal version < 310. As described by @Jeadie the scenario we would like to add support for is the following:  the binaries are compiled once as part of CI/CD and then redistributed and run on different devices (different hardware/software version include Metal framework version). 

As bf16 in Metal supported on all macOS 14+ devices, can we rely on OS version to detect where `supports_attn_softmax` should be enabled? Similar to metal-rs logic https://github.com/gfx-rs/metal-rs/blob/master/src/device.rs#L437?

PR #1088, opened 2025-01-23
Allow chat streaming to use tools

Jeadie:
## Description
 - Add `tool_calls` to `.chunk.delta` within `ChatCompletionChunkResponse` (i.e. `/v1/chat/completion` streaming). 
 - This is done by checking whether the stream _could be_ parsed to the tool call structure (_could be_ meaning that we'd get an EOF, not a format error, when parsing it to a tool structure).

### DIFF
--- a/mistralrs-core/src/pipeline/sampling.rs
+++ b/mistralrs-core/src/pipeline/sampling.rs
@@ -36,8 +36,6 @@ pub(crate) async fn finish_or_add_toks_to_seq(
     );
     // Handle streaming requests
     if seq.get_mut_group().is_streaming {
-        let token_index = seq.get_toks().len();
-
         let mut tool_use_still_possible = false;
         let mut tool_use_is_done = false;
         if let Some(ref t) = seq.tools {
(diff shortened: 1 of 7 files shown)


mistralrs-core/src/pipeline/sampling.rs
```diff
@@ -6,11 +6,33 @@ use rand_isaac::Isaac64Rng;
...
-    sequence::{Sequence, SequenceRecognizer},
+    sequence::{Sequence, SequenceRecognizer, StopReason},
+    tools::ToolCallingMatcher,
+    ToolCallResponse,
 };
 
 use super::Pipeline;
 
+/// Takes raw UTf8 text and parses any possible tool calls from it.
+fn parse_text_tools(
```
EricLBuehler, line 17: Can you please move this to `tools/mod.rs`?

EricLBuehler (changes requested):
Thanks for the PR! Just a small comment.

mistralrs-core/src/pipeline/sampling.rs
Jeadie: Done

EricLBuehler (approved):
Thank you!

PR #1113, opened 2025-01-29
Tensor parallelism and pipeline parallelism

EricLBuehler:
Resource: https://colossalai.org/docs/concepts/paradigms_of_parallelism/


EricLBuehler:
@guoqingbao @csoko automatic tensor parallelism is implemented now, adding pipeline parallelism here too.

guoqingbao:
> @guoqingbao @cdoko automatic tensor parallelism is implemented now, adding pipeline parallelism here too.
> 
> You can run DeepSeek R1 or the 70B distill quite fast with this already :)

Great work!

PR #1133, opened 2025-02-12
MPI support for multi-node

EricLBuehler:
https://docs.open-mpi.org/en/main/index.html

https://github.com/rsmpi/rsmpi


EricLBuehler:
In favor of #1238.

Issue #1162, opened 2025-02-27
build failed on aarch64


jamesvren:
## Minimum reproducible example
The minium example to reproduce the error. Simpler examples make it easier and faster to fix!
just `cargo build --release`

## Error
What is the error? If the error is very long, please try to extract an excerpt of it and post a link to a [Gist](https://gist.github.com/)
'''
error[E0308]: mismatched types
   --> mistralrs-quant/src/distributed/socket.rs:128:23
    |
128 |         Ok(Id::uninit(id_bytes))
    |            ---------- ^^^^^^^^ expected `[u8; 128]`, found `[i8; 128]`
    |            |
    |            arguments to this function are incorrect
    |
    = note: expected array `[u8; 128]`
               found array `[i8; 128]`
note: associated function defined here
   --> mistralrs-quant/src/distributed/mod.rs:196:16
    |
196 |         pub fn uninit(_internal: [::core::ffi::c_char; 128usize]) -> Self {
    |                ^^^^^^ ------------------------------------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `mistralrs-quant` (lib) due to 1 previous error
'''

## Other information
Please specify:
- Operating system: Linux
- GPU or accelerator information
   Just CPU, no GPU or accelerator

## Latest commit or version
Which commit or version you ran with.
Lastest code, it seems introduce by https://github.com/EricLBuehler/mistral.rs/commit/844cdc0ba78bb3d16615f2f9ff8f24311f8c284f

PR #1164, opened 2025-02-28
build fix for arm.

jamesvren:
core::ffi::c_char is defined as u8 instead of i8 in some platforms, so cannot use i8 as type in some codes.
fix https://github.com/EricLBuehler/mistral.rs/issues/1162


EricLBuehler (changes requested):
@jamesvren this looks great, but can you please check the formatting issues?

EricLBuehler (approved):
Thank you!

Issue #1166, opened 2025-02-28
Errors when providing an input image to MiniCPM-O 2.6


jgonera:
## Describe the bug
I'm trying to use mistral.rs to run MiniCPM-O on Google Cloud Run (with NVIDIA L4 GPU). I created a custom Dockerfile (see at the end) based on `Dockerfile.cuda-all` and built the latest master branch. The main addition in my Dockerfile is downloading the model and storing it in the Docker image.

Everything works fine when I send a request with text only, e.g.
```python 
    response = requests.post(
        "http://localhost:9000/v1/chat/completions",
        json={
            "model":"minicpmo_2_6",
            "messages": [
                {
                    "role": "user",
                    "content": "What is your name?",
                }
            ],
            "max_tokens": 256,
            "frequency_penalty": 1.0,
            "top_p": 0.1,
            "temperature": 0,
        }
    )
```

When I try sending the example from [your docs](https://github.com/EricLBuehler/mistral.rs/blob/master/docs/MINICPMO_2_6.md) it fails. The request I'm sending is like this:

```
    response = requests.post(
        "http://localhost:9000/v1/chat/completions",
        json={
            "model":"minicpmo_2_6",
            "messages": [
                {
                    "role": "user",
                    "content": [
                        {
                            "type": "image_url",
                            "image_url": {
                                "url": "https://www.nhmagazine.com/content/uploads/2019/05/mtwashingtonFranconia-2-19-18-108-Edit-Edit.jpg"
                            },
                        },
                        {
                            "type": "text",
                            "text": "(<image>./</image>) What is shown in this image? Write a detailed response analyzing the scene.",
                        },
                    ],
                }
            ],
            "max_tokens": 256,
            "frequency_penalty": 1.0,
            "top_p": 0.1,
            "temperature": 0,
        }
    )
```

I also tried with a different image URL and a base64-encoded image. The result is always the same, an error like this in the logs:

```
[...]

PR #1184, opened 2025-03-10
Complete qwen2_5_vl, and some fixes

brrr:
Complete: qwen2_5_vl model
Add: examples for qwen2_5_vl
Fix: for deterministic sampling, top k SHOULD be Some(1) rather than None
Fix: unquantLinear forward error when device is cuda but not using batch_matmul


MonolithFoundation:
Would like this merged in. Facing Qwen2.5VL issues as well. @EricLBuehler 

btw, can we making some common used mod inside mistralrs_core to be public? So that not everyone using mistralrs need clone the source and modify upon it. It could be easier to maintain a seperated repo host various customized models but powered by mistralrs.

EricLBuehler (changes requested):
Hi @brrr - thanks for the great work! I haven't done any testing yet, but it looks good at an initial glance. Could you please fix the `cargo fmt` and `cargo clippy` issues?

brrr:
As @MonolithFoundation  mentioned, it will be helpful to port new models that making some mistral core structure public. During the preocess of qwen 2.5vl porting, I've made some of those public(crate) and wrote a simple model loader and pipeline, but dicarded them before final push. That because:
1. Most time, SHOULD NOT use core inner structure directly, but use lifetime hooks with some kind of context. Seperated repos using the inner structure directly MAY lead to lots of break changes as long as upstream changing, make them hard to maintance.
2. Besides lifetime hooks and context, we SHOULD also provide a scaffold which MAY include Tracing and Performance metric like OpenTelemetry, Evals to test each  hook function input/output (tensor shape, precision, even somehow to compare or use a persisted tensor such as exported npy file from Transformers)
3. "Convertion Kits": SHOULD provide OEM Kits including prefix cache, KV cache, attention implement... but also provide a simple way to let downstream change some parts easily. For example a agentic framework with built-in LLM inferrence engine MAY want to use their own prefix and kv cache impletments. 
I think it will be the key factor to make Mistralrs massive adoption, plus  "Blazingly fast ".

mistralrs-core/src/vision_models/qwen2_5_vl/inputs_processor.rs
```diff
@@ -0,0 +1,749 @@
...
+            .nth(0)
+            .unwrap()
+            .unwrap()
+        };
+
+        // todo: set_use_matmul_via_f16(true) from "pipline/inputs_processor" cause a significant loss of precision.
+        // It’s hard to figure it out during subsequent debugging
+        // Anyhow, globally setting matnuml precision MAY not be a ideal solution.
+        // For now, change the precision back
+        set_use_matmul_via_f16(false);
```
EricLBuehler, line 472: I'm planning on deprecating and then removing this shortly. It was only for a specific GPU arch and provided a relatively small boost.

mistralrs-core/src/vision_models/qwen2_5_vl/inputs_processor.rs
```diff
@@ -0,0 +1,749 @@
...
+
+                let vid_pad = tokenizer
+                    .encode(Qwen2_5VLProcessor::VIDEO_PAD, false)
+                    .expect("Detokenization failed!")
+                    .get_ids()
+                    .to_vec();
+                let continuous_vid_pad = find_sequences(&ids, vid_pad[0]);
+                all_continuous_vid_pad.push(continuous_vid_pad);
+
+                seq.set_toks(ids);
```
EricLBuehler, line 352: If we are using `set_toks`, we need to free & reallocate the sequence for the pagedattn scheduler. Otherwise, problems like #1166 occur.

See: https://github.com/EricLBuehler/mistral.rs/blob/b8237b2de79c2a1090c1cd28c4d1cb8e15077c89/mistralrs-core/src/vision_models/qwen2vl/inputs_processor.rs#L398-L402

(we should probably also split this into one that does the free & reallocate automatically, and one like this marked `unsafe`).

mistralrs-core/src/vision_models/qwen2_5_vl/vision.rs
```diff
@@ -0,0 +1,515 @@
...
+            let llm_grid_h = h / self.spatial_merge_size;
+            let llm_grid_w = w / self.spatial_merge_size;
+            let index = Tensor::arange(0i32, (t * llm_grid_h * llm_grid_w) as i32, &Device::Cpu)?
+                .reshape((t, llm_grid_h, llm_grid_w))?;
+            let pad_h = vit_merger_window_size - llm_grid_h % vit_merger_window_size;
+            let pad_w = vit_merger_window_size - llm_grid_w % vit_merger_window_size;
+            let num_windows_h = (llm_grid_h + pad_h) / vit_merger_window_size;
+            let num_windows_w = (llm_grid_w + pad_w) / vit_merger_window_size;
+            let index_padded = {
+                let h = Tensor::full(-100i32, (t, pad_h, llm_grid_w), &Device::Cpu)?;
```
EricLBuehler, line 382: Let's avoid using "magic values" - can you make this a `const` so we can identify it by name?

mistralrs/Cargo.toml
```diff
@@ -35,3 +35,8 @@ flash-attn = ["cuda", "mistralrs-core/flash-attn"]
 accelerate = ["mistralrs-core/accelerate"]
 mkl = ["mistralrs-core/mkl"]
 nccl = ["mistralrs-core/nccl"]
+
+[[example]]
```
EricLBuehler, line 39: I see that without this, the lower-level examples are not directly runnable. For consistency, could you please add corresponding ones for all of the other lower-level examples, or remove this one.

mistralrs/examples/qwen2_5vl/main.rs
```diff
@@ -0,0 +1,65 @@
+use anyhow::Result;
+use clap::Parser;
+use mistralrs::{
+    IsqType, TextMessageRole, TextMessages, VisionLoaderType, VisionMessages, VisionModelBuilder,
```
EricLBuehler, line 4: Can you please make it so we always use `IsqType::Q8_0` (for these models that is probably reasonable), and remove the usage of `TextMessages`?

mistralrs/examples/qwen2_5vl/main.rs
```diff
@@ -0,0 +1,65 @@
...
+        ) {
+            Ok(http_resp) => Ok(http_resp.bytes()?.to_vec()),
+            Err(e) => anyhow::bail!(e),
+        }
+    }).await??;
+
+    let image = image::load_from_memory(&bytes)?;
+
+    //force map all layers to gpu
+    let device_mapper = DeviceMapSetting::Map(DeviceMapMetadata::from_num_device_layers(vec![
```
EricLBuehler, line 35: By default, automatic device mapping will be used to optimally spread layers across multiple-GPUs/CPU, while preserving space for activations. Can you please remove this?

EricLBuehler (changes requested):
Hi @brrr!

I tested this out and it works well! I added a few (relatively minor) comments which should be addressed before merge; otherwise everything looks great.

MonolithFoundation:
@EricLBuehler @brrr  In this PR https://github.com/EricLBuehler/mistral.rs/pull/1185 it's port needed components to public, and we written an thirdparty-model zoo including qwen2.5 vl, it works well. https://github.com/lucasjinreal/Namors

It's far more easy to write customized model with this modification.

mistralrs-core/src/vision_models/qwen2_5_vl/inputs_processor.rs
EricLBuehler: Removed it in #1196 :)

mistralrs-core/src/vision_models/qwen2_5_vl/inputs_processor.rs
brrr: Done.

mistralrs-core/src/vision_models/qwen2_5_vl/vision.rs
brrr: Done.

mistralrs/examples/qwen2_5vl/main.rs
brrr: Done.

mistralrs/examples/qwen2_5vl/main.rs
brrr: > By default, automatic device mapping will be used to optimally spread layers across multiple-GPUs/CPU, while preserving space for activations. Can you please remove this?
 
Done.

And one more thing, when I removed manual device mapping, I want to have a test with the CPU and Gpu(Metal) mixture inference. When without Accelerate feature, matmul will use f16 which MAY cause NaN error. I intended to use  clamp_for_f16 to prevent that happening, but I think it's not elegant. @EricLBuehler  you MAY  reconsider a comprehensive  refactoring of these code? Especially, when Apple releases new chip which supports native fp8, it's will still be a problem.

https://github.com/EricLBuehler/mistral.rs/blob/b8237b2de79c2a1090c1cd28c4d1cb8e15077c89/mistralrs-quant/src/lib.rs#L191-L199

mistralrs/Cargo.toml
brrr: Removed

EricLBuehler:
> As @MonolithFoundation mentioned, it MAY be helpful to port new models that making some mistral core structure public. During the preocess of qwen 2.5vl porting, I've made some of those public(crate) and wrote a simple model loader and pipeline, but dicarded them before final push. That because:
> 
> 1. Most time, SHOULD NOT use core inner structure directly, but use lifetime hooks with some kind of context. Seperated repos using the inner structure directly MAY lead to lots of break changes as long as upstream changing, make them hard to maintance.
> 2. Besides lifetime hooks and context, we SHOULD also provide a scaffold which MAY include Tracing and Performance metric like OpenTelemetry, Evals to test each  hook function input/output (tensor shape, precision, even somehow to compare or use a persisted tensor such as exported npy file from Transformers)
> 3. "Convertion Kits": SHOULD provide OEM Kits including prefix cache, KV cache, attention implement... but also provide a simple way to let downstream change some parts easily. For example a agentic framework with built-in LLM inferrence engine MAY want to use their own prefix and kv cache impletments.
>    I think it will be the key factor to make Mistralrs massive adoption, plus  "Blazingly fast ".

I think I mostly agree with this. #1185 is a good step, and I'll evaluate if it's reasonable to export that functionality - I feel that we already export too much from `mistralrs-core`. Perhaps we could design a system for trait-based hooks into the loading and pipeline process, exporting the utilities that we have access to but nothing internal.

> 3. "Convertion Kits": SHOULD provide OEM Kits including prefix cache, KV cache, attention implement... but also provide a simple way to let downstream change some parts easily

I think this could also be done with the trait-based hook system. Similar to what we already have with [CustomLogitsProcessor](https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/sampler.rs#L169)!

brrr:
> @brrr the changes you made look good. There was an internal-API that had a breaking change in #1190, which is causing these errors. Could you please take a look?

Done.

EricLBuehler (approved):
Thank you!

PR #1185, opened 2025-03-10
Making some interface public so that users can reuse

MonolithFoundation:
This PR making some interface public, so that users can using mistral.rs as a lib, rewritten any thirdparty models definitions.

With current modification, all functionality are remain untouched, but users can written Qwen2.5 VL or any customized LLava model outside of mistral.rs source code, make the code maintrainance more simpler and model support evolving more faster.


EricLBuehler (changes requested):
Hi @MonolithFoundation! Thanks for the PR. I agree with the change, it absolutely makes sense. If you could check the formatting and clippy errors, that would be great, and I will merge this PR.

Additionally, can you please export some of the core crates used, including:
- `candle_core`
- `candle_nn`
- (if the `flash-attn` feature flag is set) `candle_flash_attn`
- (if the `flash-attn-v3` feature flag is set) `candle_flash_attn_v3`
- `mistralrs-quant`
- `mistralrs-vision`

If you make new model implementations in a seperate repo, please create an issue an I will include them as a link (or even perhaps upstreaming).

MonolithFoundation:
Obsolutely!

Will fixing the suggestions now.

MonolithFoundation:
@EricLBuehler Hi, I had fixed clippy and related issues for exposing related (minimal) interface as suggested.

Here is the detailed modification:

- fixed clippy issues for exposing related interface:

Added:

```
 pub fn is_empty(&self) -> bool {
        self.rows.is_empty()
    }
```
to: `AnyMoeTrainingInputs` for fix clappy;

Added:

```
 pub fn is_empty(&self) -> bool {
        self.tokens.is_empty()
    }
```

to: `Sequence` for fix clappy;

- making the exposing level same to pass clippy.

The final checking:

image

And I contributing drafting a customized model repo (very early stage) at: https://github.com/lucasjinreal/Namors which deploys https://github.com/lucasjinreal/Namo-R1 with mistral.rs ( A extreamly tiny MLLM with Qwen2.5 500M).

I hope once this PR merged we can remove dependencies in `Namors` and using git as deps directly.

MonolithFoundation:
The PR ready to merge! The example thirdparty model located at this repo: https://github.com/lucasjinreal/Namors and it's works like a charm!

MonolithFoundation:
@EricLBuehler Hi, eric, is there anything further I can do to help merge this PR?

Issue #865, opened 2024-10-19
Memory leak and channel closure issues when reusing/dropping Model


solaoi:
## Describe the bug

When initializing and dropping the Model repeatedly:

1. Memory usage continuously increases as GGUF models aren't properly cleaned up
2. Channel is erroneously closed after the first iteration

### Steps to Reproduce

1. Create a service that initializes and drops the model multiple times
2. Run the following code:

```main.rs
use anyhow::Result;
use mistralrs::{GgufModelBuilder, PagedAttentionMetaBuilder, TextMessageRole, TextMessages};
use std::time::Duration;
use tokio::time::sleep;

struct ChatService {
    model: Option<mistralrs::Model>,
}

impl ChatService {
    async fn new() -> Result<Self> {
        Ok(Self { model: None })
    }

    async fn initialize_model(&mut self) -> Result<()> {
        self.model = Some(
            GgufModelBuilder::new(
                "gguf_models/mistral_v0.1/",
                vec!["mistral-7b-instruct-v0.1.Q4_K_M.gguf"],
            )
            .with_chat_template("chat_templates/mistral.json")
            .with_paged_attn(|| PagedAttentionMetaBuilder::default().build())?
            .build()
            .await?,
        );
        Ok(())
    }

    async fn chat(&self, prompt: &str) -> Result<String> {
        let messages = TextMessages::new().add_message(TextMessageRole::User, prompt);

        let response = self
            .model
            .as_ref()
            .unwrap()
            .send_chat_request(messages)
            .await?;

        Ok(response.choices[0]
            .message
            .content
            .clone()
            .unwrap_or_default())
    }
}

#[tokio::main]
async fn main() -> Result<()> {
    for i in 0..3 {
        println!("Iteration {}", i);

        let mut service = ChatService::new().await?;
        service.initialize_model().await?;

        let response = service.chat("Write a short greeting").await?;
        println!("Response: {}", response);

        // Model is dropped here, but GGUF remains in memory
        drop(service);

        // Wait to make memory usage observable
        sleep(Duration::from_secs(5)).await;
    }

    Ok(())
}
```

`Cargo.toml` is here:

```
[package]
name = "memory_bug_mistral"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
tokio = { version = "1", features = ["full"] }
anyhow = "1.0"
[...]

PR #1211, opened 2025-03-15
Engine ID fix

andrewlimmer:
When dropping MistralRs, it does not emit the correct ENGINE_ID to Terminate the engine. 

Currently if you try to drop a MistralRs instance, and start a new instance. The old engine is still running, and a new engine is created. 

This PR fixes the error described in this issue:
https://github.com/EricLBuehler/mistral.rs/issues/865#issue-2598762587

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.a
 .DS_Store
 .idea
+/gguf_models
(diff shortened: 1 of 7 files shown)


andrewlimmer:
Honestly, I added a commit to this PR by mistake.

PR #1214, opened 2025-03-16
adapt to rig crate as client

benliao:


EricLBuehler:
@benliao can you please describe the situation when this change would be used/necessary that is not user error?

Just curious!

benliao:
@EricLBuehler When I use [rig.rs](https://github.com/0xPlaygrounds/rig). I use the same [code](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/agent.rs) with official openai API code or sglang delployed service, works fine. but failed whith mistral.rs server. Also encountered error with [cherry-studio](https://github.com/CherryHQ/cherry-studio). After my patch, they work fine. Thanks.

EricLBuehler (approved):
Thank you for the explanation!

PR #1284, opened 2025-04-18
update to llguidance 0.7.16

mmoskal:
- this updates llguidance to 0.7.16 from crates.io - the openssl dep is now gone from `toktrie_hf_tokenizers`
- use `llguidance::Matcher` instead of Constraint - it has additional APIs and is generally cleaner
- use ParserFactory for building Matchers - this enables [slicer optimization](https://github.com/guidance-ai/llguidance/blob/main/docs/optimizations.md#slicer-optimization) and is anyway the only way to do it now
- use Matcher.validate_tokens() first, so we can often skip computing masks
- rework the spec-decoding code a bit

@EricLBuehler I need comments on the last bit (the rest should be uncontroversial) - I changed `sample_target_sequence_speculative()` to take draft tokens and stop sampling once it diverges, instead of sampling everything first and only dropping tokens later - it saves some sampling and simplifies handling of constraints, but if it's a problem I can revert

I tested the various `*.py` scripts that deal with constrained decoding, but didn't really test the spec-decoding yet (I'm pretty sure spec-decoding + constrained-decoding isn't working now (before this PR) judging from the code)

### DIFF
--- a/mistralrs-core/src/pipeline/llg.rs
+++ b/mistralrs-core/src/pipeline/llg.rs
@@ -1,21 +1,18 @@
 use std::sync::Arc;
 
 use anyhow::Result;
-use llguidance::{
-    api::{ParserLimits, TopLevelGrammar},
-    toktrie::{InferenceCapabilities, TokEnv},
-    TokenParser,
-};
+use llguidance::{api::TopLevelGrammar, ParserFactory};
 use tokenizers::Tokenizer;
 
 use crate::Constraint;
 
-pub fn build_tok_env(tokenizer: Tokenizer) -> TokEnv {
-    let bt = toktrie_hf_tokenizers::ByteTokenizer::from_tokenizer(tokenizer)
-        .expect("Failed to create ByteTokenizer from Tokenizer");
-    let env = toktrie_hf_tokenizers::ByteTokenizerEnv::new(bt, None)
+pub fn build_llg_factory(tokenizer: Tokenizer) -> Arc<ParserFactory> {
+    let env = toktrie_hf_tokenizers::ByteTokenizer::from_tokenizer(tokenizer)
+        .expect("Failed to create ByteTokenizer from Tokenizer")
+        .into_tok_env(None)
         .expect("Failed to create ByteTokenizerEnv");
-    Arc::new(env)
+    let factory = ParserFactory::new_simple(&env).expect("Failed to create ParserFactory");
+    Arc::new(factory)
 }
 
 pub fn llg_grammar_from_constraint(constraint: &Constraint) -> Result<Option<TopLevelGrammar>> {
@@ -30,18 +27,9 @@ pub fn llg_grammar_from_constraint(constraint: &Constraint) -> Result<Option<Top
 }
 
 pub fn constraint_from_llg_grammar(
-    tok_env: TokEnv,
+    factory: &ParserFactory,
     grm: TopLevelGrammar,
 ) -> Result<llguidance::Constraint> {
-    let parser = TokenParser::from_grammar(
-        tok_env,
-        grm,
-        llguidance::Logger::new(0, 1),
-        InferenceCapabilities {
-            ..Default::default()
-        },
-        ParserLimits::default(),
-        vec![],
-    )?;
+    let parser = factory.create_parser(grm)?;
     Ok(llguidance::Constraint::new(parser))
 }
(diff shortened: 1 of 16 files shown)


mistralrs-core/src/pipeline/llg.rs
```diff
@@ -1,21 +1,18 @@
...
 
-pub fn build_tok_env(tokenizer: Tokenizer) -> TokEnv {
-    let bt = toktrie_hf_tokenizers::ByteTokenizer::from_tokenizer(tokenizer)
-        .expect("Failed to create ByteTokenizer from Tokenizer");
-    let env = toktrie_hf_tokenizers::ByteTokenizerEnv::new(bt, None)
+pub fn build_llg_factory(tokenizer: Tokenizer) -> Arc<ParserFactory> {
+    let env = toktrie_hf_tokenizers::ByteTokenizer::from_tokenizer(tokenizer)
+        .expect("Failed to create ByteTokenizer from Tokenizer")
+        .into_tok_env(None)
         .expect("Failed to create ByteTokenizerEnv");
```
EricLBuehler, line 13: Nit: Can you update this function to return a some sort of result instead of using `.expect`?

mistralrs-core/src/pipeline/sampling.rs
```diff
@@ -348,7 +346,6 @@ pub async fn sample_sequence(
     return_logprobs: bool,
     rng: Arc<std::sync::Mutex<Isaac64Rng>>,
     use_async_pool: bool,
-    add_to_trie: bool,
```
EricLBuehler, line 351: 👍

mistralrs-core/src/pipeline/sampling.rs
```diff
@@ -454,28 +452,44 @@ pub struct SpeculativeSample {
...
+            seq,
+            return_logprobs,
+            rng.clone(),
+            true, // TODO(EricLBuehler): does this hurt perf?
+            true,
+        )
+        .await?;
+        let sampled_token = sample.token;
+        sampled.push(SpeculativeSample { sample });
+        if sampled_token != draft.sample.token {
```
EricLBuehler, line 490: 👍 

I agree with this, makes sense!

mistralrs-core/src/pipeline/speculative.rs
```diff
@@ -480,21 +478,24 @@ impl Pipeline for SpeculativePipeline {
...
-                }
+                let accepted_tokens = samples.into_iter().map(|s| s.sample).collect::<Vec<_>>();
 
                 // ======================= Narrow caches to account for rejections ============================
                 let n_not_accepted = self.gamma - accepted_tokens.len();
+
+                // rollback any tokens that were not accepted
+                match seq.recognizer {
+                    SequenceRecognizer::Llguidance(ref mut llg) => {
+                        llg.rollback(n_not_accepted)
```
EricLBuehler, line 493: 👍 

Makes sense to have `rollback` here again, as the one in `sample_target_sequence_speculative` is associated with the sampling of draft tokens. Perhaps you can put a comment describing this at that callsite?

EricLBuehler (changes requested):
Hi @mmoskal! Thanks for the PR & updating it! 

I reviewed the last part and the updated speculative decoding logic is sound to me - I made a few small comments.

The slicer optimization seems really interesting! I was actually using the structured outputs earlier today with a JSON schema through Pydantic but ran into an error that the `maxProperties` key was unimplemented. Is this something that would be addressed in this update?

mmoskal:
Thanks! Will look into fixes tomorrow!

the max properties can be ignored after this update with “lenient” flag on schema [docs](https://github.com/guidance-ai/llguidance/blob/main/docs/json_schema.md#whitespace-handling)

This would generate warnings. I wonder how would I expose them?

actually supporting max properties is possible at least in some cases I guess

mmoskal:
Tracking https://github.com/guidance-ai/llguidance/issues/168

mistralrs-core/src/pipeline/speculative.rs
mmoskal: Whoops! I forgot to remove this - the new sample_target_sequence_speculative() only moves forward llg state for the tokens accepted

EricLBuehler (approved):
Thank you! I'll follow along on that issue 👍

PR #1285, opened 2025-04-18
Add the Qwen 3 and Qwen 3 MoE models

EricLBuehler:
🚨🚨 All models are working!


AsbjornOlling:
How did you deal with the qwen 3 chat template?

It seems like it includes a loop with a python slicing syntax, that isn't supported by minijinja: `{%- for message in messages[::-1] %}`

EricLBuehler:
@AsbjornOlling we currently replace it with `|reverse`.

 @mitsuhiko it looks like minijinja doesn't support signed step values? I get an error stating that number cannot be converted to u64.

PR #1295, opened 2025-04-25
Add masked_scatter fused op

EricLBuehler:


EricLBuehler:
Doesn't work. Only reference impl was torch that uses cub...

PR #1309, opened 2025-05-02
Fix CPU hogging in interactive mode

beeender:
The log enabler should be checked after the sleep instead of a busy
loop checking.

Since the interactive mode always disables the token speed logger, 100%
CPU was taken by this loop always.

(There is another busy loop problem which will be fixed in another PR.)


EricLBuehler (approved):

EricLBuehler:
Thank you!

PR #1311, opened 2025-05-03
Add Metal precompilation support 

EricLBuehler:


GaetanLepage:
Hi,

This feature breaks the nix build of `mistral-rs` as metal is not available in the build sandbox.
Do you think pre-compilation could be toggleable by an env variable?

Context: [bumping mistral.rs to 0.6.0 in nixpkgs](https://github.com/NixOS/nixpkgs/pull/415771#pullrequestreview-2917641215)

PR #1371, opened 2025-05-25
Fix CPU block is_gpu init

EricLBuehler:
## Summary
- fix CPU BlockEngine is_gpu flag for paged attention implementations

## Testing
- `cargo test --workspace --quiet` *(fails: Could not clone git dependency)*


EricLBuehler:
In #1369.

PR #1400, opened 2025-06-02
PagedAttention kv-cache quantization (F8E4M3)

EricLBuehler:
## Summary by CodeRabbit

- **New Features**
  - Added support for FP8 (8-bit floating point) quantized key-value caches in PagedAttention for CUDA and Metal backends with optional per-tensor scaling factors.
  - Introduced a configurable cache data type for PagedAttention via a new `cache_type` setting (`auto` and `f8e4m3`).
  - Enabled specifying the PagedAttention cache type through CLI arguments, Python, and Rust builder APIs.
  - Added support for new model architectures: Qwen3 and GLM4.

- **Enhancements**
  - Improved Metal device memory allocation to respect system memory limits.
  - Extended kernels and backends to transparently handle new data types and scaling factors.
  - Added thread-safe scale tracking in PagedAttention for improved attention computations.

- **Documentation**
  - Updated APIs and Python type hints to include new cache type options and model support.
  - Added documentation for KV cache quantization feature with usage examples.

- **Bug Fixes**
  - Prevented underflow in sequence choice count decrement with saturating subtraction.
  - Improved error handling and argument validation for cache type and scaling factors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2529,8 +2529,7 @@ checksum = "241eaef5fd12c88705a01fc1066c48c4b36e0dd4377dcdc7ec3942cea7a69956"
 [[package]]
 name = "llguidance"
 version = "0.7.29"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b36091de5db3301cf2a2a16fec023b69dce306d2fdc9ba6a8a1fe9b404d6ad37"
+source = "git+https://github.com/guidance-ai/llguidance.git?rev=2ce5ab8#2ce5ab8196f16dd8beba5a3d874eb1ab74e0268c"
 dependencies = [
  "anyhow",
  "derivre",
@@ -5399,8 +5398,7 @@ dependencies = [
 [[package]]
 name = "toktrie"
 version = "0.7.29"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24abc046cdf691cae38efcc45a52e68e03ab26954157c6c15e2e7c9f6e46fef4"
+source = "git+https://github.com/guidance-ai/llguidance.git?rev=2ce5ab8#2ce5ab8196f16dd8beba5a3d874eb1ab74e0268c"
 dependencies = [
  "anyhow",
  "bytemuck",
@@ -5412,8 +5410,7 @@ dependencies = [
 [[package]]
 name = "toktrie_hf_tokenizers"
 version = "0.7.29"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "200f5b166ecb572393f0fdacabc7904a6f5f9bc2766e784e4d2b0c7a80e2bed9"
+source = "git+https://github.com/guidance-ai/llguidance.git?rev=2ce5ab8#2ce5ab8196f16dd8beba5a3d874eb1ab74e0268c"
 dependencies = [
  "anyhow",
  "log",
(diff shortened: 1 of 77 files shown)


sempervictus:
@EricLBuehler - any reason for making this metal-specific? Do CUDA devices not support the compressed format?

EricLBuehler:
> @EricLBuehler - any reason for making this metal-specific? Do CUDA devices not support the compressed format?

No, not at all. Just prototyping for Metal initially, CUDA will be equivalent.

sempervictus:
The good news is that it compiles and doesn't blow up on boot. The bad news is that it blows up on model-load:
```rust
mistralrs-svc0  | src/cuda/quantization/fp8/nvidia/quant_utils.cuh:240: Tout vllm::fp8::scaled_convert(const Tin &, float) [with Tout = unsigned char; Tin = unsigned short; vllm::Fp8KVCacheDataType kv_dt = vllm::Fp8KVCacheDataType::kFp8E4M3]: block: [0,0,0], thread: [127,0,0] Assertion `false` failed.
mistralrs-svc0 exited with code 0
```
Looks like that spawned 512 threads all with the same result (just returned a bit out of order):
```rust
mistralrs-svc0  | src/cuda/quantization/fp8/nvidia/quant_utils.cuh:240: Tout vllm::fp8::scaled_convert(const Tin &, float) [with Tout = unsigned char; Tin = unsigned short; vllm::Fp8KVCacheDataType kv_dt = vllm::Fp8KVCacheDataType::kFp8E4M3]: block: [0,0,0], thread: [511,0,0] Assertion `false` failed.
```

Of likely relevance - this is a CC7 system, so no bf16 though it looks like where relevant, that should be detected and bail.

EricLBuehler:
Hi @sempervictus! Thanks for testing this. It should work now.

sempervictus:
@EricLBuehler clean rebuild and one small PR later, seeing:
```
mistralrs-svc3  | src/cuda/quantization/fp8/nvidia/quant_utils.cuh:240: Tout vllm::fp8::scaled_convert(const Tin &, float) [with Tout = unsigned char; Tin = unsigned short; vllm::Fp8KVCacheDataType kv_dt = vllm::Fp8KVCacheDataType::kFp8E4M3]: block: [0,0,0], thread: [248,0,0] Assertion `false` failed.
```
when built and run with #1468 (cuda 12.8 until the candle PR lands)

sempervictus:
:man_facepalming: that one's on me, sorry. Do you want to fmt that yourself so i dont pollute your branch with my one-off commits anymore? :-)

EricLBuehler:
> 🤦‍♂️ that one's on me, sorry. Do you want to fmt that yourself so i dont pollute your branch with my one-off commits anymore? :-)

Haha no problem!

Since CC<8 doesn't support fp8/bf16, we might need to emulate it. What do you think?

Do you have some other contact?

sempervictus:
Didn't realize fp8 was also off the table - interesting, thank you. I wonder how far the approach we saw in that flashattention kernel for older systems can be taken to emulate via padding and stripping at runtime (i'm guessing you have a better idea though).

IMO, it doesn't make much sense to hold fp16 KV caches when we're quantizing models down q4k most of the time, especially given the cost of memory on these older GPUs which "only" have 32G a pop.

sempervictus:
@EricLBuehler would it be feasible to quantize-down the KV cache type to the same format as the model itself and leverage INT8s where FP8/16 is not available?

EricLBuehler:
@sempervictus adding int8/other KV quants next!

PR #1419, opened 2025-06-03
Provide chat_templates to container users

sempervictus:
Models often come without chat templates requiring mapping them from the source repository into a container for access by the mistralrs-server.

Copy the templates from the build tree into the root of the image to permit use via `--chat-template /chat_templates/something.json`

TODO:
  With the increase in quantized models and support for other
formats, the initial benchmark run during model load can be used to qualify/select existing chat templates embedded into the binary for models which do not come with any (to include output of the functional failures in each test allowing users to modify the ones already provided correctly to suit the model being loaded).

## Summary by CodeRabbit

- **New Features**
  - Chat templates are now included in the runtime environment, making them available for users running models that may not provide templates by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,3 +37,6 @@ COPY --from=builder /mistralrs/target/release/mistralrs-web-chat /usr/local/bin/
 
 # Make the binaries executable
 RUN chmod +x /usr/local/bin/mistralrs-bench /usr/local/bin/mistralrs-server /usr/local/bin/mistralrs-web-chat
+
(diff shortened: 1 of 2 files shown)


EricLBuehler (approved):
Thank you!

> With the increase in quantized models and support for other
formats, the initial benchmark run during model load can be used to qualify/select existing chat templates embedded into the binary for models which do not come with any (to include output of the functional failures in each test allowing users to modify the ones already provided correctly to suit the model being loaded).

Sounds interesting, can you please elaborate a bit?

PR #1420, opened 2025-06-03
Web search improvements (bm25, web chat)

EricLBuehler:
## Summary by CodeRabbit

- **New Features**
  - Added optional web search functionality with configurable context size in the chat interface.
  - Integrated fallback BM25 ranking for search results when no embedding model is available.
  - Enabled embedding model support for text and vision models to enhance search relevance.
- **Refactor**
  - Improved handling of blocking operations within search and extraction processes to enhance application responsiveness during asynchronous tasks.
  - Streamlined WebSocket message handling with grouped parameters and optional web search options.
- **UI Enhancements**
  - Introduced a Web Search control panel with enable toggle and context size selector.
  - Updated message sending to include web search options when enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,6 +357,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "bm25"
+version = "2.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9874599901ae2aaa19b1485145be2fa4e9af42d1b127672a03a7099ab6350bac"
+dependencies = [
+ "cached",
+ "deunicode",
+ "fxhash",
+ "rust-stemmers",
+ "stop-words",
+ "unicode-segmentation",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


sempervictus:
There are some purpose-built mechanisms for this function - we use searxng behind a rotating TOR proxy pretty heavily and along with things like scrape4ai. 

Architecturally-speaking (am a security architect), IMO the logic components (models) should probably run as isolated and far away from the world as they can since they eventually get re-trained with privileged information which can be siphoned out so as to access access external resources/content/function via application layer gateways such as MCP servers. Otherwise we have the unsanitized input problem wherein attackers will eventually poison content crafted to be retrieved by the search logic of these things with injections to be directly retrieved into their context window to corrupt it.

EricLBuehler:
> There are some purpose-built mechanisms for this function - we use searxng behind a rotating TOR proxy pretty heavily and along with things like scrape4ai.
> 
> Architecturally-speaking (am a security architect), IMO the logic components (models) should probably run as isolated and far away from the world as they can since they eventually get re-trained with privileged information which can be siphoned out so as to access access external resources/content/function via application layer gateways such as MCP servers. Otherwise we have the unsanitized input problem wherein attackers will eventually poison content crafted to be retrieved by the search logic of these things with injections to be directly retrieved into their context window to corrupt it.

Interesting point @sempervictus! Will add some logic that allows users of mistral.rs to bypass the builtin search logic and write their own.

PR #1453, opened 2025-06-09
Add MCP server endpoints

EricLBuehler:
## Summary
- add `rust-mcp-sdk` dependency
- implement simple MCP server handler and startup
- expose new `--mcp-port` option in server CLI

## Testing
- `cargo fmt --all`
- `cargo clippy --workspace --tests --examples -- -D warnings`
- `cargo test --workspace --quiet` *(fails: command cancelled)*

------
https://chatgpt.com/codex/tasks/task_e_6847025b7ed4832291abb07ddd9d65cb

## Summary by CodeRabbit

- **New Features**
  - Added support for serving the Model Communication Protocol (MCP) via a new optional server mode, configurable with a command-line port argument.
  - Introduced a new MCP server handler that processes chat requests and returns responses using the MCP protocol.
  - Enabled parallel serving of models over both HTTP API and MCP protocol.

- **Chores**
  - Added new dependencies and features to support MCP server functionality.
  - Updated documentation to include MCP protocol usage, configuration, and examples.

- **Bug Fixes**
  - Improved type handling for server port arguments to enhance reliability.

### DIFF
--- a/mistralrs-server/src/mcp_server.rs
+++ b/mistralrs-server/src/mcp_server.rs
@@ -268,6 +268,12 @@ impl HttpMcpHandler {
                 result: Some(serde_json::to_value(&self.server_info).unwrap()),
                 error: None,
             },
+            "ping" => JsonRpcResponse {
+                jsonrpc: "2.0".to_string(),
+                id: request.id,
+                result: Some(json!({})),
+                error: None,
+            },
             "tools/list" => {
                 let tools: Vec<Tool> = self.tools.values().map(|t| t.as_tool_record()).collect();
                 let result = ListToolsResult {


matthewhaynesonline:
@EricLBuehler I ran into some hiccups but this is a really awesome idea! When following along with the docs (let me know if I missed anything obvious):

1. When cURL-ing against `/mcp/stream` the hyper server errors (maybe `404`?) 
2. When cURL-ing against `/sse`, it doesn't error but also doesn't return anything
3. When using the MCP inspector package, it is able to retrieve tools, but when I send an example chat it fails with `"missing field 'method'"`
4. Ctrl + C (SIGINT) seems to hang which I don't think happens on `master`? 

https://github.com/user-attachments/assets/198b5131-d56a-4fe0-88b5-d99ce05ca18c

mistralrs-server/src/main.rs
```diff
@@ -1,6 +1,8 @@
 use anyhow::Result;
 use clap::Parser;
 use mistralrs_core::{initialize_logging, ModelSelected, TokenSource};
+use rust_mcp_sdk::schema::LATEST_PROTOCOL_VERSION;
```
matthewhaynesonline, line 4: @EricLBuehler for my own curiosity, since there is a feature flag for version `"2025_03_26"` on the crate dep, will this `LATEST_PROTOCOL_VERSION` constant be pinned to that or will it be whatever the latest schema / proto version is (and could change in the future)?

EricLBuehler: @matthewhaynesonline we will need to keep the version of `rust_mcp_sdk` in sync with the latest schema/proto version.

EricLBuehler:
> @EricLBuehler I ran into some hiccups but this is a really awesome idea! When following along with the docs (let me know if I missed anything obvious):
>
>1) When cURL-ing against /mcp/stream the hyper server errors (maybe 404?)
>2) When cURL-ing against /sse, it doesn't error but also doesn't return anything
>3) When using the MCP inspector package, it is able to retrieve tools, but when I send an example chat it fails with "missing >field 'method'"
>4) Ctrl + C (SIGINT) seems to hang which I don't think happens on master?

@matthewhaynesonline I pushed some changes and these should all be fixed. 

We are using the new `Streamable HTTP` transport instead of the legacy SSE now, too!

PR #1465, opened 2025-06-12
chore: `Dockerfile` - Drop runtime rayon thread ENV

polarathene:
`ENV RAYON_NUM_THREADS` is not likely important to set a fixed value for for the runtime image. Dropping it as discussed in https://github.com/EricLBuehler/mistral.rs/pull/1458

---

This PR additionally revises `RUN` instructions for `apt-get`, moving `DEBIAN_FRONTEND=noninteractive` to an `ARG` and using HereDoc syntax for easier to grok multi-line sequence of commands.

The CPU `Dockerfile` had a redundant and empty `RUN`, so I've also removed that.
- [Originally introduced to add Python](https://github.com/EricLBuehler/mistral.rs/pull/249)
- [Later removed when Python was not required](https://github.com/EricLBuehler/mistral.rs/pull/303), however that PR left an empty install command since then.

**NOTE:** Presently the CUDA variant of the `Dockerfile` apparently requires Python again due to recent changes, but the actual cause [should be investigated and resolved](https://github.com/EricLBuehler/mistral.rs/pull/1459#issuecomment-2964541535) in future.

---

Some of the runtime image dependencies installed seem odd, I'm lacking time to investigate there but I'm under the impression that quite a few are unnecessary.

## Summary by CodeRabbit

- **Chores**
  - Updated Dockerfiles to improve clarity and maintainability of build and runtime instructions.
  - Refactored package installation steps for better readability and non-interactive builds.
  - Adjusted thread-related environment variables and build arguments for improved configuration in CUDA-enabled images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,6 @@ FROM debian:bookworm-slim AS runtime
 ENV HUGGINGFACE_HUB_CACHE=/data \
     PORT=80 \
     MKL_ENABLE_INSTRUCTIONS=AVX512_E4 \
-    RAYON_NUM_THREADS=8 \
     LD_LIBRARY_PATH=/usr/local/lib
 
 # Install only essential runtime dependencies and clean up
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,7 @@
+# syntax=docker/dockerfile:1
+
 # Stage 1: Build environment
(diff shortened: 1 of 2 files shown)


EricLBuehler (approved):
Thanks, makes sense!

PR #1483, opened 2025-06-19
Fix the usual String != &str bit

sempervictus:


sempervictus:
right... this _is_ your personal repo :-). Apologies, used to making pulls on folks' PRs into their GH so they trickle in a bit less obnoxiously

PR #1489, opened 2025-06-19
Add candle fork to crates.io

EricLBuehler:


polarathene:
Is there an issue getting changes from the fork upstreamed? Ideally you don't need to maintain a fork, especially if it starts to diverge and become a hassle to keep in sync over time.

EricLBuehler:
> Is there an issue getting changes from the fork upstreamed? Ideally you don't need to maintain a fork, especially if it starts to diverge and become a hassle to keep in sync over time.

Currently exploring various options to allow publishing on crates.io, this is one of them. The best option is using the official Candle of course though!

polarathene:
> The best option is using the official Candle of course though!

That didn't quite answer the question of why this was necessary in the first place 😅 

Are there PRs to candle for each feature in your fork, and you're just waiting on those to be reviewed/merged and released upstream? Will your fork be [at risk of drifting from upstream](https://github.com/EricLBuehler/candle/issues/84) too much?

PR #1525, opened 2025-06-28
Split Marlin and Paged Attention kernels for faster build

guoqingbao:
This PR addresses the slow build issue in Mistral.rs by splitting the Marlin and PagedAttention CUDA kernels into separate .cu files based on data type. This allows the kernels to be compiled in parallel, significantly improving overall project build performance.

As a result, the total build time can be reduced from over 10 minutes to under 3 minutes, depending on the host CPU.

---

### 🔍 How to Observe the Effects

To measure the build time improvement:

1. **Build the current version:**

   ```bash
   cargo clean
   cargo build --features cuda
   ```

2. **Apply this PR**, then rebuild:

   ```bash
   cargo clean
   cargo build --features cuda
   ```

You should notice a significant reduction in build time — typically from **10+ minutes** down to **less than 3 minutes**, depending on your host CPU.


sempervictus:
Looks like it works and even on a 96c xeon the improvement is appreciable. Thank you.

PR #1548, opened 2025-07-10
Add the capability to build for ios

rubiktubik:
This small change allows to build this library also for the target ios


EricLBuehler (approved):
Thanks!

Issue #1587, opened 2025-07-28
Conflicting initialization of tracing library


christer-eriksson:
## Describe the bug
mistral_core panic when tracing library is initialized from application including it

## Latest commit or version
95d362f8d0f487e3c999159b033f3deb028d4d83

Code to reproduce 

`
use anyhow::Result;
use mistralrs::{
    IsqType, PagedAttentionMetaBuilder, TextMessageRole, TextMessages, TextModelBuilder,
};
use tracing::info;

fn initialize_logging() {
    match tracing_subscriber::fmt()
        .with_max_level(tracing::Level::INFO)
        .try_init()
    {
        Ok(_) => println!("✅ Tracing initialized successfully"),
        Err(e) => println!("❌ Tracing initialization failed: {}", e),
    }
}

async fn run_mistral() -> Result<()> {
    let model = TextModelBuilder::new("HuggingFaceTB/SmolLM3-3B")
        .with_isq(IsqType::Q8_0)
        .with_logging()
        .with_paged_attn(|| PagedAttentionMetaBuilder::default().build())?
        .build()
        .await?;

    let messages = TextMessages::new()
        // .enable_thinking(false)
        .add_message(
            TextMessageRole::System,
            "You are an AI agent with a specialty in programming.",
        )
        .add_message(
            TextMessageRole::User,
            "Hello! How are you? Please write generic binary search function in Rust.",
        );

    let response = model.send_chat_request(messages).await?;

    println!("{}", response.choices[0].message.content.as_ref().unwrap());
    dbg!(
        response.usage.avg_prompt_tok_per_sec,
        response.usage.avg_compl_tok_per_sec
    );

    Ok(())
}

#[tokio::main]
async fn main() -> Result<()> {
    initialize_logging(); // 1
    info!("I can do the logging!");
    run_mistral().await.expect("Error in mistral"); //2
    Ok(())
}
`

Error 

`
thread 'main' panicked at /home/christer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-subscriber-0.3.19/src/fmt/mod.rs:517:14:
Unable to install global subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
[...]

PR #1588, opened 2025-07-28
fix: use try_init when initialize tracing

christer-eriksson:
https://github.com/EricLBuehler/mistral.rs/issues/1587


EricLBuehler (approved):
Thank you!

PR #1593, opened 2025-07-29
Add Blackwell 128x128 FP8 gemm

EricLBuehler:


sempervictus:
mmmm doughnuts :-).
Currently out of B200s "at work" and our own blackwells haven't arrived yet (vendor has suggested some possibility of accelerating our delivery as we actually work with them and their very large clients, will be happy to give them a shout-out in here and relevant spots if they do). Will be happy to test soon as we have slack CC10+ kit in our hands again.

PR #1598, opened 2025-07-31
Use smaller model in streaming example

jncraton:
This change adjusts the streaming example to use a more modern and much smaller model. The example will still produce a coherent response, but it should be roughly 10x faster and require downloading one tenth as much model data.


EricLBuehler (approved):
Thank you!

PR #1606, opened 2025-08-02
Improved Llama 4-style MoE performance

EricLBuehler:


EricLBuehler:
Already 46+ T/s vs llama.cpp ~22 T/s on GH200.

PR #1608, opened 2025-08-03
Fix cuda ring compilation

EricLBuehler:


ryanli:
This breaks the build for `features=+cuda,-ring`. Sent a fix in #1611.

PR #1614, opened 2025-08-04
Send the mcp server an initialization notification to let it know the client is done initializing plus finish off making all 3 supported mcp transports properly increment their request ids

sonicrules1234:
According to the MCP specification, the client needs to send this. Not sending this was keeping some mcp servers from working with mistral.rs.  Also finished off all 3 supported mcp transports properly increment their request ids


EricLBuehler (approved):
Thank you!

PR #1617, opened 2025-08-04
Update ISQ Python example

jncraton:
This change updates the Python ISQ example in the docs to load a non-quantized model. When running the current example, I get the following error message:

> ValueError: You are trying to in-situ quantize a GGUF model. This
> will not do anything.

This updated example works correctly for me.


EricLBuehler (approved):
Thank you!

PR #1650, opened 2025-10-11
Handle cpu dtype requirements for conv

EricLBuehler:

PR #1653, opened 2025-10-19
Handle case where gemma 3n q != (k=v) devices

EricLBuehler:
This could be triggered when the device mapping causes layers that reuse kv caches to receive mismatched devices.

@sempervictus this fixes the issues you reported in #1650.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,14 +114,14 @@ jobs:
           config: .typos.toml
 
   msrv:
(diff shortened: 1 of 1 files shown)


sempervictus:
Building.

I was seeing this on a _single_ laptop 4090 by the way, so not sure if device mapping logic is even applied here. Been building w/out `nccl` feature locally so i could offload to CPU but the OOM thing sort of precludes any real utilization for now.

PR #1677, opened 2025-10-29
Documentation book  for  Mistralrs & mistralrs-server

Ammar-Alnagar:
this includes 2 markdown files , one including the full documentation of the framework as a book style that works incrementally with the reader to allow them to understand the project better,

the second markdown is more of a guide on how to use mistralrs on a lower level in code aswell as serving models (like vllm LLMEngine and asyncEngine)


Ammar-Alnagar:
@EricLBuehler  let me know if there are any comments or changes needed.

PR #1711, opened 2025-11-10
[tmp] CUDA allocation log

EricLBuehler:


sempervictus:
catch22 situation - i need #1722 to build, but upstream lacks the debug capability. Maybe the memory tracing should be a PR on upstream's candle as well?

PR #1752, opened 2025-12-06
No longer getting cache issue from HF runs

### LINKED ISSUE #1749: Embedding pipeline unreachable cache method
Using the embedding loader and running a request:

```rust
mistralrs-core\src\pipeline\embedding.rs:706:9:
internal error: entered unreachable code
note: run with `RUST_BACKTRACE=1` environment variable to display a ba
``` 

This is the latest rev, which is working: 
```rust
mistralrs = { git = "https://github.com/EricLBuehler/mistral.rs", rev = "d946a96f0a37f2ddb2ef0bfa1f31d2d8ded44ce3", features = [
    "cuda",
] }
```

Cooksey99:
### The Issue

Embedding models in mistral.rs were completely broken due to a missing guard check in the engine code.

### Root Cause

Mistral.rs supports multiple types of models:
•  Text generation models - have a KV (key-value) cache for attention mechanisms
•  Hybrid models (Mamba-Attention) - have both KV cache and Mamba state
•  Embedding models - have NO cache at all (they just encode text to vectors)

The engine code had several locations checking if a model is "hybrid" by calling `.cache().is_hybrid()`, however, embedding models implement `.cache()` as `unreachable!()` because they don't have a cache:

```rust
impl CacheManagerMixin for EmbeddingPipeline {
    fn cache(&self) -> &EitherCache {
        unreachable!() // This is what's hitting and causing the panic
    }
}
```

The engine was calling `.cache()` without first checking if the model even has a cache, causing embedding models to panic with "entered unreachable code".

---

Fixes #1749

### DIFF
--- a/mistralrs-core/src/pipeline/macros.rs
+++ b/mistralrs-core/src/pipeline/macros.rs
@@ -291,48 +291,48 @@ macro_rules! get_embedding_paths {
         } else {
             $crate::api_get_file!(api, "modules.json", model_id)
         };
-        
+
         if modules_path.exists() {
             let modules: Vec<$crate::pipeline::EmbeddingModule> =
                 serde_json::from_str(&std::fs::read_to_string(&modules_path)?)?;
             for module in modules {
-            match module.ty {
-                $crate::pipeline::EmbeddingModuleType::Transformer => {
-                    parsed_modules.push($crate::pipeline::EmbeddingModulePaths::Transformer {
-                        path: module.path.clone(),
-                    });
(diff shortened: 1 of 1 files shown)


Cooksey99:
I have not gotten to do full testing of every scenario. I'll get to that when I can, or someone else is welcome to do so.

EricLBuehler (approved):
Thank you @Cooksey99! This was a bug introduced with the first mamba models.

PR #1771, opened 2025-12-13
Implement GGUF for Mistral3

### LINKED ISSUE #1754: Add support for Ministral-3-3B
Hi, not sure if it should be already supported but I tried

```
./target/release/mistralrs-server gguf -m . -f ~/Downloads/Ministral-3-3B-Instruct-2512-UD-Q4_K_XL.gguf
```

downloaded from https://huggingface.co/unsloth/Ministral-3-3B-Instruct-2512-GGUF

and I get the following error

```
2025-12-06T10:42:45.269043Z  INFO mistralrs_server_core::mistralrs_for_server_builder: avx: true, neon: false, simd128: false, f16c: true
2025-12-06T10:42:45.269085Z  INFO mistralrs_server_core::mistralrs_for_server_builder: Sampling method: penalties -> temperature -> topk -> topp -> minp -> multinomial
2025-12-06T10:42:45.269111Z  INFO mistralrs_server_core::mistralrs_for_server_builder: Model kind is: gguf quantized from gguf (no adapters)
2025-12-06T10:42:45.269169Z  INFO mistralrs_core::utils::tokens: Could not load token at "/home/wighawag/.cache/huggingface/token", using no HF token.
2025-12-06T10:42:45.269422Z  INFO mistralrs_core::utils::tokens: Could not load token at "/home/wighawag/.cache/huggingface/token", using no HF token.
2025-12-06T10:42:45.269443Z  INFO mistralrs_core::pipeline::paths: Loading `/home/wighawag/Downloads/Ministral-3-3B-Instruct-2512-UD-Q4_K_XL.gguf` locally at `/home/wighawag/Downloads/Ministral-3-3B-Instruct-2512-UD-Q4_K_XL.gguf`
2025-12-06T10:42:45.269472Z  INFO mistralrs_core::pipeline::gguf: GGUF file(s) ["/home/wighawag/Downloads/Ministral-3-3B-Instruct-2512-UD-Q4_K_XL.gguf"]
2025-12-06T10:42:45.269509Z  INFO mistralrs_core::pipeline::gguf: Prompt chunk size is 1024.
Error: unknown dtype for tensor 23
```

Cooksey99:
Mistral3 does not currently support GGUF files. Instead, you need to use ISQ via `--isq`, which makes it tougher to use on machines with fewer resources.

This PR does 2 things:
1. Implements GGUF architecture for Mistral3 models.
2. Adds useful error information when `candle` throws a `dtype` error.

Instead of a standard `dtype` error, you'll see something like this:
```
Error: Critical failure loading model part 0
Verify you are using a supported quantization type
Supported types: F32, F16, BF16, Q4_0, Q4_1, Q5_0, Q5_1, Q8_0, Q8_1, Q2K, Q3K, Q4K, Q5K, Q6K, Q8K
Candle error: unknown dtype for tensor 23
```
> *The supported types are pulled from the `candle` `GgmlDType` enums. Since it's private, there's not a good way to keep them in sync, so this will unfortunately not sync automatically if `candle` adds additional support. I'm open to a separate solution, but I really like the clear explanation.*

---
Fixes #1754

### DIFF
--- a/mistralrs-core/src/models/quantized_llama.rs
+++ b/mistralrs-core/src/models/quantized_llama.rs
@@ -348,8 +348,7 @@ impl TryFrom<ContentMetadata<'_>> for PropsGGUF {
     type Error = anyhow::Error;
 
     fn try_from(c: ContentMetadata) -> std::result::Result<Self, Self::Error> {
-        c.verify_arch("llama")?;
-        c.verify_arch("mistral3")?;
+        c.verify_arch_any(&["llama", "mistral3"])?;
 
         let required = [
             "attention.head_count",
--- a/mistralrs-core/src/pipeline/gguf.rs
+++ b/mistralrs-core/src/pipeline/gguf.rs
@@ -330,11 +330,9 @@ impl Loader for GGUFLoader {
 
         // If auto, convert to Map
         let num_layers = model.get_metadata()[&format!("{arch}.block_count")].to_u32()? as usize;
-<<<<<<< mistral3-gguf
 
-=======
         let mut max_kv_tokens: Option<usize> = None;
->>>>>>> master
+
         if let DeviceMapSetting::Auto(params) = mapper.clone() {
             let devices = device_map::get_all_similar_devices(device)?;
             // Initial dtype
--- a/mistralrs-core/src/utils/gguf_metadata.rs
+++ b/mistralrs-core/src/utils/gguf_metadata.rs
@@ -149,6 +149,12 @@ impl ContentMetadata<'_> {
 
         Ok(())
     }
+
+    pub fn verify_arch_any(&self, expected_arch: &[&str]) -> Result<()> {
+        expected_arch
+            .iter()
+            .try_for_each(|arch| self.verify_arch(arch))
+    }
 }
 
 // These traits below are a workaround for converting candles GGUF `Value` enum type wrapper.


philipmw:
I hit this issue today. Is there any remaining work here before you can merge this, @Cooksey99 ?

mistralrs-core/src/pipeline/gguf.rs
```diff
@@ -283,6 +282,9 @@ impl Loader for GGUFLoader {
             self.quantized_filenames.clone(),
             silent
         );
+
+        info!("About to load!");
```
EricLBuehler: Please remove debug logging

mistralrs-core/src/models/quantized_llama.rs
```diff
@@ -347,8 +347,6 @@ impl TryFrom<ContentMetadata<'_>> for PropsGGUF {
     type Error = anyhow::Error;
 
     fn try_from(c: ContentMetadata) -> std::result::Result<Self, Self::Error> {
-        c.verify_arch("llama")?;
```
EricLBuehler, line 351: Lets validate both to ensure: something like c.verify_arch_any(&["llama", "mistral3"])?;

EricLBuehler (changes requested):
Thamk you @Cooksey99 for the PR! Some small comments but largely looks good.

Cooksey99:
> I hit this issue today. Is there any remaining work here before you can merge this, @Cooksey99 ?

I appreciate you reaching out to check on this! @philipmw 

Those changes have been implemented. I didn't see an exact method to check `llama` and `mistral3` together, so it's just two lines for now. I'm not sure if it's something you added since this change, but that'd be nice to have if we need to add `devstral` or similar in the future. @EricLBuehler

mistralrs-core/src/models/quantized_llama.rs
```diff
@@ -348,6 +348,7 @@ impl TryFrom<ContentMetadata<'_>> for PropsGGUF {
 
     fn try_from(c: ContentMetadata) -> std::result::Result<Self, Self::Error> {
         c.verify_arch("llama")?;
```
EricLBuehler: `verify_arch` is an exact check, I think maybe something like `verify_arch_any` that checks if the gguf arch is any of the allowed ones would work better.

EricLBuehler (changes requested):
Thank you @Cooksey99! Just one more comment.

It looks like this PR also has a merge conflict with master. Once we resolve those two, I think this should be good to merge.

mistralrs-core/src/models/quantized_llama.rs
Cooksey99: Ah, I had brain fart trying to look at it while on vacation. I went ahead and added that method in the latest commit.

EricLBuehler: No worries!

Issue #1781, opened 2025-12-22
called `Option::unwrap()` on a `None` value  at mistralrs-core/src/paged_attention/block_engine.rs:148


grinapo:
## Describe the bug
```
thread '<unnamed>' (1004759) panicked at mistralrs-core/src/paged_attention/block_engine.rs:148:44:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: mistralrs_core::paged_attention::block_engine::Allocator<T>::allocate
   5: mistralrs_core::paged_attention::block_engine::BlockEngine::allocate_block_with_eviction
   6: mistralrs_core::paged_attention::block_engine::BlockEngine::append_token_slot_to_seq
   7: mistralrs_core::paged_attention::scheduler::PagedAttentionScheduler::schedule
   8: <mistralrs_core::paged_attention::scheduler::PagedAttentionScheduler as mistralrs_core::scheduler::Scheduler>::schedule
   9: mistralrs_core::engine::Engine::run::{{closure}}
  10: tokio::runtime::context::runtime::enter_runtime
  11: tokio::runtime::runtime::Runtime::block_on
```

It may be related to 
```
2025-12-22T14:46:50.586091Z ERROR mistralrs_core::engine: step - Model failed with error: DriverError(CUDA_ERROR_OUT_OF_MEMORY, "out of memory")
   0: candle_core::error::Error::bt
   1: <core::result::Result<O,E> as candle_core::cuda_backend::error::WrapErr<O>>::w
   2: mistralrs_quant::cublaslt::api::CublasLTBatchMatmul::fwd
   3: <mistralrs_quant::cublaslt::api::CublasLTBatchMatmul as candle_core::custom_op::CustomOp2>::cuda_fwd
   4: candle_core::storage::Storage::apply_op2
   5: candle_core::custom_op::<impl candle_core::tensor::Tensor>::apply_op2_arc
   6: candle_core::custom_op::<impl candle_core::tensor::Tensor>::apply_op2
   7: <mistralrs_quant::unquantized::UnquantLinear as mistralrs_quant::QuantMethod>::forward
   8: mistralrs_quant::MatMul::qmethod_matmul
   9: mistralrs_core::models::phi3::Model::forward
[...]

PR #1785, opened 2025-12-30
Fix panic on None value in block_engine.rs:148 (#1781)

EricLBuehler:
The block allocator's `allocate()` method previously called `.unwrap()` on
`free_blocks.pop()`, causing a panic when no free blocks were available.

This could occur when:
- GPU memory pressure caused eviction attempts to fail
- The evicted blocks still had refcount > 0 and weren't actually freed
- Edge cases where `can_allocate` check passed but actual allocation failed

Changes:
- Make `Allocator::allocate()` return `Option` instead of panicking
- Make `allocate_block_with_eviction()` return `Option` to propagate failure
- Update `BlockEngine::allocate()` to return `bool` indicating success/failure
- Update `append_token_slot_to_seq()` to return `Result` for error handling
- Update scheduler to gracefully handle allocation failures by retrying or
  preempting sequences instead of panicking

This allows the system to recover from transient memory pressure rather
than crashing.

### DIFF
--- a/mistralrs-core/src/paged_attention/block_engine.rs
+++ b/mistralrs-core/src/paged_attention/block_engine.rs
@@ -144,10 +144,10 @@ struct Allocator<T> {
 }
 
 impl<T> Allocator<T> {
-    fn allocate(&mut self) -> Arc<PhysicalTokenBlock> {
-        let block = self.free_blocks.pop().unwrap();
+    fn allocate(&mut self) -> Option<Arc<PhysicalTokenBlock>> {
+        let block = self.free_blocks.pop()?;
         block.deref_mut().refcount = 1;
-        block
+        Some(block)
     }
 
     fn free_block(&mut self, block: Arc<PhysicalTokenBlock>) {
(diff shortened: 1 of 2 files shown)


glaziermag:
Wanted to check maintainer intent on this one.

#1781 and this PR (#1785) propose making allocation paths fallible (instead of panicking) when eviction does not actually free a block. This PR is currently stale/dirty, and current master still has:

expect("Should have free blocks after eviction")
in [block_engine.rs](app://-/index.html#).

Were the later paged-attention refactors (#1808, #1829) intended to supersede this, or did this PR just fall out during churn?

If the concern is still valid, would you prefer:

reviving/rebasing #1785, or
a fresh minimal PR on current master with a regression test for the “no free block after eviction” path?

glaziermag:
Update to my February comment: the paged-attention rework has since removed `block_engine.rs` entirely — current master has `block_pool.rs`/`kv_cache_manager.rs`/`scheduler.rs` in its place, and a code search finds no remaining `expect("Should have free blocks after eviction")`. The panic path this PR (and #1781) targeted no longer exists in the cited form, so this can likely be closed against current master.

PR #1787, opened 2026-01-02
Fixing idx position bug in Qwen Vision Models

### LINKED ISSUE #1786: ERROR mistralrs_core::engine: completion step - Model failed with error: shape mismatch in where_cond, lhs: [3, 792], rhs: [3, 791]
## Describe the bug
```
ERROR mistralrs_core::engine: completion step - Model failed with error:
shape mismatch in where_cond, lhs: [3, 792], rhs: [3, 791]
```

It seems like this error occurs when processing multiple images sequentially with the same Qwen2_5-VL model instance in mistral.rs. I believe the shape mismatch suggests an issue with attention mechanism state (likely KV cache) not being properly reset between inference calls - but not sure.

The model being run is: https://huggingface.co/allenai/olmOCR-2-7B-1025 - which is "a release of the olmOCR model that's fine tuned from Qwen2.5-VL-7B-Instruct"

The processing loop looks something like
```
for (idx, image) in images.into_iter().enumerate() {
  ...
  match inference.process_image(image, prompt).await {
            Ok(text) => {
                info!("Page {} OCR result:\n{}", page_num, text);
                document_result.add_page(output::aggregator::PageResult {
                    page_number: page_num,
                    text,
                    metadata: None,
                });
            }
            Err(e) => {
                tracing::warn!("Failed to process page {}: {}", page_num, e);
                failed_pages.push(page_num);
                // Add placeholder for failed page
                document_result.add_page(output::aggregator::PageResult {
                    page_number: page_num,
                    text: format!("[ERROR: Failed to extract text from page {}]", page_num),
                    metadata: None,
                });
            }
        }
```

## Built for Apple M4
`MISTRALRS_METAL_PRECOMPILE=0 cargo build --release  `

Using the latest main
`mistralrs = { git = "https://github.com/EricLBuehler/mistral.rs.git", features = ["metal"] }`

It doesn't look like this is an issue with quantizing - the same error occurs with all quantizing options I've tried.

## Latest commit or version
Running on latest main.

schulte-lukas:
### Problem: Qwen2.5-VL models (like olmOCR) were failing during completion with:
```
ERROR mistralrs_core::engine: completion step - Model failed with error: 
shape mismatch in where_cond, lhs: [3, 1624], rhs: [3, 1623]
```

### Root Cause: 
During completion, input_ids contains only the new token (shape [1, 1]), but seqlens contains the full sequence length including KV cache. The position calculation in get_rope_index was using the small input_ids tensor, causing a mismatch with the attention mask created from seqlens. Fix:
1. Added completion mode detection (lines 390-402): When seqlens doesn't match input_ids dimensions, create a dummy tensor with the correct sequence length for position calculation

If there is a more elegant way to do this than creating a dummy tensor please let me know.

Fixes: https://github.com/EricLBuehler/mistral.rs/issues/1786

### DIFF
--- a/mistralrs-core/src/vision_models/qwen2_5_vl/mod.rs
+++ b/mistralrs-core/src/vision_models/qwen2_5_vl/mod.rs
@@ -357,6 +357,7 @@ impl Qwen2_5VLModel {
             )?);
         }
         let ropeidx_attn_mask = Tensor::stack(&ropeidx_attn_mask_bs, 0)?;
+
         let mut ropeidx_attn_mask_indices_bs = Vec::new();
         for len in seqlens.iter() {
             ropeidx_attn_mask_indices_bs.push(Tensor::from_vec(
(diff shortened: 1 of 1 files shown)


schulte-lukas:
Hey there @EricLBuehler - would you be able to give this a look?

PR #1793, opened 2026-01-04
Add utils.metal include to fused_glu.metal

vigsterkr:
this is required in order to get the metal files compiled.


EricLBuehler:
Thank you!

PR #1887, opened 2026-02-11
fix(core): use unix seconds for streaming chunk created timestamp

### LINKED ISSUE #1805: ChatCompletionChunkResponse "created" should be in seconds but is in milliseconds.
## Describe the bug
`maybe_send_streaming_response()` in [mistralrs-core/src/sequence.rs](https://github.com/EricLBuehler/mistral.rs/blob/5067d4f31be010a171f1c9fdf1106aaf456b1450/mistralrs-core/src/sequence.rs#L1441) defines a `ChatCompletionChunkResponse` like this:

```
ChatCompletionChunkResponse {
  id: seq.id.to_string(),
  choices: swap_streaming_chunks,
  created: seq.timestamp,
  model: model.clone(),
  system_fingerprint: SYSTEM_FINGERPRINT.to_string(),
  object: "chat.completion.chunk".to_string(),
  usage: usage_opt,
}
```

Where seq.timestap is in milliseconds, but the openai api states that it should be in seconds:
https://platform.openai.com/docs/api-reference/chat-streaming/streaming

## Latest commit or version
5067d4f31be010a171f1c9fdf1106aaf456b1450

glaziermag:
## Summary
This PR attempts to fix OpenAI compatibility for streaming responses by ensuring the `created` field is emitted in Unix **seconds** (not milliseconds).

## Problem
In streaming paths, `created` was set from `seq.timestamp`, which is initialized with `now.as_millis()`.
OpenAI-compatible `created` values should be Unix epoch seconds.

## Changes
- Updated streaming chat chunk response:
  - `created: seq.timestamp` -> `created: seq.creation_time() as u128`
- Updated streaming completion chunk response:
  - `created: seq.timestamp` -> `created: seq.creation_time() as u128`

File changed:

## Why this is correct
- `seq.creation_time()` is set from `SystemTime::now().duration_since(UNIX_EPOCH).as_secs()` at sequence creation.
- Non-streaming response paths already use second-based creation timestamps, so this makes behavior consistent across streaming and non-streaming APIs.

## Validation
- Ran:
  - `cargo check -p mistralrs-core`
- Result:
  - Passed

## References
- Attempts to fix #1805
- OpenAI API docs (streaming/completions `created` is Unix seconds)

### DIFF
--- a/mistralrs-core/src/sequence.rs
+++ b/mistralrs-core/src/sequence.rs
@@ -1562,7 +1562,7 @@ impl SequenceGroup {
                 .send(Response::Chunk(ChatCompletionChunkResponse {
                     id: seq.id.to_string(),
                     choices: swap_streaming_chunks,
-                    created: seq.timestamp,
+                    created: seq.creation_time() as u128,
                     model: model.clone(),
                     system_fingerprint: SYSTEM_FINGERPRINT.to_string(),
                     object: "chat.completion.chunk".to_string(),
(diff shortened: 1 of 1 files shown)


EricLBuehler (approved):
Thank you @glaziermag!

PR #1893, opened 2026-02-15
feat: add optional save_file for url image generation response format

setoelkahfi:
The `ImageGenerationResponseFormat::Url` option is not configurable. This PR adds optional `save_path`  to store the generated image. If the path is a directory, it will append the default name that is already set up.

### DIFF
--- a/mistralrs-core/src/pipeline/normal.rs
+++ b/mistralrs-core/src/pipeline/normal.rs
@@ -790,10 +790,10 @@ impl Loader for NormalLoader {
                 calibration_file.display(),
                 tokens.len()
             );
-            let bos_toks = chat_template.bos_tok().map(|b| vec![b]).unwrap_or_default();
-            let bos_tok_id = tokenizer
-                .token_to_id(&bos_toks[0])
-                .expect("Somehow the bos token is not present.");
+            let bos_tok_id = chat_template
+                .bos_tok()
+                .as_deref()
+                .and_then(|tok| tokenizer.token_to_id(tok));
 
             match self.config.organization {
                 IsqOrganization::Default => model.begin_track_stats()?,
@@ -804,7 +804,10 @@ impl Loader for NormalLoader {
             let n_chunks = tokens.len().div_ceil(CHUNK_SIZE);
             let start = Instant::now();
             for (i, chunk) in tokens.chunks(CHUNK_SIZE).enumerate() {
-                let chunk = [vec![bos_tok_id], chunk.to_vec()].concat();
+                let mut chunk = chunk.to_vec();
+                if let Some(bos_tok_id) = bos_tok_id {
+                    chunk.insert(0, bos_tok_id);
+                }
                 let chunk_len = chunk.len();
 
                 let start = Instant::now();
--- a/mistralrs-core/src/pipeline/vision.rs
+++ b/mistralrs-core/src/pipeline/vision.rs
@@ -684,10 +684,10 @@ impl Loader for VisionLoader {
                 calibration_file.display(),
                 tokens.len()
             );
-            let bos_toks = chat_template.bos_tok().map(|b| vec![b]).unwrap_or_default();
-            let bos_tok_id = tokenizer
-                .token_to_id(&bos_toks[0])
-                .expect("Somehow the bos token is not present.");
+            let bos_tok_id = chat_template
+                .bos_tok()
+                .as_deref()
+                .and_then(|tok| tokenizer.token_to_id(tok));
 
             // NOTE: We ONLY calibrate the text bits of these models!!
             // So only those should be tracked!
@@ -700,7 +700,10 @@ impl Loader for VisionLoader {
             let n_chunks: usize = tokens.len().div_ceil(CHUNK_SIZE);
             let start = Instant::now();
             for (i, chunk) in tokens.chunks(CHUNK_SIZE).enumerate() {
-                let chunk = [vec![bos_tok_id], chunk.to_vec()].concat();
+                let mut chunk = chunk.to_vec();
+                if let Some(bos_tok_id) = bos_tok_id {
+                    chunk.insert(0, bos_tok_id);
+                }
                 let chunk_len = chunk.len();
 
                 let start = Instant::now();


mistralrs-core/src/request.rs
```diff
@@ -24,13 +24,72 @@ pub enum Constraint {
...
-#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass(eq, eq_int))]
-#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass(eq))]
 /// Image generation response format
 pub enum ImageGenerationResponseFormat {
-    Url,
-    B64Json,
+    Url { path: Option<PathBuf> },
+    B64Json(),
```
EricLBuehler: I think this should stay as a unit variant `B64Json`. Adding empty parens makes it a tuple variant, which changes serde  serialization from "B64Json" to {"B64Json":[]} and breaks existing clients.

Was there a reason that you changed this? If not, I think reverting it would be better.

mistralrs-core/src/request.rs
```diff
@@ -24,13 +24,72 @@ pub enum Constraint {
...
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
-#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass(eq, eq_int))]
-#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass(eq))]
 /// Image generation response format
 pub enum ImageGenerationResponseFormat {
-    Url,
-    B64Json,
+    Url { path: Option<PathBuf> },
```
EricLBuehler: Changing this from a unit variant to a struct variant breaks serde compatibility, as `Url` no longer deserializes. Clients now need {"Url":{"path":null}}.

Perhaps we can add `save_path: Option<PathBuf>` as a separate field on `RequestMessage::ImageGeneration instead`, which avoids all the serde/PyO3/utoipa breakage and lets this enum keep Copy?

mistralrs-core/src/pipeline/response.rs
```diff
@@ -25,19 +25,37 @@ pub async fn send_image_responses(
...
+                        if p.is_dir() || p.to_string_lossy().ends_with(std::path::MAIN_SEPARATOR) {
+                            p.join(&default_name)
+                        } else {
+                            p
+                        }
+                    }
+                    None => PathBuf::from(&default_name),
+                };
+                // Ensure parent directories exist
+                if let Some(parent) = saved_path.parent() {
```
EricLBuehler: This is a path traversal risk when exposed through the HTTP server API, as a remote client could write to arbitrary filesystem locations. If path stays, it should be restricted to the Rust/Python SDK and not accepted from the HTTP endpoint, or sanitized to prevent path traveral.

mistralrs-core/src/request.rs
setoelkahfi: Actually, this sounds cleaner.

mistralrs-core/src/request.rs
setoelkahfi: This was `rustc` suggestion:
```
error: Unit variant `B64Json` is not yet supported in a complex enum
       = help: change to an empty tuple variant instead: `B64Json()`
       = note: the enum is complex because of non-unit variant `Url`
  --> mistralrs-core/src/request.rs:33:5
```

mistralrs-core/src/request.rs
setoelkahfi: Looks like `DiffusionGenerationParams` is the better place. What do you think? I have'nt explore the pyo3 binding yet, so not familiar with the design.

setoelkahfi: This satisfies pyo3:

```rust
#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass)]
#[cfg_attr(feature = "pyo3_macros", pyo3(get_all))]
#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
pub struct DiffusionGenerationParams {
    pub height: usize,
    pub width: usize,
    pub save_path: Option<PathBuf>,
}

generate_repr!(DiffusionGenerationParams);

impl Default for DiffusionGenerationParams {
    /// Image dimensions will be 720x1280.
    fn default() -> Self {
        Self {
            height: 720,
            width: 1280,
            save_path: None,
        }
    }
}

```

Besides, I have another small ergonomic change there if it's ok:
```rust

impl DiffusionGenerationParams {
    /// Square image
    pub fn square(size: usize) -> Self {
        Self {
            height: size,
            width: size,
        }
    }
}

```

mistralrs-core/src/request.rs
setoelkahfi: which apparently comes from [pyo3](https://github.com/PyO3/pyo3/issues/3749)

mistralrs-core/src/pipeline/response.rs
```diff
@@ -28,7 +28,10 @@ pub async fn send_image_responses(
             .unwrap_or(ImageGenerationResponseFormat::Url)
         {
             ImageGenerationResponseFormat::Url => {
-                let saved_path = format!("image-generation-{}.png", Uuid::new_v4());
+                let saved_path = match seq.image_gen_save_path() {
+                    Some(path) => path.to_string_lossy().into_owned(),
+                    None => format!("image-generation-{}.png", Uuid::new_v4()),
+                };
```
setoelkahfi, line 34: I cleaned up my branch @EricLBuehler. The url path now should point to an image name. Open for suggestion.

mistralrs-core/src/request.rs
```diff
@@ -91,6 +91,7 @@ pub enum RequestMessage {
         prompt: String,
         format: ImageGenerationResponseFormat,
         generation_params: DiffusionGenerationParams,
+        save_path: Option<PathBuf>,
```
EricLBuehler: Maybe rename this one to `save_file` to make the purpose a bit clearer?

mistralrs/examples/flux/main.rs
```diff
@@ -23,6 +23,7 @@ async fn main() -> Result<()> {
             "A vibrant sunset in the mountains, 4k, high quality.".to_string(),
             ImageGenerationResponseFormat::Url,
             DiffusionGenerationParams::default(),
+            None,
```
EricLBuehler, line 26: 👌

mistralrs-core/src/request.rs
setoelkahfi: Done.

Issue #1827, opened 2026-01-19
Modular cargo features in Mistral.rs


copumpkin:
I just pulled in mistral to my project and my Cargo.lock got a mile longer, and I seem to have a bunch of stuff I'm not using. I needed to do some simple in-process inference using the Metal backend, but I don't need the MCP support, or audio support, or the server. I'm wondering if there'd be any interest in creating cargo features representing that functionality so that e.g., audio-only dependencies become conditional on whether you're using the audio functionality.

I could contribute a PR adding some if the maintainers are open to it.

krampenschiesser:
Might be worth taking a look at [candle](github.com/huggingface/candle), the underlying library. It also supports inference.

EricLBuehler:
Hi @copumpkin! Would be happy to accept a PR to make this by default on, but you can compile them out. That sounds like an excellent addition.

PR #1894, opened 2026-02-16
Modular cargo features for mistralrs-core (audio/mcp)

glaziermag:
Addresses #1827 by making the audio and MCP integrations optional without changing the default build.

## Summary

- Add default-on `audio` and `mcp` features to `mistralrs-core`.
- Forward both features from the top-level `mistralrs` crate, also default-on.
- Make `mistralrs-audio` and `mistralrs-mcp` optional so `--no-default-features` removes their runtime dependency graphs.
- Extract the current tool-calling API into `mistralrs-tool-types`, used by core, MCP, and code execution. This avoids duplicate public types and prevents `mistralrs-code-exec` from pulling MCP back into minimal builds.
- Compile MCP client/configuration APIs and public audio request/building APIs out when their features are disabled. No dummy public compiled-out types are exposed.
- Keep only a private internal audio placeholder where model plumbing requires a common internal shape.
- Keep Phi-4 Multimodal, Gemma 3n, and Gemma 4 usable for their non-audio modalities in no-audio builds; attempted audio paths return an explicit `audio`-feature-required error.
- Document the feature matrix and minimal-build examples in `docs/src/content/docs/reference/cargo-features.md`.

This branch was reconstructed directly on current `master` (`b7956205fae93f93feed2b87237e752ee8083b56`). It intentionally excludes the stale branch's CI/configuration churn, deleted legacy feature document, and unrelated lint fixes.

## Validation

Validated locally on 2026-07-12 with Rust/Cargo 1.97.0.

Formatting and build matrix:

```bash
cargo fmt --all -- --check
git diff --check origin/master
cargo check -p mistralrs-core --no-default-features
cargo check -p mistralrs --no-default-features
cargo check -p mistralrs-core --no-default-features --features audio
cargo check -p mistralrs-core --no-default-features --features mcp
cargo check -p mistralrs --no-default-features --features audio
cargo check -p mistralrs --no-default-features --features mcp
cargo check -p mistralrs-core
cargo check -p mistralrs
cargo check --workspace
cargo test -p mistralrs-core --no-default-features --lib --no-run
cargo test -p mistralrs-code-exec --no-run
```

Focused tests:

```bash
cargo test -p mistralrs-core --no-default-features --lib modular_feature_tests -- --nocapture
# 8 passed

cargo test -p mistralrs-core --no-default-features --lib no_audio_tests -- --nocapture
# 2 passed

cargo test -p mistralrs-core --no-default-features --lib speech_arch_reports_audio_feature_requirement_when_audio_disabled -- --nocapture
# 1 passed

cargo test -p mistralrs-tool-types
# 3 unit tests passed; doc tests passed
```

Dependency-graph audit:

```bash
cargo tree -p mistralrs-core --no-default-features -e normal --prefix none
```

The no-default normal graph contains none of `mistralrs-audio`, `symphonia`, `hound`, `mistralrs-mcp`, or `rust-mcp-schema`. The default graph contains all five, confirming backward-compatible defaults.

## Clippy note

`cargo clippy -p mistralrs-core --no-default-features -- -D warnings` is not claimed as passing on Rust 1.97.0: it stops on `clippy::unneeded_wildcard_pattern` at the unchanged current-master line `mistralrs-core/src/pipeline/loaders/multimodal_loaders.rs:5320`. Allowing that one lint exposes additional unchanged current-master Rust 1.97 lints. Those unrelated baseline fixes are deliberately outside this PR.

## Hosted CI

GitHub Actions for head `8418d630ee38d6603c0eddefb1ca72c550552799` completed as follows:

- Passed: Rustfmt, Typos, doc links, Rust docs, generated-doc reproducibility, docs build, Metal check, and stable checks on Ubuntu, macOS, and Windows.
- Passed: stable test suites on Ubuntu, macOS, and Windows.
- Skipped as configured: CUDA and docs deploy.
- Failed only: Clippy, on the four unchanged current-master Rust 1.97 diagnostics described above. The hosted log reports no branch-added Clippy diagnostic.


docs/CARGO_FEATURES.md
```diff
@@ -286,6 +286,33 @@ These features are primarily for library development and are not typically used
 
 ---
 
+## Modular Core Features (`mistralrs-core`)
```
EricLBuehler: I believe that it would also be beneficial/useful to expose this in the `mistralrs` crate as its the main Rust usage point, and update docs accordingly.

Right now there's no way to disable audio/mcp for `mistralrs`, I think propagating the feature flags there would be good.

mistralrs-core/Cargo.toml
```diff
@@ -112,7 +111,22 @@ objc2-metal = { workspace = true, optional = true }
...
+    "dep:rubato",
+    "dep:rustfft",
+    "dep:hound",
+    "dep:apodize",
+]
+
+# MCP (Model Context Protocol) integration (client + types).
+mcp = ["dep:mistralrs-mcp"]
+
+pyo3_macros = ["pyo3", "mcp", "mistralrs-mcp/pyo3_macros"]
```
EricLBuehler, line 129: This means enabling pyo3_macros force-enables mcp. This is correct for the Python SDK path (which needs MCP types registered as pyclasses), but it's worth noting that the `tool_types.rs`
fallback types carry `#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass)]` annotations.

These would never fire in practice, since pyo3_macros implies mcp, and when mcp is on the fallback
types aren't compiled. Those annotations are likely dead code as I dont think having pyo3 without the mcp compiled in makes sense.

mistralrs-core/src/tool_types.rs
```diff
@@ -0,0 +1,93 @@
...
+/// Called function with name and arguments.
+#[cfg_attr(feature = "pyo3_macros", pyo3::pyclass)]
+#[cfg_attr(feature = "pyo3_macros", pyo3(get_all))]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct CalledFunction {
+    pub name: String,
+    pub arguments: String,
+}
+
+#[cfg(all(test, feature = "mcp"))]
```
EricLBuehler: Same as the other comment; this looks like dead code as this crate is only included when we dont have the mcp feature.

mistralrs-core/src/tool_types.rs
```diff
@@ -0,0 +1,93 @@
+use serde::{Deserialize, Serialize};
```
EricLBuehler: This module duplicates items from `mistralrs-mcp`, which will make ensuring parity a bit harder. Perhaps we can find a better solution that doesnt have module duplication?

mistralrs-core/src/lib.rs
```diff
@@ -96,13 +111,59 @@ pub use device_map::{
     DeviceLayerMapMetadata, DeviceMapMetadata, DeviceMapSetting, LayerDeviceMapper,
 };
 pub use gguf::{GGUFArchitecture, GGUF_MULTI_FILE_DELIMITER};
+
+#[cfg(feature = "audio")]
 pub use mistralrs_audio::AudioInput;
+#[cfg(not(feature = "audio"))]
+#[derive(Clone, Debug, Default)]
+pub struct AudioInput {
```
EricLBuehler: Same comment with the tool types, duplicating types is suboptimal I think.

For uniformity, I think that we just shouldnt provide audio/mcp related items if they are compiled out. What do you think?

Some of these types are used internally of course, so maybe just make a new module to put the dummy types into, and make them `pub(crate)`?

mistralrs-core/src/lib.rs
```diff
@@ -96,13 +111,59 @@ pub use device_map::{
...
+    CalledFunction, Function, Tool, ToolCallback, ToolCallbackWithTool, ToolType,
+};
+
+#[cfg(feature = "mcp")]
 pub use mistralrs_mcp::{
     McpClient, McpClientConfig, McpServerConfig, McpServerSource, McpToolInfo,
 };
+#[cfg(not(feature = "mcp"))]
+#[derive(Clone, Debug, Default)]
+pub struct McpClientConfig {
```
EricLBuehler: Same as above. I think maybe its best to just not provide dummy mcp-related things if the user compiled it out. 

For instance: the `mistralrs_mcp::McpClientConfig` has several public fields. If a user constructs this `McpClientConfig`, their code would break.

mistralrs-core/src/pipeline/loaders/vision_loaders.rs
```diff
@@ -169,6 +175,7 @@ pub enum VisionLoaderType {
     Idefics3,
     #[serde(rename = "minicpmo")]
     MiniCpmO,
+    #[cfg(feature = "audio")]
```
EricLBuehler: I think that we should still be able to use each of these models without audio.

Perhaps just update the `mistralrs-core` audio input / each model's audio input processing steps to be disabled or error if audio date is provided?

mistralrs-core/src/tuning.rs
```diff
@@ -483,7 +484,11 @@ pub fn auto_tune(req: AutoTuneRequest) -> Result<AutoTuneResult> {
         | ModelSelected::XLoraGGML { .. } => {
             anyhow::bail!("Auto-tuning is not supported for pre-quantized GGUF/GGML models.");
         }
-        ModelSelected::DiffusionPlain { .. } | ModelSelected::Speech { .. } => {
+        ModelSelected::DiffusionPlain { .. } => {
+            anyhow::bail!("Auto-tuning is not supported for diffusion or speech models.");
```
EricLBuehler: Since `ModelSelected::DiffusionPlain` is seperate the error message should be too. Same below for the speech variant.

mistralrs-core/src/model_selected.rs
EricLBuehler: CI docs warning here: "this URL is not a hyperlink". Need to wrap it in `<...>`

mistralrs-core/src/lib.rs
```diff
@@ -34,6 +34,18 @@ pub const MISTRALRS_GIT_REVISION: &str = match option_env!("MISTRALRS_GIT_REVISI
     None => "unknown",
 };
 
+#[cfg(not(feature = "mcp"))]
+fn warn_if_mcp_config_ignored(mcp_client_config: &Option<McpClientConfig>) -> bool {
```
EricLBuehler: Let's put this into a util module to avoid clutter in `lib.rs`.

EricLBuehler (changes requested):
Thanks for the PR @glaziermag!

I made several comments throughout.

Largely, it looks good. The approach of gating at the loader/pipeline level is elegant, but I made some comments regarding the API surface.

My feeling is that we should not provide dummy audio/mcp types when a user intentionally compiles that feature out. For enums, I think it's fine if we compile certain variants out too, but the loader type variant is an exception, as you should still be able to use the model without audio.

docs/CARGO_FEATURES.md
glaziermag: Good call. mistralrs now forwards audio/mcp feature flags into mistralrs-core, and docs were updated in [CARGO_FEATURES.md](app://-/index.html#).

mistralrs-core/Cargo.toml
glaziermag: Thank you, that’s a fair point.  I moved shared tool types/attrs into a shared crate and kept pyo3_macros linkage aligned with mcp, so we don’t expose accidental MCP-off pyclass surfaces.

mistralrs-core/src/tool_types.rs
glaziermag: I aligned the cfgs and reduced feature-off dead paths; no-default clippy with -D warnings now passes for both mistralrs-core and mistralrs.

mistralrs-core/src/lib.rs
glaziermag: Ok.  I introduced mistralrs-tool-types as a single source of truth, and both core and MCP now re-export shared definitions instead of duplicating them.

mistralrs-core/src/tool_types.rs
glaziermag: Thanks-I addressed this by extracting shared types so we no longer duplicate module shapes between core and MCP, which should reduce parity drift.

mistralrs-core/src/lib.rs
glaziermag: Thank you, I agree with this direction. MCP-facing exports are now properly feature-gated under mcp, and dummy public stand-ins were removed.

mistralrs-core/src/tuning.rs
glaziermag: Thank you, agreed. I split the auto-tune errors so diffusion and speech each have dedicated messages, and added tests in `mistralrs-core/src/tuning.rs`.

mistralrs-core/src/model_selected.rs
glaziermag: Good catch, thank you. I wrapped the URL in angle brackets to satisfy the rustdoc hyperlink lint in [model_selected.rs](app://-/index.html#).

mistralrs-core/src/lib.rs
glaziermag: Thanks for the suggestion. I reduced [lib.rs](app://-/index.html#) clutter by moving fallback/internal wiring behind feature-gated modules and keeping the top-level public surface focused.

mistralrs-core/src/pipeline/loaders/vision_loaders.rs
glaziermag: Thanks.  I updated phi4mm and gemma3n to remain selectable/loadable in no-audio builds, and no longer hard-fail loader selection when audio is disabled.

The no-audio behavior now focuses on audio-input handling constraints rather than compiling out those model paths. I also updated the no-audio tests to assert loader selection succeeds for both architectures.

glaziermag: Updated this thread to match your stronger intent: in no-audio builds, `VisionLoaderType` still includes `Phi4MM`/`Gemma3n`, and `AutoVisionLoader::get_loader(...)` now selects those loaders instead of hard-failing. I also updated tests to assert no-audio loader selection succeeds (`loader_selects_without_audio_feature`), and verified with `cargo test -p mistralrs-core --no-default-features loader_selects_without_audio_feature -- --nocapture` (2/2 passing). Audio request/decode behavior remains feature-gated; this change is specifically to keep these model architectures usable without audio enabled.

glaziermag:
Testing validated: PR compiled and 'no-default-features' test suite passed successfully on vm.

glaziermag:
@EricLBuehler Hi — just checking in! I addressed all 10 review comments back in February (extracted shared types into `mistralrs-tool-types`, removed dummy exports, split auto-tune error messages, wrapped the URL lint, etc.). Would you be able to take another look when you get a chance? Happy to make further changes if needed.

glaziermag:
Hi @EricLBuehler — acknowledging your review feedback from Feb 17 about not providing dummy audio/mcp types when those features are compiled out. You're right that the current approach of stub types undermines the compile-time guarantees that `#[cfg]` feature gating should provide.

I plan to rework this to:
1. Use `#[cfg(not(feature = "audio"))]` to compile out the enum variants entirely rather than providing empty stubs
2. Keep the loader-level gating but remove the dummy type wrappers
3. Ensure the API surface accurately reflects which features are available at compile time

This is a substantial rework given the branch has diverged significantly (41 commits each way), so it will take a dedicated follow-up. Leaving this open as a tracking point.

glaziermag:
**Housekeeping note:** This branch currently bundles the CI fix from #2115 (`.typos.toml`, `openapi_doc.rs`, `distributed/layers.rs`). Once #2115 is merged, this branch will need a rebase onto updated `master` to drop the duplicate CI fix commit and resolve the resulting conflicts.

glaziermag:
Marked ready for review. Validation evidence and narrowed claim wording are already attached in the PR discussion/body. This PR is ready under the scoped claim described in the PR.

Ready for maintainer review under the narrowed claim in the PR body. This is a targeted/invariant fix and should not be read as full closure of the broader linked issue unless the PR body explicitly says so.

glaziermag:
@EricLBuehler This is ready for re-review. All ten comments from your February review were addressed at the time (see the threaded replies); the branch is now synced with current master, with the feature gating re-derived onto the code that changed since — the #2354 scheduler/sequence rewrite, the reworked gemma3n/gemma4/phi4 inputs processors (the audio-off bail now composes with the `supports_audio` runtime guard added in July), and the new lora/quant exports.

Validation on Rust 1.97.0 at this head:

```bash
cargo clippy -p mistralrs-core -p mistralrs-server-core -p mistralrs-tool-types -p mistralrs-mcp -p mistralrs-code-exec --tests -- -D warnings   # clean
cargo clippy -p mistralrs-core --no-default-features --tests -- -D warnings                                                                      # clean
cargo check  -p mistralrs-core --no-default-features                                                                                             # compiles
cargo fmt --all -- --check                                                                                                                       # clean
cargo test -p mistralrs-core --no-default-features modular_feature   # 9 passed
cargo test -p mistralrs-core --no-default-features no_audio          # 4 passed
```

Default features still include `audio` + `mcp`, so downstream crates and pyo3 are unaffected. One note: comment 7's file has since been renamed `vision_loaders.rs` → `multimodal_loaders.rs`; the loader-selection tests moved with it.

Issue #1906, opened 2026-02-19
ERROR mistralrs_core::engine: prompt step - Model failed with error: Error while loading function: "Error while loading function: fused_glu_bfloat"


julienlesbegueriesperso:
## Describe the bug
I've got a bug while trying to chat with a model.

### Installation suceeded : 
Command : `curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/EricLBuehler/mistral.rs/master/install.sh | sh`

info: Detected OS: macos
info: Rust is installed: rustc 1.93.0 (254b59607 2026-01-19)

info: Detecting hardware capabilities...
info: macOS detected - enabling metal

Installation Summary
====================
Features: metal

Proceed with installation? [Y/n]

info: Installing mistralrs-cli with features: metal
     Ignored package `mistralrs-cli v0.7.0` is already installed, use --force to override

success: mistral.rs installed successfully!

Device : Mac Book Pro with M3 96Go RAM
Command typed : mistralrs run -m Qwen/Qwen3-4B
The installation worked well but while sending a message I got:

Welcome to interactive mode! Because this model is a text model, you can enter prompts and chat with the model.

Commands:
- `\help`: Display this message.
- `\exit`: Quit interactive mode.
- `\system <system message here>`:
    Add a system message to the chat without running the model.
    Ex: `\system Always respond as a pirate.`
- `\clear`: Clear the chat history.
- `\temperature <float>`: Set sampling temperature (0.0 to 2.0).
- `\topk <int>`: Set top-k sampling value (>0).
- `\topp <float>`: Set top-p sampling value in (0.0 to 1.0).
====================
> coucou
2026-02-19T08:42:56.749684Z ERROR mistralrs_core::engine: prompt step - Model failed with error: Error while loading function: "Error while loading function: fused_glu_bfloat"
[...]

PR #1914, opened 2026-02-19
fix(quant): ensure Metal BF16 fused_glu symbol is always emitted

glaziermag:
## Summary
Attempts to fix Metal BF16 fused GLU kernel load failures reported in #1906 ().

## Changes
- Ensure BF16 kernel symbol  is always emitted in .
- Use  for BF16 specialization so symbol emission does not depend on native  availability.
- Refactor fused GLU instantiation macros to a named-instantiation helper for consistency and lower regression risk.

## Why
Rust runtime selects  for BF16 in Metal backend, so the Metal library must always export this exact host name.

### DIFF
--- a/mistralrs-quant/src/utils/ops.rs
+++ b/mistralrs-quant/src/utils/ops.rs
@@ -3527,64 +3527,6 @@ mod tests {
         );
     }
 
-    #[cfg(all(feature = "metal", target_os = "macos"))]
-    #[test]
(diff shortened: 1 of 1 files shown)


EricLBuehler:
Also, since PR https://github.com/EricLBuehler/mistral.rs/pull/1861 we always enforce bfloat16 support.

glaziermag:
Thanks for the review and clarification. You're right to point to the existing BF16 enforcement path. I'll close this PR.

Issue #1897, opened 2026-02-17
`Qwen/Qwen2.5-3B-Instruct`  works on debug app, error on TestFlight build


setoelkahfi:
## Describe the bug

I'm using Tauri to build my macOS app using mistral.rs as the on-device inference library. When running on the debug app, the development app, this model works fine, load fast, and no issue. Bun when running the TestFlilght release, it loads as fast as debug, but spits out error when chatting.

`Chat inference failed: ChatModelError { msg: "Error while loading function: \"Error while loading function: fused_glu_float\"", incomplete_response: ChatCompletionResponse { id: "1", choices: [Choice { finish_reason: "error", index: 0, message: ResponseMessage { content: Some(""), role: "assistant", tool_calls: None, reasoning_content: None }, logprobs: None }], created: 1771361833, model: "Qwen/Qwen2.5-3B-Instruct", system_fingerprint: "local", object: "chat.completion", usage: Usage { completion_tokens: 0, prompt_tokens: 116, total_tokens: 116, avg_tok_per_sec: 29000.0, avg_prompt_tok_per_sec: 0.0, avg_compl_tok_per_sec: 0.0, total_time_sec: 0.004, total_prompt_time_sec: 0.0, total_completion_time_sec: 0.0 } } }`

I bumped my minimum system version from the Tauri default 10.13 to 14.0, that's the mininmum version for Metal 3, still no luck.

https://support.apple.com/en-us/102894

## Latest commit or version
Which commit or version you ran with.

https://github.com/setoelkahfi/mistral.rs/tree/feature/00-image-generation-path

EricLBuehler:
Hi @setoelkahfi! Thank you for the issue. I merged #1898 which should fix this. Can you please confirm that it is resolved?

setoelkahfi:
> Hi @setoelkahfi! Thank you for the issue. I merged #1898 which should fix this. Can you please confirm that it is resolved?

Works like a charm!

EricLBuehler:
Glad I could help! Let me know if you have any other issues.

setoelkahfi:
@EricLBuehler This [issue](https://github.com/EricLBuehler/mistral.rs/issues/1903) could be related, since I'm using a shared code base.

PR #1915, opened 2026-02-20
fix: `black-forest-labs/FLUX.1-schnell` crashes with narrow-out-of-bounds panic

setoelkahfi:
Related https://github.com/EricLBuehler/mistral.rs/issues/1897

When running image generation with exeeding prompt length, it crashes the app. 

- Clamp CLIP input to max_position_embeddings to prevent OOB panic
- Ensure CLIP tokenizer truncates to model's max sequence length (77). 
- Add runtime clamp as a safety check to avoid narrow-out-of-bounds panic in ClipTextEmbeddings when prompts exceed this limit.


EricLBuehler (changes requested):
Thank you for the PR @setoelkahfi! However, is there a cleaner way to solve this that does not involve truncation (which might cause unintended breakage)?

PR #1916, opened 2026-02-20
fix(gguf): `verify_arch_any` used AND logic instead of OR

n-engine:
## Summary
- `verify_arch_any` used `try_for_each` which requires ALL architectures to match (AND), instead of matching ANY one (OR)
- This caused GGUF models with `llama` architecture to fail with `Expected "mistral3" architecture, got "llama"` when the loader accepts both `["llama", "mistral3"]`
- Replace with `any()` check so matching one expected architecture suffices

## Test plan
- [x] Load a Llama 3.1 8B GGUF (architecture: `llama`) through the quantized_llama loader which calls `verify_arch_any(&["llama", "mistral3"])`
- [ ] Verify Mistral3 GGUF models still load correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mistralrs-core/src/utils/gguf_metadata.rs
+++ b/mistralrs-core/src/utils/gguf_metadata.rs
@@ -150,10 +150,19 @@ impl ContentMetadata<'_> {
         Ok(())
     }
 
-    pub fn verify_arch_any(&self, expected_arch: &[&str]) -> Result<()> {
-        expected_arch
-            .iter()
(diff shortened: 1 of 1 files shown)


starpit:
we ran into the same issue, thanks for the PR.
our issue stems from #1771 

https://github.com/EricLBuehler/mistral.rs/pull/1771/changes#diff-da1e6f56f0e565985ccaa246f41d45f33271525bb3ae0d3a776cb282ce797676R351

Issue #1799, opened 2026-01-04
error loading functiongemma


vigsterkr:
## Describe the bug
i'm trying to load [functiongemma](https://huggingface.co/unsloth/functiongemma-270m-it), which as far as i understand based in mistral.rs should be loaded as a vision model, but once i did that i'm getting:
```
"A weight is negative, too large or not a valid number"
```

any ideas? moreover, I wonder why only text models are supported with gguf? i.e. when i tried to load the gguf version of functiongemma, then i've gotten:
```
"gguf loading is only supported for text models"
```

## Latest commit or version
0ff79d4e83a5fdb48c61803a3ba88c58a6d905f6

PR #1931, opened 2026-02-22
fix(core): reject text-only Gemma3 configs in AutoVisionLoader

glaziermag:
Issue: https://github.com/EricLBuehler/mistral.rs/issues/1927

attempts to fixes #1927

## Summary
- reject text-only Gemma3 configs in AutoVisionLoader instead of routing them through the vision path
- add early guards for `model_type` values ending in `_text` (including `gemma3_text`)
- add an architecture guard for `Gemma3ForCausalLM` with a helpful remediation message
- add a unit test that verifies `gemma3_text` is rejected with guidance to use the text loader

## Before and After
Before:
- AutoVisionLoader could accept text-only Gemma3 configs and continue down the vision pipeline.
- This could fail later with confusing weight/config errors.

After:
- AutoVisionLoader now rejects these configs immediately with an actionable error that points to the text loader path.

## Validation
- `cargo check -p mistralrs-core` passed

### DIFF
--- a/mistralrs-core/src/pipeline/loaders/vision_loaders.rs
+++ b/mistralrs-core/src/pipeline/loaders/vision_loaders.rs
@@ -287,10 +287,32 @@ impl std::fmt::Display for VisionLoaderType {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::AutoVisionLoader;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


glaziermag:
Closing as duplicate in favor of #1932, which is scoped to issue #1799 and includes the proper auto-close linkage ().

PR #1932, opened 2026-02-22
fix(core): route Gemma 3 variants by parsed config in AutoVisionLoader

glaziermag:
## Summary

Gemma 3 has both text-only and multimodal checkpoints. AutoVisionLoader previously treated Gemma 3 as multimodal based on name/model_type heuristics, which could misclassify text-only Gemma 3 (e.g., gemma3_text / FunctionGemma / Gemma 3 1B) and send it down the vision pipeline.

This change makes Gemma 3 auto loading config-modality-aware by parsing Gemma3Config and routing/labeling modalities accordingly.

## What changed

- Stop using string/architecture heuristics (e.g. `*_text`, `gemma3_text`, `Gemma3ForCausalLM`) to reject or force Gemma 3 down a vision-only path.
- In AutoVisionLoader, parse `Gemma3Config` and distinguish:
  - `Gemma3Config::Text`
  - `Gemma3Config::WithVision`
- Keep delegation inside the existing Gemma 3 loader family (no new loader types).
- Make `Gemma3Loader::modalities` config-aware:
  - `Text` -> `[Text]`
  - `WithVision` -> `[Text, Vision]`

## Tests

Add focused regression tests covering:

- `gemma3_text` config is accepted and reports text-only modalities
- multimodal Gemma 3 config reports text+vision modalities

## Scope

- Refs #1799: fixes the misclassification / misrouting aspect (text-only Gemma 3 should not be treated as requiring vision).
- Does not add Gemma 3 GGUF architecture support.
- Does not by itself guarantee full #1799 closure without end-to-end verification that `google/functiongemma-270m-it` loads successfully through the intended path.

## Follow-up

Track GGUF handling separately: current GGUF loader paths do not include Gemma 3 architecture support, so Gemma 3 GGUF enablement should land as a dedicated change.

## Notes

- CI/typos: allowlist `layrnorm` because `pre_layrnorm` is an external checkpoint key string (not prose).


EricLBuehler (changes requested):
Hi @glaziermag! This PR likely does not fix #1799, as gemma 3 has the 1b variant which is text only.

glaziermag:
Thanks for the review — I’ll adapt this accordingly.

glaziermag:
@EricLBuehler Good point about the 1b text-only variant. This PR handles it by routing based on the parsed config (`text_config` vs `vision_config` fields) rather than hardcoding model names, so the 1b text-only variant would be routed through the text loader path since it lacks `vision_config`. The separate PR #1964 adds explicit Gemma3 GGUF text-only support for that case. Would you be open to re-reviewing with that context? Happy to add explicit handling for the 1b variant if you'd prefer it in this PR.

glaziermag:
Closing this PR per @EricLBuehler's review feedback from Feb 27 — this approach doesn't address #1799 since Gemma 3 has a text-only 1B variant that needs a different routing strategy. The gemma3 text-only support has been reworked in #1964 instead. Thank you for the review!

PR #1942, opened 2026-02-25
ci: disable docs deployment on forks

haricot:
**Description**
This PR fixes an issue where the `docs` GitHub Actions workflow fails on forks because GitHub Pages is not configured or enabled by default. 
By adding the `if: github.repository == 'EricLBuehler/mistral.rs'` condition to the `deploy` job, we ensure that the documentation is only built and deployed on the main repository, preventing unnecessary CI failures for contributors.
**Fixes**
Resolves the `Get Pages site failed` error occurring in forks during the `actions/configure-pages` step.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,6 +17,7 @@ concurrency:
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


EricLBuehler (approved):
Thank you!

PR #1881, opened 2026-02-10
Fix swapped width/height in Qwen VL preprocessor

jess-sol:
resize_exact expects width then height, but was being passed height/width. This would get corrected, but not before degrading image quality (by first downsizing, then upsizing), causing subtle systemic spatial reasoning issues.

Affects Qwen2-VL, Qwen2.5-VL, and Qwen3-VL preprocessors.

PR #1950, opened 2026-02-27
fix(vision): correct Qwen VL multi-turn image processing and thinking model token decoding

EricLBuehler:
## Summary

- Fix multi-turn image processing for Qwen3-VL, Qwen2-VL, and Qwen2.5-VL producing broken output or crashing when images of different sizes appear across conversation turns
- Fix `<think>`/`</think>` delimiters not appearing in output for thinking models

### Vision fixes (Qwen3-VL, Qwen2-VL, Qwen2.5-VL)

- **`resize_exact` argument swap**: `resize_exact` takes `(width, height)` but was called with `(height, width)`
- **Per-image preprocessing**: all images in a batch were resized to the max height/width across the batch, distorting smaller images in multi-turn conversations. Each image is now processed at its own resolution using `Tensor::cat` instead of `Tensor::stack`, matching HF behavior
- **`min_pixels`/`max_pixels` fallback** (Qwen3-VL only): reads from `size.shortest_edge`/`size.longest_edge` when `min_pixels`/`max_pixels` keys are absent in `preprocessor_config.json`
- **`get_rope_index` grid alignment** (Qwen3-VL/MoE only): prefix-cache-trimmed `input_ids` caused vision spans to misalign with `grid_thw` entries; now always uses `input_ids_full` for MRoPE computation and narrows the resulting `position_ids`
- **Prefix cache trimming**: narrows `pixel_values` by patch count from `grid_thw` instead of image count

### Sampling fix

- Enable `include_special` during token decoding when think tag mode is active so `<think>`/`</think>` delimiters are visible in output

### DIFF
--- a/mistralrs-core/src/pipeline/sampling.rs
+++ b/mistralrs-core/src/pipeline/sampling.rs
@@ -38,9 +38,10 @@ pub(crate) async fn finish_or_add_toks_to_seq(
     let tok_env = metadata.tok_env().ok_or(candle_core::Error::Msg(
         "`finish_or_add_toks_to_seq` requires the pipeline to have a token trie".to_string(),
     ))?;
-    // Include special tokens only when tool calling is active, so tool parsers
-    // can see delimiters like <tool_call>, [TOOL_CALLS], <|python_tag|>.
-    let include_special = seq.tools.is_some();
+    // Include special tokens when tool calling is active (so tool parsers can see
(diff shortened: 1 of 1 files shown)


EricLBuehler:
@jess-sol this closes https://github.com/EricLBuehler/mistral.rs/pull/1881 as it fixes the issues you described.

PR #1957, opened 2026-03-01
fix(mistralrs-mcp): fix WebSocketTransport dropping concurrent responses

### LINKED ISSUE #1947: mistralrs-mcp: WebSocketTransport can drop other in-flight responses under concurrency
`WebSocketTransport::send_request` can misroute responses under concurrent use when multiple callers share one `WebSocketTransport` instance.

## Where

- `mistralrs-mcp/src/transport.rs`
- `WebSocketTransport::send_request`

## What happens

Each `send_request()` call:

1. Generates an ID
2. Sends JSON-RPC over the shared WebSocket write half
3. Enters a read loop on the shared read half
4. If it receives a response with a different `id`, it keeps looping instead of routing/buffering that message for the matching waiter

There is no shared dispatch table or unmatched-message buffer for other in-flight requests.

## Why this is a bug

With two concurrent callers on one shared transport:

- Task A sends request `id=1`
- Task B sends request `id=2`
- Server replies out of order: `id=2` then `id=1`

If Task A reads first, it can consume response `id=2` while waiting for `id=1`.
Task B can then wait indefinitely (or until timeout/closure) because its response was already consumed.

## Impact (scope)

This manifests specifically when multiple callers issue concurrent requests through the same `WebSocketTransport` instance. In that scope, request/response correlation can break and callers can hang.

karthikreddy-02:
- Replace shared read mutex with a background dispatch task
- Each send_request() registers a oneshot channel in a HashMap keyed by request ID before writing, then awaits its private receiver
- Background reader routes responses to the correct waiter by ID
- Fixes ghost responses and indefinite hangs under concurrent use
- Also fix WebSocket handshake header omission when custom headers used
- Add regression test: mock server replies out of order, both callers succeed Fixes #1947

### DIFF
--- a/mistralrs-mcp/src/transport.rs
+++ b/mistralrs-mcp/src/transport.rs
@@ -1,12 +1,17 @@
 use anyhow::Result;
-use futures_util::stream::{SplitSink, SplitStream};
+use futures_util::stream::SplitSink;
 use futures_util::{SinkExt, StreamExt};
 use http::{Request, Uri};
 use serde_json::Value;
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::Duration;
 use tokio::net::TcpStream;
+use tokio::sync::Mutex;
 use tokio_tungstenite::{connect_async, tungstenite::Message, MaybeTlsStream, WebSocketStream};
+
+type DispatchTable =
+    Arc<Mutex<HashMap<u64, tokio::sync::oneshot::Sender<Result<Value, anyhow::Error>>>>>;
 
 /// Transport layer for MCP communication
 #[async_trait::async_trait]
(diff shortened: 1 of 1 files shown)


karthikreddy-02:
@mitsuhiko can you look into this? Thank you!

EricLBuehler (changes requested):
Hi @karthikreddy-02! Thanks for the PR.

Can you please do a bit of cleanup in regards to comments/docs/code formatting added?

karthikreddy-02:
Will do!

karthikreddy-02:
Thanks for the review @EricLBuehler! I've rebased onto the latest master (conflict was with the JSON-RPC helper refactor in #2019 — now resolved so `send_request` reuses `build_jsonrpc_request`/`extract_jsonrpc_result`) and done a cleanup pass on the code I added:

- Removed the verbose/multi-paragraph doc blocks, the ASCII diagram, and the narrating inline comments; kept comments to terse one-liners
- Fixed the indentation/formatting in `WebSocketTransport::new` and stripped trailing whitespace
- Made the background reader fail all pending waiters when the connection closes, so callers get an error instead of hanging (matches the doc's error-propagation claim)

`cargo fmt --check`, `cargo clippy`, and the regression test all pass. Ready for another look.

PR #1967, opened 2026-03-06
fix(sdk): propagate enable_thinking from TextMessages and add set_sampler_repetition_penalty

jm-observer:
## Summary

Two small but impactful bug fixes in the Rust SDK:

- **`enable_thinking` was silently dropped** when converting `TextMessages`
  into `RequestBuilder`. The `From<TextMessages> for RequestBuilder` impl
  always set `enable_thinking: None`, ignoring whatever the caller had
  configured on the `TextMessages` instance. This meant calling
  `.enable_thinking(true/false)` on a `TextMessages` had no effect.

- **`set_sampler_frequency_penalty` does not exist** on `RequestBuilder`,
  but the websocket UI handler was calling it. This would fail to compile
  unless the method happened to be defined elsewhere. The correct method
  for repetition penalty is `set_sampler_repetition_penalty`, which is now
  added to `RequestBuilder`.

## Changes

- `mistralrs/src/messages.rs`: Forward `value.enable_thinking` in the
  `From<TextMessages>` conversion instead of hardcoding `None`.
- `mistralrs/src/messages.rs`: Add `RequestBuilder::set_sampler_repetition_penalty`
  (sets `sampling_params.repetition_penalty`).
- `mistralrs-cli/src/ui/handlers/websocket.rs`: Call
  `set_sampler_repetition_penalty` instead of the non-existent
  `set_sampler_frequency_penalty`.

### DIFF
--- a/mistralrs-cli/src/ui/handlers/websocket.rs
+++ b/mistralrs-cli/src/ui/handlers/websocket.rs
@@ -73,7 +73,7 @@ fn apply_gen_params(
         .set_sampler_topp(top_p)
         .set_sampler_topk(top_k)
         .set_sampler_max_len(max_tokens)
-        .set_sampler_frequency_penalty(rep_penalty);
+        .set_sampler_repetition_penalty(rep_penalty);
 
     builder
 }
--- a/mistralrs-cli/src/ui/handlers/websocket.rs
+++ b/mistralrs-cli/src/ui/handlers/websocket.rs
@@ -73,6 +73,7 @@ fn apply_gen_params(
         .set_sampler_topp(top_p)
         .set_sampler_topk(top_k)
         .set_sampler_max_len(max_tokens)
+        .set_sampler_frequency_penalty(rep_penalty)
         .set_sampler_repetition_penalty(rep_penalty);
 
     builder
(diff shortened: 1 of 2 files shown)


mistralrs-cli/src/ui/handlers/websocket.rs
```diff
@@ -73,7 +73,7 @@ fn apply_gen_params(
         .set_sampler_topp(top_p)
         .set_sampler_topk(top_k)
         .set_sampler_max_len(max_tokens)
-        .set_sampler_frequency_penalty(rep_penalty);
```
EricLBuehler: Why was this removed in favor of `set_sampler_repetition_penalty`? I think it should be both.

EricLBuehler (changes requested):
Hi @jm-observer! Looks good, just 1 small comment?

mistralrs-cli/src/ui/handlers/websocket.rs
jm-observer: 😅 I’m not sure—this was modified by AI.

EricLBuehler: No worries! Lets just make sure its both `. set_sampler_repetition_penalty` and `. set_sampler_frequency_penalty`.

PR #1977, opened 2026-03-11
Fix UQFF loading for MoE models in Qwen2Loader

### LINKED ISSUE #1963: UQFF loading fails for MoE models: expert tensors not dummied during from_uqff overlay
## Summary

   UQFF loading does not work for Mixture-of-Experts (MoE) models. When loading a UQFF-cached MoE model (e.g., Qwen3-30B-A3B), there are two failure modes depending on whether the HF model_id or a local
   directory is used:

   1. **HF model_id**: The bf16 safetensors are loaded alongside the UQFF overlay, causing CUDA OOM because expert weights (~15GB in bf16) are not dummied
   2. **Local directory** (with only `residual.safetensors` + UQFF shards): Fails with `cannot find tensor model.layers.0.mlp.experts.0.gate_proj.weight`

   ## Reproduction

   ```bash
   # Step 1: ISQ + write UQFF (works)
   # Using TextModelBuilder with .with_isq(Q4K).write_uqff(path)
   # Successfully writes model-0.uqff, model-1.uqff, residual.safetensors, config.json, tokenizer.json

   # Step 2: Load from UQFF (fails)
   # Using UqffTextModelBuilder::new(model_id, vec![shard_paths]).into_inner().build()
   ```

   **Hardware:** RTX 3090 (24GB VRAM), 31GB RAM
   **Model:** `Qwen/Qwen3-30B-A3B` (MoE, 128 experts, 48 layers)
   **mistralrs version:** 0.7 (latest crates.io release)

   ## Root Cause Analysis

   ### Failure mode 1: HF model_id → CUDA OOM

   When `from_uqff` is set, the `normal_model_loader!` macro creates dummy regexes via `isq_layer_regexes()` for ISQ-able layers. However, for MoE models where `is_moqe` is false (standard uniform
   quantization, not mixed-quantization MoE), the regexes from `isq_layer_regexes()` may not cover MoE expert weight patterns (e.g., `model.layers.*.mlp.experts.*.gate_proj.weight`).

   As a result, expert weights are loaded from the bf16 safetensors instead of being dummied. Since expert weights constitute the bulk of the model (~15GB for Qwen3-30B-A3B), this exceeds VRAM.

   ### Failure mode 2: Local directory → missing tensor

   When using a local directory containing only `residual.safetensors` (norms, embeddings, gate routing) + UQFF shards as `model_id`, the `get_model_paths` function correctly falls back to
   `residual.safetensors`. However, `FusedExperts::new` / `load_fused_per_expert` in `mistralrs-core/src/moe/experts.rs` calls `get_with_hints()` for individual expert weights and does **not** handle the
   `CannotFindTensor` error gracefully — it propagates as a hard error instead of creating a dummy/zero tensor (as `ColumnParallelLayer::new` and `RowParallelLayer::new` do for dense layers).

   ## Expected Behavior

   UQFF loading should work for MoE models the same way it works for dense models:
   1. Expert weights should be dummied during initial model construction (not loaded from bf16)
   2. `load_from_artifacts()` replaces the dummy expert layers with the quantized UQFF weights

   ## Suggested Fix

   Either:
   - **Option A:** Make `isq_layer_regexes()` include MoE expert weight patterns when `from_uqff` is set, so expert weights are dummied regardless of `is_moqe`
   - **Option B:** In `FusedExperts::new` / `load_fused_per_expert`, handle `CannotFindTensor` by creating dummy layers when the dummy regex mechanism is active (similar to how dense layer constructors handle
    it)
   - **Option C:** In `get_model_paths`, when `loading_from_uqff` is true and `residual.safetensors` exists in the UQFF path, prefer it over the full model safetensors

   ## Notes

   - There are zero MoE models in the [official UQFF collection](https://huggingface.co/collections/EricB/uqff-670e4a49d56ecdd3f7f0fd4c), suggesting this is an untested code path
   - Dense models (e.g., Qwen3-14B) work perfectly with UQFF loading via both the Rust SDK and local directory approach
   - The UQFF files themselves are valid — ISQ + `write_uqff` produces correct output with packed expert tensors (337 quantized tensors across 2 shards)

glaziermag:
This PR fixes #1963 where MoE models loaded via `--from-uqff` or a local directory containing UQFF artifacts fail to load. The issue was that the `Qwen2Loader` only included dense MLP weights in `isq_layer_regexes`, causing missing MoE expert weights to be treated as fp16 models during UQFF loading. This hits the VRAM limit or results in `CannotFindTensor`. Testing done dynamically via cargo CLI builds during resolving.


emanueleDiVizio:
Great work, thanks for looking into this @glaziermag !

PR #1978, opened 2026-03-11
feat(server): add configurable concurrency limit to router builder

Mattbusel:
## Motivation

LLM inference is uniquely memory-intensive — a single request can hold gigabytes of KV-cache in VRAM/RAM for the duration of the stream.  When many requests arrive simultaneously the server can:

- **OOM-kill** the process (GPU or CPU), causing a complete outage
- Thrash between contexts, degrading throughput for all users
- Trigger CUDA out-of-memory errors with no clean recovery path

Today `mistralrs-server` exposes no mechanism to bound the number of in-flight requests.

## Change

Add `with_max_concurrency(n: usize)` to `MistralRsServerRouterBuilder`.  When set, the axum router is wrapped with Tower's [`ConcurrencyLimitLayer`](https://docs.rs/tower/latest/tower/limit/struct.ConcurrencyLimitLayer.html) which **queues** requests beyond the limit rather than rejecting them — providing back-pressure without dropping traffic.

```rust
let router = MistralRsServerRouterBuilder::new()
    .with_mistralrs(state)
    .with_max_concurrency(8)   // cap at 8 simultaneous in-flight requests
    .build()
    .await?;
```

The feature is **opt-in** — omitting `with_max_concurrency` leaves the existing behaviour unchanged.

## Implementation

- Adds `tower = { version = "0.5", features = ["limit"] }` to the workspace and `mistralrs-server-core` Cargo manifests (`tower-http` was already present).
- Threads `max_concurrency: Option<usize>` through the builder → `init_router`.
- Applies `ConcurrencyLimitLayer` at the outermost router layer (after CORS and body-limit) so it covers all endpoints uniformly.

## Testing

The change is additive; existing tests continue to pass.  Manual smoke test with `with_max_concurrency(2)` confirms that a third concurrent request queues until one of the first two completes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -165,6 +165,7 @@ ctrlc = "3.5.1"
 directories = "6.0.0"
 # Need to keep rustyline at 15.0.0 as 16.0.0 has a breaking change that conficts with `ctrlc`'s handling
 rustyline = { version = "15.0.0", default-features = false, features = ["with-file-history"] }
+tower = { version = "0.5", features = ["limit"] }
 tower-http = "0.6.8"
 utoipa-swagger-ui = { version = "9.0.2", features = ["axum", "vendored"] }
 futures-util = "0.3.31"
--- a/mistralrs-server-core/Cargo.toml
+++ b/mistralrs-server-core/Cargo.toml
@@ -30,6 +30,7 @@ reqwest.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 tokio.workspace = true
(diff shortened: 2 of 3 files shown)


EricLBuehler:
Hi @Mattbusel! Thanks for the PR. We already have a method for this in `MistralRsForServerBuilder::with_max_seqs` which is designed to do exactly this. `MistralRsServerRouterBuilder` is uniquely for the server.

Please feel free to reopen if you change this to be a setting that's more unique.

Issue #545, opened 2024-07-03
[FEATURE REQUEST] Dynamic Model Loading and Unloading for Efficient VRAM Management


Qualzz:
Hello! I had a thought. To minimize constant load for tasks that occur infrequently, is there a way to keep the Docker container running with the HTTP server, but only load the model when a query is made and then unload it from memory after a user-defined period? The default could be 5 minutes or indefinitely.

For example, I may have a task that requires a call to the MistralRS HTTP server and then triggers a Stable Diffusion generation. In that case, I'll need the VRAM to be available.

Another example: I have a Discord bot that performs various tasks, like regular chat with Gemma2 9B (because it's fluent in French). However, the bot may also need to look up information in a large document using RAG. For that, It switch to another model with a larger context window (but less fluent in French) to answer the query and then switch back to Gemma 2 again to rephrase it with its own persona and bot context. I do a lot of model hot-swapping in a constrained VRAM environment.

One amazing feature would be the ability to set a keep_alive (in seconds) parameter, similar to setting the temperature in a query. Mistral would use that variable to free up the memory after the specified amount of time.

For large model I would set it to 0, meaning immediate unload once the query is done, while for smaller models, I would use a certain period of time to avoid reloading it during a chat session.

I'm not sure if this is a stupid idea or if I'm the only one who needs something like this.

EricLBuehler:
Hi @Qualzz! Thank you for suggesting this. 

This idea falls under the topic of model hot-swapping - which we already have in some form with our Dynamic LoRA Adapter Activation feature. However, unloading and then loading a model is much more powerful - and best of all - is probably relatively easy to implement in the APIs. This can be done shortly, I will keep you updated here.

I think that changing the loaded model would be better implemented for your use case as killing the mistral.rs process and the restarting it with a different model. The reason is that otherwise, there would be a *lot* of code bloat for what is otherwise a rare situation. Is this not an option?

Qualzz:
@EricLBuehler

How would that scenario work in practice?

I have a backend that, upon receiving a request, makes a call to MistralRS. If I have two users making requests within a few seconds of each other, with the current implementation, I can keep the Mistral process running, and it will work, but the model will remain loaded in VRAM indefinitely.

If I terminate the process after each request, it becomes complicated as it might interfere with the second query. Additionally, I don't want to load two instances of Mistral simultaneously.

EricLBuehler:
[...]

PR #2030, opened 2026-03-25
Address possible memory leak during model sleep/unload

glaziermag:
Tentatively addresses #545.

## Problem

After calling `/v1/sleep` (model unload), VRAM was not fully returned to the OS. The likely cause: tensors held in `RebootState` were dropped on the HTTP handler thread, which does not have the CUDA OS context bound to it. `cuMemFreeAsync` requires the context to be current on the calling thread; without this binding, deallocations silently fail at the driver level and the memory pool retains the allocations.

## Fix

Three steps, in order:

**1.** Join the engine worker thread before dropping `RebootState`, ensuring the async engine has fully exited and released its own tensor references before the HTTP thread attempts the drop.

**2.** Bind the CUDA context to the HTTP thread (`dev.cuda_stream().context().bind_to_thread()`) before `drop(reboot_state)`. This ensures `cuMemFreeAsync` has a valid context to execute against.

**3.** After the drop, call `device.synchronize()` to flush any in-flight async frees, then trim the CUDA default memory pool to its currently-used watermark to release the idle reserve back to the OS.

## Pool trim approach

`cuMemPoolTrimTo(pool, 0)` (trim everything) is avoided because in a multi-model server the default memory pool is shared. Trimming to zero would evict blocks held idle by other still-active models, forcing expensive OS reallocations. Instead the trim queries `CU_MEMPOOL_ATTR_USED_MEM_CURRENT` to get the active-use watermark and trims only to that value. If the query fails (old driver), the trim is skipped entirely.

## Clean branch

The original branch (`fix-memory-leak-unload-v4`) was based on the fork's master rather than `origin/master`, so its diff includes ~200 unrelated upstream files. A clean branch containing only this commit has been pushed as `fix-memory-leak-unload-clean` on the same fork.

## Files changed

- `mistralrs-core/src/lib.rs`
- `mistralrs-server-core/src/handlers.rs`
- `mistralrs-server-core/src/mistralrs_server_router_builder.rs`


glaziermag:
**Update (2026-04-15):** The original branch (`fix-memory-leak-unload-v4`) is rebased on the fork's master, not upstream, so the diff shows ~200 unrelated files. A clean isolated branch (`fix-memory-leak-unload-clean`) has been pushed that cherry-picks only the single commit onto current `origin/master`.

Additionally, `cuMemPoolTrimTo(pool, 0)` has been replaced with a safer alternative:

**Old (aggressive):**
```rust
sys::cuMemPoolTrimTo(pool, 0);  // releases ALL cached capacity — also evicts other models' idle blocks
```

**New (targeted):**
```rust
// Query currently-used bytes first
let attr_ok = sys::cuMemPoolGetAttribute(pool, CU_MEMPOOL_ATTR_USED_MEM_CURRENT, &mut used_bytes ...);
if attr_ok == CUDA_SUCCESS {
    sys::cuMemPoolTrimTo(pool, used_bytes as usize);  // releases idle reserve only
}
// If attribute query fails, skip trim — OS reclaims via pool refcount eviction
```

Trimming to 0 in a multi-model server evicts blocks that other still-active models may be about to reuse, forcing expensive re-allocations from the OS and potential OOM. Trimming to `USED_MEM_CURRENT` releases only the idle over-reserve from the unloaded model without touching live allocations.

glaziermag:
Closing in favor of focused, single-responsibility PRs:

- **#2113** — CUDA context binding + memory pool trim during unload (the core fix)

The breaking API change (`model_id: String` → `Option<String>`) and vLLM/SGLang route aliases (`/sleep`, `/wake_up`, etc.) should be submitted as separate PRs if still desired.

PR #2034, opened 2026-03-28
fix(scheduler): defer incompatible paged-attention buckets

### LINKED ISSUE #2024: paged_attention: O(N²) memory allocator thrashing and continuous batching failure
# paged_attention: O(N²) memory allocator thrashing and continuous batching failure via `bucket_and_preempt_sequences`

## Describe the bug
The `PagedAttentionScheduler` introduces an $O(N^2)$ allocation pattern and disrupts continuous batching due to the `bucket_and_preempt_sequences` logic in `mistralrs-core/src/paged_attention/scheduler.rs`.

When the scheduler encounters sequences of different lengths, it buckets them, selects only one bucket (typically resulting in batch size 1), and preempts all other sequences. This preemption alters their `SequenceState`, calls `kv_mgr.free(seq_id)` on their cached blocks, and returns them to the `waiting` queue.

Because `bucket_and_preempt_sequences` is invoked during both the `Prompt` and `Completion` scheduling phases, sequence length divergence disrupts PagedAttention batching and can place longer sequences into a continuous re-prefill loop.

### The Execution Path & Re-Prefill Cycle
1. Sequence A is currently generating (e.g., total context length 120, state = `RunningCompletion`).
2. Sequence B is submitted as a new prompt (e.g., length 50, state = `RunningPrompt`).
3. On the next tick, both are evaluated for scheduling. They are placed in different length buckets.
4. The scheduler selects Sequence B (shortest length) and calls `_preempt` on Sequence A. Sequence A's KV Cache blocks are freed, and it is pushed to the `waiting` queue.
5. On the subsequent tick, the scheduler retrieves Sequence A from `waiting`, allocates new blocks, and sets its state to `SequenceState::RunningPrompt` (Line 287).
6. **The Result:** The Engine processes Sequence A as a prefill prompt, recomputing the entire 120-token prior context.
7. Once recomputed, Sequence A returns to the completion phase. If its length again mismatches Sequence B's length, it is preempted, its KV cache is freed, and it is re-prefilled again.

Under concurrent requests of varying lengths, active longer generations can be repeatedly preempted and forced to perform prompt re-prefills without generating new completion tokens.

## Reproduction
1. Start `mistralrs-server` with Paged Attention enabled.
2. Submit a longer generation request (e.g., prompt length 100).
3. Concurrently submit multiple requests with varying smaller prompt lengths (e.g., lengths 50, 51, 52).
4. Observe the GPU usage increase and throughput decrease as the scheduler performs frequent KV Cache allocations/frees and operates at `bs=1`, causing the longer prompt to recompute repeatedly.

glaziermag:
## Summary

Treats PagedAttention length, image-modality, and token-offset incompatibility as a scheduling constraint rather than memory pressure.

This replaces the previous narrow #2034 implementation because both issues modify the same bucketing helper and completion drain loop. The one-commit branch is based directly on current `master` and retains #2034's oldest-bucket FCFS behavior while fixing the repeated-preemption path from #2024.

## Fix

- Partition incompatible sequences around the bucket containing the oldest `(timestamp, sequence_id)` request.
- Use that same deterministic `(timestamp, sequence_id)` ordering for memory-pressure victim selection.
- Keep deferred prompts allocated and in `RunningPrompt`, then emit those prompt buckets before entering completion scheduling.
- Transition only prompts that the paged engine actually executed to `RunningCompletion` (or their one-shot terminal state); deferred prompts remain prompts until emitted.
- Keep deferred completions allocated and in `RunningCompletion` without reserving their next decode slot until selected.
- Remove the completion modality reinsert-while-draining loop.
- Leave `_preempt` available only on actual KV allocation failure; incompatibility alone no longer calls `_preempt`, frees blocks, resets state, or moves a sequence back to `waiting`.
- Preserve the existing current-`master` FCFS ordering correction from #2250.

## Regression coverage

Current head: `72710c75154b6a9d090108ed807abf69a9388263`.
Exact tree: `8c6a046e597d0c9391aa778362d25bbc54345073`.

Focused scheduler tests cover:

- oldest-bucket selection and deterministic sequence-ID tie breaking
- equal-timestamp KV pressure evicting the higher-ID request even from adverse queue order
- prompt allocation and `RunningPrompt` preservation across incompatible lengths
- engine-owned prompt-to-completion transition for only the emitted bucket, followed by decode rather than prompt replay
- completion allocation and `RunningCompletion` preservation across incompatible lengths
- deferred completion progress after the older bucket finishes
- image-modality deferral without reinsertion or free
- true KV memory pressure still preempting a lower-priority request
- idle scheduling returning an empty batch

## Validation

Rust 1.97.0, offline dependency cache:

```bash
cargo test -p mistralrs-core paged_attention::scheduler::tests --lib --offline
# 8 passed; 0 failed

cargo test -p mistralrs-core paged_attention --lib --offline
# 64 passed; 0 failed

cargo fmt --all -- --check
git diff --check
```

GitHub CI run `29202853420` completed green for Rustfmt, Doc Links, Typos, Docs, Ubuntu/macOS/Windows checks, Metal check, and Ubuntu/macOS/Windows test suites. The CUDA workflow was skipped by its normal trigger conditions.

`cargo clippy -p mistralrs-core --lib --tests -- -D warnings` reports no diagnostic in either changed file. Both local Clippy and the GitHub Clippy job stop on the same four pre-existing current-`master` warnings in `multimodal_loaders.rs`, `block_pool.rs`, and `lib.rs`.

## Scope and dependency

- Changed files: `mistralrs-core/src/paged_attention/scheduler.rs` and the paged prompt state handoff in `mistralrs-core/src/engine/mod.rs`.
- No dependency on another open PR; based directly on `b7956205fae93f93feed2b87237e752ee8083b56` (`master`).
- Supersedes the earlier #2034 head rather than opening a conflicting second scheduler PR.

Fixes #2024

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -5,6 +5,7 @@ extend-exclude = [
     "examples/server/phi3_duckduckgo_mistral.rs.ipynb",
     "mistralrs-web-chat/static/",
     "mistralrs-cli/static/",
+    "mistralrs-quant/kernels/mmq_gguf/",
     "CLAUDE.md",
 ]
 ignore-hidden = false
@@ -12,6 +13,7 @@ ignore-hidden = false
 [default]
 extend-ignore-re = [
     "cudaDevAttrMaxSharedMemoryPerBlockOptin",
+    "CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN",
     '"tese"',
     "setp\\.ne\\.b32",
     # metal_kernels
--- a/.typos.toml
+++ b/.typos.toml
@@ -5,7 +5,6 @@ extend-exclude = [
     "examples/server/phi3_duckduckgo_mistral.rs.ipynb",
     "mistralrs-web-chat/static/",
     "mistralrs-cli/static/",
-    "mistralrs-quant/kernels/mmq_gguf/",
     "CLAUDE.md",
 ]
 ignore-hidden = false
@@ -13,7 +12,6 @@ ignore-hidden = false
 [default]
 extend-ignore-re = [
     "cudaDevAttrMaxSharedMemoryPerBlockOptin",
-    "CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN",
     '"tese"',
     "setp\\.ne\\.b32",
     # metal_kernels
--- a/mistralrs-cli/src/commands/bench.rs
+++ b/mistralrs-cli/src/commands/bench.rs
@@ -99,7 +99,7 @@ pub async fn run_bench(
     if warmup > 0 {
         info!("Running {} warmup iteration(s)...", warmup);
         for _ in 0..warmup {
-            let _ = run_single_bench(&mistralrs, 32, 16).await?;
+            run_single_bench(&mistralrs, 32, 16).await?;
         }
         info!("Warmup complete.");
 
--- a/mistralrs-cli/src/commands/bench.rs
+++ b/mistralrs-cli/src/commands/bench.rs
@@ -99,7 +99,7 @@ pub async fn run_bench(
     if warmup > 0 {
         info!("Running {} warmup iteration(s)...", warmup);
         for _ in 0..warmup {
-            run_single_bench(&mistralrs, 32, 16).await?;
+            let _ = run_single_bench(&mistralrs, 32, 16).await?;
         }
         info!("Warmup complete.");
 
--- a/mistralrs-cli/src/ui/handlers/api.rs
+++ b/mistralrs-cli/src/ui/handlers/api.rs
@@ -21,7 +21,7 @@ use crate::ui::types::{
     AppState, ChatFile, DeleteChatRequest, LoadChatRequest, NewChatRequest, RenameChatRequest,
     SelectRequest,
 };
-use crate::ui::utils::{get_cache_dir, is_chat_id_safe};
+use crate::ui::utils::get_cache_dir;
 
 fn validate_image_upload(
     filename: Option<&str>,
@@ -373,9 +373,7 @@ pub async fn delete_chat(
     Extension(app): Extension<Arc<AppState>>,
     Json(req): Json<DeleteChatRequest>,
 ) -> impl IntoResponse {
-    if !is_chat_id_safe(&req.id) {
-        return (StatusCode::BAD_REQUEST, "invalid chat id").into_response();
-    }
+
     let path = format!("{}/{}.json", app.chats_dir, req.id);
     match fs::remove_file(&path).await {
         Ok(_) => (StatusCode::OK, "Deleted").into_response(),
@@ -387,9 +385,7 @@ pub async fn load_chat(
     Extension(app): Extension<Arc<AppState>>,
     Json(req): Json<LoadChatRequest>,
 ) -> impl IntoResponse {
-    if !is_chat_id_safe(&req.id) {
-        return (StatusCode::BAD_REQUEST, "invalid chat id").into_response();
-    }
+
     let path = format!("{}/{}.json", app.chats_dir, req.id);
     if let Ok(bytes) = fs::read(&path).await {
         if let Ok(chat) = serde_json::from_slice::<ChatFile>(&bytes) {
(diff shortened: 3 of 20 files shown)


glaziermag:
**Update (2026-04-15):** Removed an undocumented hardcoded batch cap that was introduced in this PR.

The original commit added:
```rust
if (batched_prompt_tokens + num_new_tokens > 16384 || batched_sequences >= 10) && batched_sequences > 0 {
    break;
}
```

This cap was neither mentioned in the PR description nor the issue, and has significant problems:
- **Not configurable** — silently throttles models with >16384-token context windows (e.g. Llama 32k, Qwen 128k) with no warning
- **Hardware-specific** — the 16384 limit is CUDA/cuBLAS-specific and incorrect for Metal or CPU backends
- **Redundant** — `max_num_seqs` and the KV allocator's `allocate_slots` failure path already enforce capacity correctly

The FCFS `.reverse()` removal and the preemption loop improvements remain untouched. Only the batch cap and its tracking variables were removed.

glaziermag:
**Second update (2026-04-15):** Removed dead variable warning.

The `num_new_tokens` binding (and its `let _ = num_new_tokens` suppressor) was left over from the batch cap removal and was never actually used by the prefix cache lookup — that path uses `num_tokens`. Both lines have been removed. This also corrects the misleading `// used by prefix cache lookup below` comment.

glaziermag:
**Housekeeping note:** This branch currently bundles the CI fix from #2115 (`.typos.toml`, `openapi_doc.rs`, `distributed/layers.rs`). Once #2115 is merged, this branch will need a rebase onto updated `master` to drop the duplicate CI fix commit and resolve the resulting conflicts.

glaziermag:
Agent 6 follow-up on existing A100 validation: this remains valid as a targeted FCFS scheduler fix. Classification: `TARGETED`; feasibility: `FEASIBLE_NOW`. The A100 evidence validates the injected paged-attention bucket priority invariant, but does not prove a broader runtime starvation workload. Safe wording should stay limited to preserving FCFS priority across paged-attention buckets. Recommendation: keep open/draft for review.

glaziermag:
Marked ready for review. Validation evidence and narrowed claim wording are already attached in the PR discussion/body. This PR is ready under the scoped claim described in the PR.

Ready for maintainer review under the narrowed claim in the PR body. This is a targeted/invariant fix and should not be read as full closure of the broader linked issue unless the PR body explicitly says so.

glaziermag:
Scope/head update: `72710c75154b6a9d090108ed807abf69a9388263` supersedes the earlier narrow implementations discussed above. Because #2024 and #2034 modify the same scheduler helper/drain loop, the current one-commit head now defers incompatible length, image-modality, and token-offset buckets without treating incompatibility as memory pressure.

Adversarial review caught and corrected two issues in intermediate (now superseded) heads: (1) emitted prompts could have remained `RunningPrompt` and replayed, so the current paged engine transitions only prompts that actually executed while deferred prompts retain prompt state; and (2) equal-timestamp bucket selection and KV-pressure ordering could disagree, so both now use `(timestamp, sequence_id)`, with an adverse-order pressure regression. The updated PR body records the exact tree, 8 focused allocation/state tests, 64-test paged-attention slice, and current-master-only Clippy blockers. Older comments limiting this PR to the previous FCFS-only scope refer to superseded heads.

glaziermag:
Closing: the scenario #2024 reproduces (text-model serving with mixed prompt lengths) is fixed on master by #2354's capability-driven ragged batching — text pipelines now opt out of uniform completion batches entirely (`requires_uniform_completion_batch() -> false`, with `effective_len = 0` in `bucket_and_preempt_sequences`), so the bucketing-preemption loop this PR removed no longer fires there, and `completion_turn_due` plus oldest-first bucket selection cover the starvation half. Every hunk here conflicts with that rewrite, so a rebase would amount to a ground-up rewrite.

The residual case — pipelines that still require uniform completion/media batches (Voxtral, block-diffusion vision) are preempted rather than deferred, including the drain-loop `_preempt` of modality-mismatched running sequences (`scheduler.rs:531`), which is more aggressive than the pre-#2354 requeue — is a narrower, different change that I'd rather propose fresh against the new `enforce_completion_compatibility`/drain-loop structure.

Issue #2026, opened 2026-03-25
build.rs skips CUDA kernel compilation on Windows, causing link errors on master


s520145:
## Minimum reproducible example

```powershell
$env:CARGO_TARGET_DIR = "C:\tmp\cb"
cargo install --git https://github.com/EricLBuehler/mistral.rs --features "cuda" mistralrs-cli --force
```

Also fails with `--features "cuda flash-attn"`.

## Error

```
error[E0425]: cannot find function `flash_attn_sinks` in crate `mistralrs_paged_attn`
  --> mistralrs-core\src\attention\backends\sinks.rs:65:38
   |
65 |         return mistralrs_paged_attn::flash_attn_sinks(
   |                                      ^^^^^^^^^^^^^^^^ not found in `mistralrs_paged_attn`

error[E0425]: cannot find function `flash_attn_sinks_varlen` in crate `mistralrs_paged_attn`
   --> mistralrs-core\src\attention\backends\sinks.rs:121:38
    |
121 |         return mistralrs_paged_attn::flash_attn_sinks_varlen(
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^ not found in `mistralrs_paged_attn`

error: could not compile `mistralrs-core` (lib) due to 2 previous errors
```

### Root cause

In [`mistralrs-paged-attn/build.rs`](https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-paged-attn/build.rs), CUDA kernel compilation is gated behind `target_family = "unix"`:

```rust
#[cfg(all(feature = "cuda", target_family = "unix"))]
fn main() -> Result<()> {
    // ... compiles flash_attn_sinks.cu and other CUDA kernels
}
```

On Windows, the non-CUDA fallback `main()` runs, which only emits `cargo::rustc-check-cfg=cfg(has_fp8)` and compiles **zero CUDA kernels** — including `flash_attn_sinks.cu`.

However, [`sinks.rs`](https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/attention/backends/sinks.rs) calls these functions under `#[cfg(feature = "cuda")]` **without** a matching `target_family` guard:

```rust
#[cfg(feature = "cuda")]
if q.device().is_cuda() {
    return mistralrs_paged_attn::flash_attn_sinks(...);  // not compiled on Windows!
}
```

v0.7.0 from crates.io compiles successfully because `sinks.rs` did not exist in that release.

## Other information

[...]

Issue #1117, opened 2025-02-01
Incorrect glibc version used in x86_64 wheels


jncraton:
I'm attempting to install the most recent version of the `mistralrs` Python package on an Ubuntu Jammy system. Unfortunately, `pip install` fails because there is no matching distribution:

```
~$ pip3 install mistralrs==0.4.0
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement mistralrs==0.4.0 (from versions: 0.1.0, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5, 0.1.7, 0.1.8, 0.1.10, 0.1.11, 0.1.12, 0.1.13, 0.1.14, 0.1.15, 0.1.16, 0.1.17, 0.1.18, 0.1.19, 0.1.20, 0.1.21, 0.1.22, 0.1.23, 0.1.24, 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.2.4, 0.2.5, 0.3.0, 0.3.1, 0.3.2)
ERROR: No matching distribution found for mistralrs==0.4.0
```

There is a built distribution for manylinux x86_64 [listed on PyPI](https://pypi.org/project/mistralrs/0.4.0/#files):

```
mistralrs-0.4.0-cp310-cp310-manylinux_2_39_x86_64.whl (16.6 MB)

Uploaded Jan 22, 2025 CPython 3.10 manylinux: glibc 2.39+ x86-64
```

The required version of gcc (2.39+) is only about a year old. My Jammy system is still on 2.35 from 2022, and that is preventing a matching built distribution from being found.

The ARM64 built distribution only requires glibc 2.17+:

```
mistralrs-0.4.0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (15.4 MB)

Uploaded Jan 22, 2025 CPython 3.10 manylinux: glibc 2.17+ ARM64
```

Linux users are often using older versions of glibc. Would it be possible to provide built distributions for Linux that require only glibc 2.17+ on x86_64 as well as ARM64?

jncraton:
I've been exploring the CI and build process to try and determine why this is happening. It appears to me that the process to build and upload the Python wheels is manual and the commands in `mistralrs-pyo3/generate_wheels.sh` are used to run it.

Here are the commands to build the AArch64 manylinux wheels:

```sh
# MAC: Aarch64 Manylinux and OSX

[...]

PR #2041, opened 2026-03-28
Automate Python Wheel Builds

jncraton:
This change automates building the Python wheels using an Actions workflow in an attempt to help diagnose, address, and prevent regressions on #1117 and similar issues. Wheels generated by a mostly successful run are available here:

https://github.com/jncraton/mistral.rs/actions/runs/23693334752

The Windows CUDA build is currently failing, but the issue appears to be the same as #2026. 

This change also adjusts the build for the base `mistralrs` Python package to use Docker without MKL to maximize x64 Linux compatibility as suggested in #1117.

I have only personally tested the x64 Linux `mistralrs` wheel, but I am hopeful that the others work correctly, as they are all built by calling `scripts/build_wheels.py`.


jncraton:
This appears to now successfully build all wheels, including the previously broken Windows CUDA wheel after rebasing against the new Gemma 4 work:

https://github.com/jncraton/mistral.rs/actions/runs/23914367331

PR #2043, opened 2026-03-31
fix(paged-attn): resolve scheduler queue loop deadlock under memory pressure

### LINKED ISSUE #1470: Caching Logic Appears to Cause Model Hangs
## Describe the bug
Seeing the same effect at #1445 on any model that does actually load in mistralrs when i try to overcome the memorybalooning of KV/PA _by limiting the prefix cache size in the GPU_ - `--prefix-cache-n 8`. Model initially responds but then stops, producing no log output, no results, but taking requests and doing nothing:

Image

```rust
2025-06-14T05:32:47.570586Z  INFO mistralrs_core: Dummy run completed in 23.081259968s.
2025-06-14T05:32:47.571469Z  INFO mistralrs_server: OpenAI-compatible server listening on http://0.0.0.0:7650.
2025-06-14T05:32:49.491634Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 0.60, Prefix cache hitrate 0.00%, 0 running, 0 waiting
2025-06-14T12:14:24.773015Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 86.60, Prefix cache hitrate 0.00%, 0 running, 0 waiting
2025-06-14T12:14:29.773098Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 120.40, Prefix cache hitrate 0.00%, 1 running, 0 waiting
2025-06-14T12:14:34.773172Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 96.00, Prefix cache hitrate 0.00%, 0 running, 0 waiting

```

Apologies for not diving into the actual codebase to help remedy what looks to me to be a memory management issue - far as i can tell from the _effects_, the caching subsystem severely over-estimates memory requirements or actually multiplies cache entries as it tries to claim multiples of loaded model size regardless of NCCL, automatic, or manual distribution - even pinning a single GPU doesn't help... so i'm guessing its a core logic concern. Disabling paged attention altogether appears to make models usable for some time but the token rate is approaching useless. Cached performance on V100s for most of the llama3 derivatives is around 20T/s, without paged attention - 5-7.

## Latest commit or version
Which commit or version you ran with.

glaziermag:
### Description
This PR addresses a deadlock issue in `PagedAttentionScheduler` where the inference engine could hang during preemption under heavy KV cache memory pressure (resolves #1470).

Previously, when a sequence was preempted in the `schedule` method via `_preempt()`, it was pushed to the front of the `self.waiting` queue. Because the main starvation loop iterated over the queue non-destructively using `self.waiting.front()`, the newly preempted sequence was immediately targeted for re-evaluation. This caused an infinite loop of preemption and failed allocations.

### Changes
* Refactored the `waiting` sequence queue iteration in `scheduler.rs` to proactively `pop_front()` the sequence before checking its allocation status.
* If the scheduling loop must break early (due to `max_num_seqs` caps or unresolvable starvation conditions requiring a later retry), the sequence is safely re-inserted via `self.waiting.push_front(seq)`.

### Testing
* Reproduced the deadlock condition under a heavy simulated workload (30 concurrent requests) on a `g2-standard-32` instance.
* Verified that the patched logic properly filters preempted sequences to the next pipeline tick, maintaining stable throughput without triggering the infinite `WARN` log cycle.

#### Blast Radius Verification
To verify the structural integrity of the scheduler state, we ran a concurrent barrage of large Qwen2.5-3B completions on a strictly bounded 4096-token KV cache environment specifically designed to exhaust the BlockPool.

**Before (Unpatched):** 
The engine immediately hit the KV cache bottleneck and permanently hung at `0.00 T/s` as it recursively thrashed the preemption loop safely catching `AllocStatus::Later`:
```text
2026-03-30T23:52:01.962646Z  WARN mistralrs_core::paged_attention::scheduler: Sequence 5 with length of 351 tokens still exceeds KV cache size even after evicting another sequence.
2026-03-30T23:52:01.962654Z  WARN mistralrs_core::paged_attention::scheduler: Sequence 5 with length of 351 tokens still exceeds KV cache size even after evicting another sequence.
... <Infinitely spammed leading to 42MB deadlocked dump> ...
```

**After (Patched):**
Under identical `aiohttp` saturation tests, the sequence correctly stepped down onto the `waiting` queue and advanced the scheduler, letting independent completions securely evaluate until block space was liberated. The load successfully finished while keeping interval processing cleanly between 1.6K ~ 2.0K T/s:
```text
2026-03-31T00:05:18.231240Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 1328.40, Prefix cache hitrate 0.00%, 4 running, 26 waiting
2026-03-31T00:05:23.231414Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 1558.40, Prefix cache hitrate 0.00%, 15 running, 15 waiting
2026-03-31T00:05:38.231924Z  INFO mistralrs_core::engine::logger: Throughput (T/s) 1848.60, Prefix cache hitrate 0.00%, 15 running, 15 waiting
... <Gracefully finishes execution> ...
```

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -5,6 +5,7 @@ extend-exclude = [
     "examples/server/phi3_duckduckgo_mistral.rs.ipynb",
     "mistralrs-web-chat/static/",
     "mistralrs-cli/static/",
+    "mistralrs-quant/kernels/mmq_gguf/",
     "CLAUDE.md",
 ]
 ignore-hidden = false
@@ -12,6 +13,7 @@ ignore-hidden = false
 [default]
 extend-ignore-re = [
     "cudaDevAttrMaxSharedMemoryPerBlockOptin",
+    "CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN",
     '"tese"',
     "setp\\.ne\\.b32",
     # metal_kernels
--- a/mistralrs-core/src/paged_attention/scheduler.rs
+++ b/mistralrs-core/src/paged_attention/scheduler.rs
@@ -181,9 +181,10 @@ impl PagedAttentionScheduler {
         let mut for_waiting_again: VecDeque<Arc<Mutex<Sequence>>> = VecDeque::new();
         while !self.waiting.is_empty() {
             let mut did_ignore = false;
-            let seq = self.waiting.front().unwrap().clone();
+            let seq = self.waiting.pop_front().unwrap();
 
             if self.running.len() >= self.config.max_num_seqs {
+                self.waiting.push_front(seq);
                 break;
             }
 
@@ -263,6 +264,7 @@ impl PagedAttentionScheduler {
                             did_ignore = true;
                         }
                     } else {
+                        self.waiting.push_front(seq);
                         break;
                     }
                 }
@@ -278,7 +280,6 @@ impl PagedAttentionScheduler {
                     kv_mgr.free(seq_id);
                     drop(kv_mgr);
                 }
-                let seq = self.waiting.pop_front().unwrap();
                 for_waiting_again.push_back(seq);
                 continue;
             }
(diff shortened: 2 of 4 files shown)


glaziermag:
Closing — the preemption deadlock fix here is subsumed by the more comprehensive scheduler rework in #2034, which fixes both the FCFS priority ordering and the preemption retry loop (including the `pop_front` change from this PR). Keeping both open would cause merge conflicts on `scheduler.rs`.

PR #2044, opened 2026-03-31
Runtime CUDA Memory Expanding to OOM

glaziermag:
Resolves #1589. Resolves #1637.

## Problem

When sequences were preempted or aborted, they were removed from the `running` queue but not from the `waiting` queue. Orphaned sequences in `waiting` continued to hold their KV block allocations, causing VRAM usage to grow monotonically over time.

This affected both the PagedAttention scheduler and the DefaultScheduler.

## Fix

**PagedAttentionScheduler (`paged_attention/scheduler.rs`):** Call `retain()` inside `free_finished_sequence_groups()` to purge any sequences in `self.waiting` that have already reached a finished state before they are scheduled.

**DefaultScheduler (`scheduler/default_scheduler.rs`):** Add a `retain()` method to the `FcfsBacker` trait and implement it for both `VecDeque`-backed and `Heap`-backed queues. Call it from `free_finished_sequence_groups()`.

## Branch isolation

This branch was rebased directly onto `origin/master` to remove a hidden dependency on commit `1d15838a` from PR #2043 (`fix-paged-attention-deadlock-1470`) that had been included as the branch base. The two fixes are logically independent and can be reviewed and merged without #2043.

## Files changed

- `mistralrs-core/src/paged_attention/scheduler.rs`
- `mistralrs-core/src/scheduler/default_scheduler.rs`

PR #2045, opened 2026-04-02
fix(core): Resolve PagedAttention VRAM leak and scheduler deadlock during OOM

glaziermag:
## Problem

Under OOM or any GPU kernel error, the paged-attention engine could enter a permanent freeze where the scheduler reported full VRAM capacity and rejected all new requests indefinitely.

The freeze had two root causes:

**1. `SequenceState::Error` not handled in `is_finished_paged_attn()`**
Sequences that errored were never recognized as finished by the paged-attention scheduler. Their KV blocks stayed allocated, and the scheduler stalled waiting for capacity that would never be returned.

**2. `handle_pipeline_forward_error!` skipped `free_finished_sequence_groups()`**
The error macro calls `continue 'lp;` to restart the engine loop. This jump bypasses the `scheduler.free_finished_sequence_groups()` call at the bottom of the loop. KV blocks held by errored sequences were not reclaimed, and the scheduler eventually saw zero free blocks with zero sequences completing.

## Fix

**`sequence.rs`:** Add `SequenceState::Error` to the `is_finished_paged_attn()` match so errored sequences are recognized as done and their blocks become reclaimable.

**`utils/mod.rs`:** Add a 7-argument form of `handle_pipeline_forward_error!` that accepts a `$scheduler` argument and calls `get_mut_arcmutex!($scheduler).free_finished_sequence_groups()` before `continue $lp`. This is used exclusively by the paged-attention callsite in `engine/mod.rs`.

The original 6-argument form is preserved for the DefaultScheduler callsites. Those callsites hold the scheduler `MutexGuard` through their entire match arm — passing it to the macro and calling `get_mut_arcmutex!` on the same `Arc` from the same thread would spin-deadlock (the macro's `try_lock()` loop would never succeed). DefaultScheduler also calls `free_finished_sequence_groups()` unconditionally at the end of the engine loop, so the omission there does not cause a leak.

## Files changed

- `mistralrs-core/src/utils/mod.rs`

### DIFF
--- a/mistralrs-core/src/engine/mod.rs
+++ b/mistralrs-core/src/engine/mod.rs
@@ -413,7 +413,8 @@ impl Engine {
                             &mut scheduled.completion,
                             self.pipeline,
                             'lp,
-                            self.prefix_cacher
+                            self.prefix_cacher,
+                            &self.scheduler
                         );
 
                         self.logger.add_tokens_processed(scheduled.completion.len());
@@ -474,7 +475,8 @@ impl Engine {
                             &mut scheduled.prompt,
                             self.pipeline,
                             'lp,
-                            self.prefix_cacher
+                            self.prefix_cacher,
+                            &self.scheduler
                         );
 
                         let total_processed_tokens: usize = scheduled
@@ -697,7 +699,8 @@ impl Engine {
                             &mut guards_mut,
                             self.pipeline,
                             'lp,
-                            self.prefix_cacher
+                            self.prefix_cacher,
+                            &self.scheduler
                         );
 
                         let total_processed_tokens: usize = guards_mut
--- a/mistralrs-core/src/sequence.rs
+++ b/mistralrs-core/src/sequence.rs
@@ -757,6 +757,7 @@ impl Sequence {
             *self.state.read().unwrap(),
             SequenceState::FinishedAborted
                 | SequenceState::FinishedIgnored
+                | SequenceState::Error
                 | SequenceState::Done(_)
         )
     }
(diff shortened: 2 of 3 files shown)


glaziermag:
**Update (2026-04-15):** Rebased onto `origin/master`. Branch was previously based on the fork master, behind upstream by ~40 commits. No code changes — all three fixes are intact:
- `SequenceState::Error` added to `is_finished_paged_attn`
- `get_mut_arcmutex!($scheduler).free_finished_sequence_groups()` called before `continue $label` in the error macro
- `.unwrap()` replaced with `let _ =` for responder sends in the error path

glaziermag:
**Second update (2026-04-15):** Fixed a spin-lock deadlock introduced by the macro change.

The 7-arg `handle_pipeline_forward_error!` arm called `get_mut_arcmutex!($scheduler)` inside the error branch. The DefaultScheduler callsites at `engine/mod.rs` lines 410 and 472 both hold the scheduler `MutexGuard` (via `get_mut_arcmutex!(self.scheduler)` at line 360) through the entire match arm. Calling `get_mut_arcmutex!` again on the same `Arc` from the same thread causes an infinite spin — `Mutex` is not reentrant in Rust, and `get_mut_arcmutex!` loops on `try_lock()`.

**Fix:** Restored the original 6-arg arm for DefaultScheduler callsites. Omitting `free_finished_sequence_groups()` there is safe — DefaultScheduler calls it unconditionally at the bottom of the engine loop, so KV state is always freed by the next iteration. The 7-arg arm (and the explicit KV free) is used only by the paged-attention callsite at line 696, where the scheduler guard is held only inside the `res = { ... }` block and is released before the macro is invoked.

glaziermag:
Closing: this PR is superseded by #2044 (waiting-queue retain fix) and #2076 (SequenceState::Error in is_finished_paged_attn). The 123-line macro duplication of handle_pipeline_forward_error! creates a maintenance hazard — the 7-arg variant will drift from the 6-arg original over time. The scheduler lock invariant (callers must drop the guard before invoking the macro) is not enforced by the type system and is a deadlock footgun. The individual fixes this PR bundles are better served by #2044 and #2076 independently.

PR #2068, opened 2026-04-06
Fix Responses API background=true + stream=true panic (#1945)

### LINKED ISSUE #1945: Responses API: background=true + stream=true fails with 'Unexpected response type'
`background: true` combined with `stream: true` currently fails background tasks with `"Unexpected response type"`.

## Summary

In the `create_response` background branch, the spawned task waits for exactly one message from `bg_rx.recv().await` and treats only `Some(Response::Done(_))` as success.

If the request is streaming, the model pipeline can emit `Response::Chunk(_)` first. That variant is not handled in the match and falls into the default `_` arm, which marks the task as failed with `unknown_error: Unexpected response type`.

## Affected code

- `mistralrs-server-core/src/responses.rs` background path in `create_response`
- Match on first received response:
  - handles `Some(Response::Done(_))`
  - falls through to `_ => mark_failed("Unexpected response type")`

## Logic proof

1. Request is accepted with `background=true` and `stream=true`.
2. Background worker does `match bg_rx.recv().await` once.
3. If first event is `Response::Chunk(_)`, it is unmatched.
4. `_` arm executes and task is marked failed.

So this is not just “background ignores streaming”; it actively fails on streamed chunk responses.

## Expected behavior

`background=true` and `stream=true` should not fail with `Unexpected response type` purely because chunk events are emitted before terminal completion.

glaziermag:
Fixes #1945. Alternative to closed PR #1989.

## Problem

The `/v1/responses` endpoint supports a `background=true` mode that spawns a Tokio task to process the request asynchronously. When `stream=true` was also set, the pipeline emitted a sequence of `Response::Chunk(…)` messages followed by a terminal `Response::Done`. The background receiver used a single `match bg_rx.recv().await` — it hit the first `Chunk` message, fell into the `_ =>` catch-all, and marked the task as failed with `"Unexpected response type"`.

## Fix

Replace the single `match` with a `while let Some(msg) = bg_rx.recv().await` loop. `Response::Chunk(_)` arms `continue` to drain stream chunks; the terminal variants (`Done`, `ModelError`, `ValidationError`, `InternalError`) break the loop with the appropriate outcome.

## Scope

This fix applies to background (`background=true`) chat completion requests. Non-background streaming requests go through a different path and are unaffected. Other response variants (`CompletionDone`, `ImageGeneration`, etc.) not applicable to this endpoint fall into the `_ =>` arm as before — that is pre-existing behavior for endpoint misuse, not a regression.

## Files changed

- `mistralrs-server-core/src/responses.rs`

### DIFF
--- a/mistralrs-server-core/src/responses.rs
+++ b/mistralrs-server-core/src/responses.rs
@@ -1559,59 +1559,67 @@ pub async fn create_response(
             }
 
             // Wait for response
-            match bg_rx.recv().await {
-                Some(Response::Done(chat_resp)) => {
-                    let response = chat_response_to_response_resource(
-                        &chat_resp,
-                        task_id.clone(),
-                        metadata_clone,
-                        &request_context,
-                    );
+            while let Some(msg) = bg_rx.recv().await {
+                match msg {
+                    Response::Chunk(_) => continue,
+                    Response::Done(chat_resp) => {
+                        let response = chat_response_to_response_resource(
+                            &chat_resp,
+                            task_id.clone(),
+                            metadata_clone.clone(),
+                            &request_context,
+                        );
 
-                    // Store if requested
-                    if store {
-                        let cache = get_response_cache();
-                        let _ = cache.store_response(task_id.clone(), response.clone());
+                        // Store if requested
+                        if store {
+                            let cache = get_response_cache();
+                            let _ = cache.store_response(task_id.clone(), response.clone());
 
-                        if let Some(mut history) = conversation_history {
-                            for choice in &chat_resp.choices {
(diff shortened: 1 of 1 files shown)


glaziermag:
**Update (2026-04-15):** Rebased onto `origin/master`. Branch was previously behind upstream by several commits. No code changes — the `while let Some(msg) = bg_rx.recv().await` fix with `Response::Chunk(_) => continue` is unchanged.

glaziermag:
Closing in favor of #2114, which is a clean reimplementation against current master with a minimal diff (no extraneous reindentation changes). The fix is the same: loop-drain Response::Chunk events in the background task handler.

PR #2076, opened 2026-04-08
Fix Gemma softcap F16 overflow NaN and scheduler hang (#2058)

### LINKED ISSUE #2058: Gemma 4 E2B: inference hangs on complex prompts via MultimodalModelBuilder (GPU confirmed)
## Summary

When using `MultimodalModelBuilder` with `google/gemma-4-E2B-it` (safetensors), inference **hangs indefinitely** on prompts with a multi-paragraph system prompt + technical question. Simple prompts (short expected response) complete successfully. The hang is reproducible across three independent runs. GPU VRAM usage confirms the model is fully loaded onto the device — this is not a CPU fallback or loading issue.

## Environment

| | |
|---|---|
| mistralrs | 0.8.1 |
| OS | Windows 11 Pro 26200 |
| GPU | NVIDIA RTX 3080 (12 GB VRAM) |
| CUDA | 12.9 / Driver 591.86 |
| RAM | 64 GB |
| Rust | 1.94.1 stable-x86_64-pc-windows-msvc |

## Reproduction

```rust
use futures::StreamExt;
use mistralrs::{MultimodalModelBuilder, Response, TextMessageRole, TextMessages};

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let model = MultimodalModelBuilder::new("google/gemma-4-E2B-it")
        .build()
        .await?;

    // PASS — simple prompt, short expected response
    let simple = TextMessages::new()
        .add_message(TextMessageRole::System, "You are a helpful assistant.")
        .add_message(TextMessageRole::User, "Say hello in one sentence.");

    let mut stream = model.stream_chat_request(simple).await?;
    while let Some(chunk) = stream.next().await {
        if let Response::Chunk(c) = chunk {
            if let Some(content) = c.choices.first().and_then(|ch| ch.delta.content.as_deref()) {
                print!("{content}");
            }
        }
    }
    println!("\n--- simple prompt done ---");

    // HANG — complex system prompt + technical question
    let complex = TextMessages::new()
        .add_message(
            TextMessageRole::System,
            "You are an expert software engineer specializing in Rust programming. \
             You write clean, efficient, idiomatic code. You follow the Rust API guidelines \
             and prefer zero-copy patterns where possible. When explaining code, be thorough \
             but concise. Always consider error handling and edge cases.",
        )
        .add_message(
            TextMessageRole::User,
            "Explain the difference between Arc<Mutex<T>> and Arc<RwLock<T>> in Rust. \
             When should you use each? Give a brief code example.",
        );

    // This call never returns — stream.next() hangs indefinitely
    let mut stream = model.stream_chat_request(complex).await?;
    tokio::time::timeout(std::time::Duration::from_secs(120), async {
        while let Some(_) = stream.next().await {}
    })
    .await
    .expect("timed out after 120s — no chunks received");

    Ok(())
}
```

## Observed Behaviour

- **Simple prompt**: Returns promptly. Observed output: `"Hello, I am ready to assist you."`
- **Complex prompt**: `stream.next().await` never resolves. No chunks are emitted. `tokio::time::timeout` fires at 120s every run.

## GPU VRAM Evidence

Monitored with `nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits` during the test:

```
Before test:   651 MiB  (background)
During load: 11,612 MiB  ← model fully mapped to GPU
During hang: 11,644 MiB  ← stays loaded, inference stalled
After kill:    651 MiB
```

Peak VRAM: **11,676 MiB** of 12,288 MiB available. The model is unambiguously on GPU. This is not a slow CPU fallback.

## Reproducibility

| Run | Simple prompt | Complex prompt |
|-----|--------------|----------------|
| 1 (fresh download) | ✅ PASS | ❌ Hang — 120s timeout |
| 2 (cached model) | ✅ PASS | ❌ Hang — 120s timeout |
| 3 (cached model) | ✅ PASS | ❌ Hang — 120s timeout |

## Additional Notes

- Model loaded via `MultimodalModelBuilder::new("google/gemma-4-E2B-it").build()`
- `dtype: ModelDType::Auto` (default — bfloat16 on CUDA)
- `device_mapping: None` (automatic — confirmed GPU is used per VRAM evidence)
- The hang occurs before any tokens are emitted — `stream.next()` never returns at all, not just slow generation
- The boundary between pass and fail appears to be system prompt length/complexity. A 1-sentence system prompt works; a 5-sentence technical one does not.

## Hypothesis

Attention computation numerical instability triggered by input sequence length or complexity. The model loads and inference begins (VRAM consumed), but the sampling loop stalls — possibly NaN values propagating through attention weights or softmax, causing token sampling to loop infinitely or produce only padding tokens that never satisfy the stop condition.

This may be the same class of issue as #2051 (NaN logits on MoE Gemma 4 variants with BF16/UQFF), but manifesting on the dense E2B architecture through a different trigger (input complexity rather than model architecture). The original #2051 report notes a post-release fix commit on `main` for MoE variants — it's unclear whether that fix also covers the dense path.

glaziermag:
Fixes #2058.

### 1. Softcap NaN in F16/BF16 (naive attention backend)

Gemma models use an attention softcap (logit / softcap → tanh → logit * softcap). The naive SDPA backend was performing the tanh in the input dtype (F16 or BF16). For F16, values outside approximately ±65504 become ±Inf before tanh, and tanh(Inf) = NaN when computed in reduced precision, producing silent NaN logits.

**Fix:** Promote attention scores to F32 before computing the softcap tanh, then cast back to the original dtype.

**Scope note:** This fix applies to the CPU/naive fallback path in `attention/backends/naive.rs`. The primary CUDA (FlashAttention) and Metal SDPA backends have their own softcap handling and are not changed here.

### 2. `SequenceState::Error` missing from `is_finished_paged_attn`

When a sequence enters `SequenceState::Error`, it was not recognized as "finished" by the paged attention scheduler. This caused KV blocks to remain allocated and the scheduler to stall waiting for capacity that would never be returned.

**Fix:** Add `SequenceState::Error` to the match arms in `is_finished_paged_attn()`.

## Files changed

- `mistralrs-core/src/attention/backends/naive.rs`

### DIFF
--- a/mistralrs-core/src/attention/backends/naive.rs
+++ b/mistralrs-core/src/attention/backends/naive.rs
@@ -36,9 +36,16 @@ pub(crate) fn naive_sdpa(
             MatMul.matmul_affine_mul(q_chunk, &k.t()?, sdpa_params.softmax_scale.into())?;
 
         if let Some(softcap) = sdpa_params.softcap {
+            let att_dtype = att.dtype();
+            if att_dtype == candle_core::DType::BF16 || att_dtype == candle_core::DType::F16 {
+                att = att.to_dtype(candle_core::DType::F32)?;
+            }
             att = (att / softcap as f64)?;
             att = att.tanh()?;
             att = (att * softcap as f64)?;
+            if att.dtype() != att_dtype {
+                att = att.to_dtype(att_dtype)?;
+            }
         }
 
         if let Some(mask) = mask_chunk {
--- a/mistralrs-core/src/sequence.rs
+++ b/mistralrs-core/src/sequence.rs
@@ -757,6 +757,7 @@ impl Sequence {
             *self.state.read().unwrap(),
             SequenceState::FinishedAborted
                 | SequenceState::FinishedIgnored
+                | SequenceState::Error
                 | SequenceState::Done(_)
         )
     }


glaziermag:
**Update (2026-04-15):** Rebased onto `origin/master` (upstream upstream HEAD). The branch was previously based on the fork's master, which was behind by ~40 commits. No code changes — both fixes (softcap F32 promotion in `naive.rs` and `SequenceState::Error` in `is_finished_paged_attn`) are unchanged.

glaziermag:
Closing in favor of atomic split PRs for single-responsibility review:

- **#2110** — softcap F16/BF16 NaN overflow fix (naive.rs only)
- **#2111** — scheduler error-state hang fix (sequence.rs only)

The two fixes are independent and should be reviewable/mergeable separately.

PR #2108, opened 2026-04-14
Fix: Resolve NaN Logit Memory Leak During Sequential Benchmarks (Issue #2095)

### LINKED ISSUE #2095: crashes with NaN/Inf logits on iteration 2+ when --gen-len >= 512
## Summary

`mistralrs bench` reliably crashes on the second iteration with "Invalid sampling probability: NaN" when `--gen-len` is set to 512 or higher. Using `--gen-len 256` is stable across all iterations.

## Environment

| Field | Value |
|-------|-------|
| **GPU** | NVIDIA GeForce RTX 5080 |
| **Compute Capability** | 12.0 |
| **VRAM** | 16303 MiB |
| **GPU Driver** | 595.97 |
| **OS** | Ubuntu 24.04.4 LTS (Noble Numbat), WSL2 on Windows 11 |
| **CUDA** | 13.2 (nvcc V13.2.51, built 2026-03-02) |
| **cuDNN** | 9.20.0.48 (`cudnn9-cuda-13-2`) |
| **Rust** | 1.94.1 (e408947bf 2026-03-25), LLVM 21.1.8 |
| **Install method** | `cargo install mistralrs-cli --version 0.8.1 --features cuda,cudnn,flash-attn` (compiled from crates.io during install) |
| **mistralrs version** | 0.8.1 |
| **Cargo features** | `cuda`, `cudnn`, `flash-attn` |
| **Model** | google/gemma-4-E4B-it (q8_0 UQFF) |

## Reproduction

```bash
# Crashes on iteration 2
mistralrs bench --from-uqff /path/to/e4b-q8_0-0.uqff \
  -m google/gemma-4-E4B-it \
  --paged-attn off \
  --prompt-len 512 --gen-len 512 --iterations 5 --warmup 1

# Stable — no crash
mistralrs bench --from-uqff /path/to/e4b-q8_0-0.uqff \
  -m google/gemma-4-E4B-it \
  --paged-attn off \
  --prompt-len 512 --gen-len 256 --iterations 5 --warmup 1
```

## Error Output

```
INFO  mistralrs::commands::bench: Iteration 1/5...
INFO  mistralrs_core::engine::logger: Throughput (T/s) 128.40, Prefix cache hitrate 0.00%, 1 running, 0 waiting
INFO  mistralrs::commands::bench: Iteration 2/5...
Error: Invalid sampling probability at index 2: NaN. The model likely produced NaN/Inf logits.
```

Iteration 1 completes successfully. The crash consistently occurs at the start of iteration 2 (during the prefill or early decode of the second request).

## Notes

- The crash is **not** triggered at `--gen-len 256`, even with 5+ iterations
- The crash is **not** triggered on the first iteration regardless of gen-len
- This suggests the KV cache or some internal state is not being fully reset between bench iterations
- The bench tool sends `RequestMessage::CompletionTokens` with sequential token IDs (`1000..1000+prompt_tokens`), so this is not a tokenizer issue
- The `--warmup` run uses only 32 prompt / 16 gen tokens and completes fine

## Suspected Root Cause

The bench tool reuses the same `MistralRs` instance across iterations. If the KV cache or sequence state from iteration 1 is not fully evicted before iteration 2 begins, the model may see stale KV entries that corrupt the logit distribution.

Looking at `mistralrs-cli/src/commands/bench.rs` line 106:
```rust
if let Ok(logger) = mistralrs.get_logger(None) {
    logger.reset();
}
```
The logger is reset after warmup, but there is no equivalent KV cache flush between benchmark iterations. Each `run_single_bench` call issues a fresh `NormalRequest`, but if the engine's sequence table retains entries from the previous request (e.g. due to caching), the second request may inherit corrupted state.

## Workaround

Use `--gen-len 256` or lower. This is stable across all tested iterations.

glaziermag:
Fixes #2095.

## Problem

Between consecutive `mistralrs bench` iterations, the previous benchmark's sequences were evicted asynchronously. The next iteration could enter the engine pipeline before the flush completed, carrying over stale KV state from the prior run. This produced NaN logits in the sampler and panics on long `--gen-len` runs.

## Root cause

`TerminateAllSeqsNextStep` sets a global `AtomicBool` (`TERMINATE_ALL_NEXT_STEP`). The engine reads this flag at the **top** of each scheduler loop iteration. Without a delay, the next benchmark request could enter the channel before the engine had a single iteration to act on the flag.

## Fix

After sending `TerminateAllSeqsNextStep`, wait 50 ms to give the engine at least one full scheduler loop iteration to read the flag and flush sequences before the next request enters the channel. This applies after warmup and between all benchmark iterations.

## Files changed

- `mistralrs-cli/src/commands/bench.rs`

### DIFF
--- a/mistralrs-cli/src/commands/bench.rs
+++ b/mistralrs-cli/src/commands/bench.rs
@@ -102,6 +102,10 @@ pub async fn run_bench(
         }
         info!("Warmup complete.");
 
+        // Clear any residual KV cache from warmup
+        let sender = mistralrs.get_sender(None).unwrap();
+        let _ = sender.send(mistralrs_core::Request::TerminateAllSeqsNextStep).await;
+
         // Reset logger counters so benchmark stats are clean
         if let Ok(logger) = mistralrs.get_logger(None) {
             logger.reset();
@@ -142,6 +146,10 @@ pub async fn run_bench(
             let ms_per_tok = 1000.0 / tok_per_sec;
             decode_results.push((tok_per_sec, ms_per_tok));
         }
+
+        // Extremely aggressive cache sweep. Flush the sequences out of the engine.
+        let sender = mistralrs.get_sender(None).unwrap();
+        let _ = sender.send(mistralrs_core::Request::TerminateAllSeqsNextStep).await;
     }
 
     // Calculate statistics
--- a/mistralrs-cli/src/commands/bench.rs
+++ b/mistralrs-cli/src/commands/bench.rs
@@ -8,7 +8,7 @@ use mistralrs_core::{
 };
 use mistralrs_server_core::mistralrs_for_server_builder::MistralRsForServerBuilder;
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::mpsc::channel;
 use tracing::info;
 
@@ -102,9 +102,12 @@ pub async fn run_bench(
         }
         info!("Warmup complete.");
 
-        // Clear any residual KV cache from warmup
+        // Flush any residual KV state left over from warmup, then yield to the
+        // engine's scheduler loop so the termination is processed before we start
+        // timing. Without the sleep the next request can race ahead in the channel.
         let sender = mistralrs.get_sender(None).unwrap();
         let _ = sender.send(mistralrs_core::Request::TerminateAllSeqsNextStep).await;
+        tokio::time::sleep(Duration::from_millis(50)).await;
 
         // Reset logger counters so benchmark stats are clean
         if let Ok(logger) = mistralrs.get_logger(None) {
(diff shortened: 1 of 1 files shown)


glaziermag:
**Update (2026-04-15):** Fixed a race condition in the original fix.

The initial implementation sent `TerminateAllSeqsNextStep` fire-and-forget then immediately enqueued the next benchmark iteration request in the same MPSC channel. Because both messages enter the same channel, the engine scheduler loop could process them in any order — the next iteration's request could begin before the termination was handled, producing the same stale KV state the fix was meant to clear.

**Correction applied:** Added `tokio::time::sleep(Duration::from_millis(50))` after each `TerminateAllSeqsNextStep` send. This yields to the Tokio runtime, giving the engine at least one full scheduler pass to drain the termination message before the next run begins. The 50ms is a conservative floor well below benchmark iteration time.

glaziermag:
**Second update (2026-04-15):** Corrected the in-code comment explaining the flush mechanism.

The comments incorrectly described the flush as preventing MPSC channel ordering nondeterminism. The actual mechanism: `TerminateAllSeqsNextStep` sets `TERMINATE_ALL_NEXT_STEP`, a global `AtomicBool`. The engine reads this flag at the top of each scheduler loop iteration. The `sleep(50ms)` gives the engine at least one full loop iteration to read and act on the flag before the next benchmark request enters the channel. No behavior changes.

glaziermag:
Closing in favor of #2112, which replaces the sleep(50ms) synchronization with a proper request/response barrier. The probe approach guarantees the engine has processed the termination without timing dependencies.

PR #2116, opened 2026-04-21
Pluggable KvCacheCodec hook on SingleCache / RotatingCache

ravituringworks:
## Motivation

We're prototyping experimental KV-cache codecs (e.g. PolarQuant + QJL,
~3 bits/dim) on top of mistral.rs and want to validate them without
maintaining a long-lived fork. Today there's no seam to install per-layer
encode/decode logic on the cache buffers; this PR adds one.

## Design

A new `KvCacheCodec` trait in `mistralrs-core::kv_cache::codec`:

```rust
pub trait KvCacheCodec: Send + Sync + std::fmt::Debug {
    fn encode(&self, tensor: &Tensor) -> Result<Tensor>;
    fn decode(&self, tensor: &Tensor) -> Result<Tensor>;
    fn name(&self) -> &str;
}
```

`SingleCache` and `RotatingCache` each carry an `Option<Arc<dyn KvCacheCodec>>`.
The default is `None`, which short-circuits to the **bit-exact** existing
`slice_set` / `narrow` paths — every call site that doesn't install a codec is
unchanged. When a codec is present, `append` runs `encode` before storing and
`current_data` runs `decode` before returning the narrowed view. A
`PassthroughCodec` ships as the canonical identity reference.

**Shape + dtype preservation** is part of the contract — the codec must return
a tensor of the same shape and dtype it received. That keeps the underlying
buffer layout uniform so `slice_set` / `narrow` need no changes. True
packed-storage codecs (different bit widths in the buffer) need a richer
trait surface; deliberately out of scope for this landing.

## Install-all helpers

`KvCache::set_codec(codec)` fans out to both K and V sub-caches (no-op on
`Shared`). `EitherCache::set_kv_cache_codec(codec) -> usize` walks every
attention layer of a `Normal` or `Hybrid` cache (skipping recurrent layers,
which have no codec concept) and returns the install count for logging.

## SDK surface

`MistralRs::set_kv_cache_codec(codec, model_id)` resolves the engine, locks
the pipeline behind `RebootState`, and forwards to the cache helper. Async
because the pipeline mutex is `tokio::sync::Mutex`; the lock is held only
for the duration of a pointer-clone per layer. `MistralRs::kv_head_dims`
returns `(k_head_dim, v_head_dim)` from `GeneralMetadata::model_metadata` so
downstream callers can size the codec from the loaded model rather than
hard-coding a per-architecture constant. Both surface on `Model` as
`set_kv_cache_codec[_with_model]` / `kv_head_dims[_with_model]`.

## Incidental

One commit (`67d4eba`) silences a pre-existing `dead_code` warning on
`MoEExperts::num_experts` — uncovered while running `cargo check` on the
branch. Happy to drop it if you'd prefer a separate PR.

## Test plan

- [x] `cargo test -p mistralrs-core kv_cache` — 8 tests covering codec
  round-trip on both `SingleCache` and `RotatingCache`, install-all on
  `Normal`-typed `EitherCache`, and the bit-exact `None` default.
- [x] `cargo check -p mistralrs-core -p mistralrs` clean.
- [x] Downstream consumer installs a candle-backed PolarQuant + QJL codec
  via `Model::set_kv_cache_codec` without touching mistral.rs internals.

## Compatibility

- `None` codec ⇒ binary-identical behaviour to current `master`. Every
  existing model / pipeline path that doesn't opt in is unaffected.
- New crate-root re-exports (`KvCache`, `SingleCache`, `RotatingCache`,
  `KvCacheCodec`, `KvCacheCodecRef`, `PassthroughCodec`) so downstream crates
  can install codecs without depending on private modules.

Happy to split into smaller PRs (trait + threading / install-all / SDK
accessors) if that's easier to review.


ravituringworks:
Hi @EricLBuehler — Copilot's latest five comments on commit `e3c6c5e` are stale: it re-anchored on the same `codec:` lines but the fix changes the **binding** above them.

- `mod.rs:399` / `424` (clone_in_cache) now bind `{ k: old_k, v: old_v }` and use `old_v.codec.clone()` for V (lines 420, 447).
- `mod.rs:593` / `665` (set_none_cache) now bind `{ k, v }` and use `v.codec.clone()` for V (lines 611, 684).
- `full_cache.rs:62-72`: `KvCache::set_codec` now returns `bool` (true for Normal/Rotating, false for Shared); the Normal branch only increments `count` on `true`, and the docstring + test (`normal_installs_codec_on_every_layer`, expected count 4 → 3) were updated to match.

`cargo test -p mistralrs-core --lib kv_cache` → 22/22 pass.

Issue #2171, opened 2026-05-26
RFC: GGUF backend for Gemma 4 (text-only, E2B-it first)


fiorelorenzo:
## Background

`GGUFArchitecture` at `mistralrs-core/src/gguf/mod.rs:16-33` lists `Llama, Mpt, Gptneox, Gptj, Gpt2, Bloom, Falcon, Mamba, Rwkv, Phi2, Phi3, Starcoder2, Qwen2, Qwen3, Qwen3MoE, Mistral3`. No Gemma variant, including no Gemma 4. The dispatch at `mistralrs-core/src/pipeline/gguf.rs:475-485` covers the same subset minus the no-op variants. Loading a Gemma 4 GGUF through `GgufModelBuilder` therefore fails at `mistralrs-core/src/gguf/content.rs:151:22` with `Unknown GGUF architecture "gemma4"` (this is what issue #2098 reports).

Gemma 4 is otherwise reachable only via `MultimodalLoaderType::Gemma4` over safetensors / UQFF (`mistralrs-core/src/pipeline/loaders/multimodal_loaders.rs:222`, model sources at `mistralrs-core/src/vision_models/gemma4/`). That path unconditionally loads the vision and audio encoders (#2099) even for text-only inference.

llama.cpp's GGUF spec already defines the architecture string `gemma4` (`ggml-org/llama.cpp:gguf-py/gguf/constants.py:947`) and the full per-layer tensor list at `:2450-2483`, so the naming convention is upstream-stable.

## Use case

Gemma 4 E2B-it Q4_K_M (~1.5 GB) as a text-only chat model for cleanup-style post-processing on consumer Apple Silicon, in the same load shape as Gemma 2 and Gemma 3 GGUF deployments via llama.cpp today. Issue #2098 reports the same need from a different downstream.

## Proposal

Add `Gemma4` to `GGUFArchitecture` and implement a text-only GGUF model in `mistralrs-core/src/models/quantized_gemma4.rs`, mapping the llama.cpp tensor names to a quantized variant of the Gemma 4 text core currently at `mistralrs-core/src/vision_models/gemma4/text.rs`.

Scope of this first PR:

- Dense E2B path only. The MoE-block fields (`enable_moe_block`, `num_experts`, `top_k_experts`, FFN_*_EXP tensors at `constants.py:2464-2467`) are out of scope for this PR and would land next as a `quantized_gemma4_moe.rs`, mirroring the Qwen3 split.
[...]

PR #2172, opened 2026-05-26
Add GGUF backend for Gemma 4 (text-only, E2B-it first)

### LINKED ISSUE #2098: GGUF pipeline panics on `gemma4` architecture instead of returning an error
## Summary

When loading a Gemma 4 GGUF file, the GGUF pipeline panics with an unhandled `unwrap()` on an unknown architecture string instead of returning a graceful error. The `auto` subcommand correctly identifies the file as GGUF before the panic occurs.

## Environment

| Field | Value |
|-------|-------|
| **GPU** | NVIDIA GeForce RTX 5080 |
| **Compute Capability** | 12.0 |
| **VRAM** | 16 GB |
| **OS** | Windows 11 |
| **mistralrs version** | 0.8.1 |
| **Model** | google/gemma-4-26B-A4B-it (Unsloth GGUF: `gemma-4-26B-A4B-it-UD-Q4_K_S.gguf`) |

## Reproduction

```bash
mistralrs bench ... auto -m <local-gguf-dir> --format gguf -f gemma-4-26B-A4B-it-UD-Q4_K_S.gguf ...
```

## Error Output

```
thread 'main' panicked at mistralrs-core-0.8.1/src/gguf/content.rs:151:22:
called `Result::unwrap()` on an `Err` value: Unknown GGUF architecture `gemma4`
```

## Notes

- Unsloth's pre-quantized GGUF files for Gemma 4 use `gemma4` as their architecture identifier in the GGUF metadata.
- The `auto` subcommand correctly detects the file as GGUF (`Model kind is: gguf quantized from gguf`) before the panic occurs at the architecture lookup.
- All Gemma 4 GGUF variants are affected: `UD-Q4_K_S`, `UD-Q5_K_S`, `MXFP4_MOE`, etc.
- `IQ4_XS` variants fail earlier with a different error (unsupported quantization type).

## Expected Behavior

Return a clear `Err` (e.g. `"Unsupported GGUF architecture: gemma4. Please raise an issue."`) rather than panicking. Ideally, add support for the `gemma4` architecture so Gemma 4 GGUFs can be loaded.

fiorelorenzo:
Implements the text-only dense Gemma 4 GGUF backend proposed in #2171.

## What lands

`Gemma4` joins `GGUFArchitecture` (`mistralrs-core/src/gguf/mod.rs`) and routes through the dispatcher at `mistralrs-core/src/pipeline/gguf.rs:475-485`. New `mistralrs-core/src/models/quantized_gemma4.rs` maps the upstream gguf-py tensor table (`ggml-org/llama.cpp:gguf-py/gguf/constants.py:2450-2483`) to a quantized variant of the existing Gemma 4 text core at `vision_models/gemma4/text.rs:240-398`. Templated on `quantized_qwen3.rs` with the Gemma 4 deltas read from GGUF metadata:

- Sliding / global head_dim split (256 / 512) via `attention.key_length_swa` / `attention.value_length_swa` and the `attention.sliding_window_pattern` cadence.
- Per-head Q-K-V RmsNorm via `blk.<i>.attn_<qkv>_norm.weight`.
- Proportional rotary on full-attention layers with `rope.freq_base` / `rope.freq_base_swa`.
- KV-shared layers via `attention.shared_kv_layers`; donor pattern matches `repos/llama.cpp/src/models/gemma4.cpp:3-29`.
- Final logit softcapping when `final_logit_softcapping` is present.

Q4_K_M, Q5_K_M, Q8_0 supported (same surface as Qwen3 GGUF).

## Out of scope (explicitly)

- MoE Gemma 4. Detection-and-bail on any `blk.<i>.ffn_*_exp.weight` with a clear error pointing at a follow-up `quantized_gemma4_moe.rs`.
- Multimodal Gemma 4 (vision + audio encoders). Vision/audio tensor tables in `constants.py:743-867` are orthogonal and large; deferred.
- MTP speculative decoding on the GGUF path. #2159 added it on the safetensors side; mirroring is a separate concern.
- Smoke / parity test against real weights. The two RFC open questions (file-naming convention, MoE detection vs day-one separate enum variant) are still worth a maintainer answer before adding a smoke; see #2171.

## Punted

- `partial_rotary_factor` hardcoded to 0.25 (upstream default). The Gemma 4 GGUF spec does not surface this. If you want it metadata-driven, point at the key name and I'll add it.
- Shared-KV layers still load placeholder `k_norm` / `v_norm` (ones). RoPE on shared layers touches only Q, so the placeholders never feed forward, but they keep the per-layer struct uniform. Happy to gate them behind an `Option` if you prefer.

## Build

```
cargo build --release --features metal -p mistralrs-core
```

Clean on macOS 26.2 / Xcode 26.3 / Metal Toolchain 32023.864 against `upstream/master @ 31c13eb`. No new warnings, `cargo fmt --check` clean for the touched files.

Closes #2098 once landed. RFC at #2171.

### DIFF
--- a/mistralrs-core/src/gguf/gguf_tokenizer.rs
+++ b/mistralrs-core/src/gguf/gguf_tokenizer.rs
@@ -107,17 +107,28 @@ pub fn convert_gguf_to_hf_tokenizer<R: std::io::Seek + std::io::Read>(
         }
     };
 
-    //token type other than 1 treated as special token
-    let mut num_special_tokens = 0;
-    #[allow(clippy::needless_range_loop)]
+    // Token types other than `NORMAL` (1) are treated as special tokens.
+    //
+    // Batch all of them into a single `add_special_tokens` call. The
+    // previous per-token loop triggered one `AddedVocabulary::refresh_added_tokens`
+    // per token, and each refresh rebuilds the aho-corasick failure-link
+    // automaton over every special token added so far. That is O(N^2) in
+    // calls and becomes effectively infinite for Gemma 3 (6670 CONTROL +
+    // BYTE + USER_DEFINED tokens in the 262144-vocab GGUF), making the
+    // load look like a hang. Llama / Qwen happen to ship <50 such tokens
+    // each, which kept this slow path invisible until Gemma 3/4 landed in
+    // the GGUF dispatch.
+    let mut special = Vec::new();
     if token_types.len() == props.tokens.len() {
-        for i in 0..props.tokens.len() {
-            if token_types[i] != 1i32 {
-                let tk = props.tokens[i].clone();
-                tokenizer.add_special_tokens(&[AddedToken::from(tk.to_string(), true)]);
-                num_special_tokens += 1;
+        for (i, ty) in token_types.iter().enumerate() {
+            if *ty != 1i32 {
+                special.push(AddedToken::from(props.tokens[i].clone(), true));
             }
         }
+    }
+    let num_special_tokens = special.len();
+    if !special.is_empty() {
+        tokenizer.add_special_tokens(&special);
     }
 
     info!(
--- a/mistralrs-core/src/gguf/gguf_tokenizer.rs
+++ b/mistralrs-core/src/gguf/gguf_tokenizer.rs
@@ -94,6 +94,14 @@ pub fn convert_gguf_to_hf_tokenizer<R: std::io::Seek + std::io::Read>(
     let (mut tokenizer, kind) = match props.model.as_str() {
         "llama" | "replit" => unigram_tokenizer(&props)?,
         "gpt2" => bpe_tokenizer(&props)?,
+        // Gemma family uses SentencePiece-style BPE: BPE merges over
+        // sub-word strings that already carry the `\u{2581}` (lower-one-eighth
+        // block) prefix in place of leading spaces, plus a byte-fallback
+        // range (`<0xNN>`). It is NOT GPT-2 byte-level BPE, so it must not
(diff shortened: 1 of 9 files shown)


fiorelorenzo:
Closing this in favour of two narrower PRs and dropping the Gemma 4 GGUF feature itself for the time being.

What was bundled here split as follows:

- **mistralrs-quant Metal build hardening** → #2176 (independent of any model work; fixes a silent `metallib` build failure on Apple Silicon).
- **Batched `add_special_tokens` in `convert_gguf_to_hf_tokenizer`** → #2177 (perf fix that benefits every GGUF tokenizer; surfaced as a load hang on the Gemma 3 1B `unsloth` GGUF before we figured out it was N² rather than a model bug).

The remaining work in this PR — Gemma 3 dense GGUF backend, Gemma 4 dense GGUF backend with PLE, SentencePiece-BPE tokenizer arm, gemma3/4 `gguf_metadata` size estimators, per-layer `NormalCacheType`, sliding-window pattern — I'm pulling because the consumer-side project that motivated it (a local dictation app) pivoted away from mistral.rs for its LLM runtime. The Gemma 4 PLE path also needs a chunked / quantized-row embedding lookup to avoid a ~9.4 GB transient F32 buffer during `QTensor::dequantize`, which I'm not going to land without an end-to-end memory budget plan that I no longer have a reason to author.

Apologies for the monolithic shape this PR landed in — the two narrow PRs above are independently reviewable and shouldn't carry any of the Gemma surface area. RFC #2171 closing as well.

Thanks for the engine work that made the build-hardening and tokenizer perf issues this easy to isolate.

PR #2176, opened 2026-05-28
build(mistralrs-quant): harden Metal kernel build against silent failures

fiorelorenzo:
## Summary

Two independent fixes to `mistralrs-quant/build.rs`. Both address silent
failure modes that have a way of producing empty 118-byte `.metallib`
artifacts that Cargo caches, surfacing later as `Error while loading
function: …` panics at runtime — i.e. far from the build step that
actually broke.

### 1. `xcrun metal` exit code is now checked (commit 1/2)

The old code did
```rust
Command::new("xcrun").args(["metal", ...]).spawn()?.wait()?;
```
and ignored the resulting `ExitStatus`. If the Metal toolchain emits a
nonzero exit but still writes a partial output, the build script reports
success and Cargo caches a broken `.air` file. Replace with
`Command::output()` + an explicit non-success check that surfaces both
stdout and stderr.

### 2. `air -> metallib` link step now passes `--sdk` and `-std=…` (commit 2/2)

When the spawn-and-wait variant of the link step is used without
`--sdk macosx` and an explicit `-std=metal3.0` (or equivalent), `air-lld`
silently ignores AIR 2.6 inputs (default target falls back to AIR 2.3)
and exits 0 anyway. The resulting `mistralrs_quant.metallib` is the
118-byte container header with no functions. At Metal runtime this
shows up as `Error while loading function: <kernel_name>` for every
kernel lookup — but only on developer machines whose `xcrun -sdk
macosx`-selected toolchain happens to default to a lower AIR target.

Pass both flags explicitly so the link is deterministic across Xcode
versions.

## Symptom that motivated the patches

Apple Silicon dev install: model loads, every Metal kernel call panics
with `Error while loading function: ...` despite a clean build. Looking
in `target/.../mistralrs_quant.metallib` shows a 118-byte file.
Reproduced on Xcode 26, macOS 26.

## Out of scope

- Same hardening for `mistralrs-paged-attn/build.rs` (similar pattern,
  not exercised on Apple Silicon by the failure mode above; happy to do
  in a follow-up if useful).
- Pinning a specific `metal3.x` target — the patch uses whatever the
  current SDK provides.

### DIFF
--- a/mistralrs-quant/build.rs
+++ b/mistralrs-quant/build.rs
@@ -248,15 +248,7 @@ fn main() -> Result<(), String> {
             for metal_file in HEADER_SOURCES {
                 compile_air_cmd.arg(sources.join(format!("{metal_file}.metal")));
             }
-            // Use `.status()` so we actually observe the exit code; the
-            // previous `.spawn().expect().wait().expect()` pattern only
-            // checked for IO failures, and a non-zero `xcrun metal` exit
-            // would slip through silently. When that happened, the
-            // metallib step further down ran with zero `.air` inputs and
-            // emitted a ~118-byte empty archive, which Cargo then cached
-            // forever (since the source `.metal` files had not changed),
-            // so every kernel lookup failed at runtime with
-            // `Error while loading function`.
+            // Run the Metal compiler once and fail the build on a nonzero exit.
             let status = compile_air_cmd
                 .status()
                 .expect("Failed to invoke metal compiler (metal -> air)");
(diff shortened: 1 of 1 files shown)


mistralrs-quant/build.rs
```diff
@@ -248,29 +248,20 @@ fn main() -> Result<(), String> {
...
-                Ok(None) => {
-                    let status = child
-                        .wait()
-                        .expect("Compiling metal -> air failed while waiting for result");
-                    if !status.success() {
-                        panic!("Compiling metal -> air failed. Exit with status: {status}")
-                    }
-                }
-                Err(e) => panic!("Compiling metal -> air failed: {e:?}"),
+            // Use `.status()` so we actually observe the exit code; the
```
EricLBuehler: I think a more appropriate comment would be something like `// Run the Metal compiler once and fail the build on a nonzero exit.`

mistralrs-quant/build.rs
```diff
@@ -281,7 +272,25 @@ fn main() -> Result<(), String> {
             };
             let metallib = out_dir.join(lib_name);
             let mut compile_metallib_cmd = Command::new("xcrun");
-            compile_metallib_cmd.arg("metal").arg("-o").arg(&metallib);
+            // Pass `--sdk` and `-std=...` to the air -> metallib link step
```
EricLBuehler: I think a simpler comment would be something like `// Keep the link step on the same SDK/std as the AIR compile step.`

EricLBuehler (changes requested):
Thank you for the PR @fiorelorenzo! I think the changes are good, but the comments explaining it are a bit too verbose. Can you please make the following suggested changes?

EricLBuehler (approved):
Thank you @fiorelorenzo!

PR #2178, opened 2026-05-29
fix(windows): Fix windows builds and cuDNN discovery

sobrinth:
Ran into some issues today building from source on windows.
Here some proposed changes to "make it work".

- Pass /Zc:preprocessor for msvc
- Add additional search path for cuDNN in install.ps1
- Add cuDNN lib path to build
fixes #2081 #2082


EricLBuehler (approved):
Thank you @sobrinth!

Issue #831, opened 2024-10-07
Runtime OTEL Metrics for Server


nikolaydubina:
Would be nice to have Prometheus metrics exported to OTEL (or at least basic /metrics). Useful for production.

https://crates.io/crates/runtime-otel-metrics

nikolaydubina:
if nobody takes this, I will add this myself in coming days.

MicheleCampi:
Hi! I saw this is still open and took a shot at it.
I put together a working proof-of-concept and wanted to check the approach with you before opening a PR.
Branch: https://github.com/MicheleCampi/mistral.rs/tree/feat/prometheus-metrics
What it does:

adds a GET /metrics endpoint exposing the Prometheus text exposition format
installs a process-wide recorder at router build time
adds an axum middleware recording http_requests_total (counter) and http_request_duration_seconds (latency), labeled by method, path, and status

It builds clean (CPU-only here), clippy -D warnings and the existing tests pass, and I've verified the endpoint live against a running server. The footprint is small — one new module plus a few lines in the router builder.
I deliberately kept the PoC minimal to confirm direction first. A few choices I'd like your input on:

1. Library: I used metrics + metrics-exporter-prometheus, since the crate already depends on tracing and this is the lightest companion for a pull-based /metrics. Issue #830 links opentelemetry-rust for traces — would you prefer I align on the OTEL stack from the start, or keep metrics on the lighter metrics facade and treat OTEL traces separately?
2. Histogram vs summary: the Prometheus exporter renders latency as a summary with quantiles by default. For HTTP request latency, bucketed histograms are often preferred since they aggregate across instances. Happy to configure either way — do you have a preference?
3. Inference-level metrics: the PoC instruments the HTTP layer. The natural next step is engine-level metrics (tokens, in-flight requests, queue depth) closer to what vLLM exposes. I'd scope those in a follow-up rather than this first PR — does that split make sense to you?

If the approach looks right, I'll open a PR. I work async and write more than I talk, so I'm glad to iterate here in writing.

MicheleCampi:
Thanks for merging #2189! As I flagged in my earlier comment, the natural next step here is the engine-level runtime metrics (throughput, in-flight requests, queue depth) — the part of this issue that goes beyond the HTTP layer.

[...]

Issue #830, opened 2024-10-07
Open Telemetry for server


nikolaydubina:
Support for OTEL Traces would be nice for running in production setting.

nikolaydubina:
if nobody takes this, I will add this myself in coming days.

LLukas22:
Would also love to see this implemented 👀

PR #2189, opened 2026-06-02
feat: add Prometheus /metrics endpoint

MicheleCampi:
Opening this as discussed in #831. I posted there a couple of days ago to
check the approach before opening a PR; opening it now so there's something
concrete to review, and I'm happy to adjust based on your preferences.

## What this adds
- a `GET /metrics` endpoint exposing the Prometheus text exposition format
- a process-wide Prometheus recorder installed at router build time
- an axum middleware recording `http_requests_total` (counter) and
  `http_request_duration_seconds` (latency), labeled by method, path, status

Footprint is small: one new module (`metrics.rs`) plus a few lines wiring the
route, recorder, and layer into the server router builder. Rebased on current
master.

## Notes / open questions (from #831)
- **Library**: uses `metrics` + `metrics-exporter-prometheus`, the lightest
  companion given the crate already depends on `tracing`. If you'd rather
  align on the `opentelemetry-rust` stack referenced in #830, I can rework it.
- **Histogram vs summary**: the exporter renders latency as a summary with
  quantiles by default; bucketed histograms aggregate better across instances.
  Happy to switch if you prefer.
- **Inference-level metrics** (tokens, in-flight requests, queue depth) would
  be a natural follow-up, kept out of this PR to keep it focused.

Builds clean, `clippy -D warnings` and existing tests pass, and I verified the
`/metrics` endpoint live against a running server.

### DIFF
--- a/mistralrs-server-core/src/metrics.rs
+++ b/mistralrs-server-core/src/metrics.rs
@@ -1,0 +1,44 @@
+//! Prometheus metrics support for the mistral.rs server.
+//! Installs a process-wide Prometheus recorder and exposes a `/metrics`
+//! endpoint rendering metrics in the Prometheus text exposition format.
+use axum::{extract::Request, middleware::Next, response::Response};
+use axum::{http::StatusCode, response::IntoResponse};
+use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
+use std::sync::OnceLock;
+use std::time::Instant;
+static PROMETHEUS_HANDLE: OnceLock<PrometheusHandle> = OnceLock::new();
+/// Install the global Prometheus recorder. Safe to call once at startup.
+pub fn install_prometheus_recorder() {
+    if PROMETHEUS_HANDLE.get().is_some() {
+        return;
+    }
+    let handle = PrometheusBuilder::new()
+        .install_recorder()
+        .expect("failed to install Prometheus recorder");
+    let _ = PROMETHEUS_HANDLE.set(handle);
+}
+/// Axum handler for `GET /metrics`. Renders the Prometheus exposition format.
+pub async fn metrics() -> impl IntoResponse {
+    match PROMETHEUS_HANDLE.get() {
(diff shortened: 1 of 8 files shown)


mistralrs-server-core/src/metrics.rs
```diff
@@ -0,0 +1,44 @@
...
+            "metrics recorder not initialized",
+        )
+            .into_response(),
+    }
+}
+/// Axum middleware that records per-request metrics: a request counter and a
+/// latency histogram, both labeled by method, path, and response status.
+pub async fn track_metrics(req: Request, next: Next) -> Response {
+    let method = req.method().to_string();
+    let path = req.uri().path().to_string();
```
EricLBuehler: Can we avoid using the raw URI path as the `path` label? Prometheus creates a separate time series for every unique label set, so routes like `/v1/files/{id}`, `/v1/responses/{response_id}`, sessions, approval ids, and arbitrary 404 paths would create unbounded cardinality.

Maybe with the matched route pattern instead (e.g. `/v1/files/{id}`), using axum's [`MatchedPath`](https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html)?

EricLBuehler (changes requested):
Thanks for the PR @MicheleCampi!

One small comment but it looks great.

Additionally, can you please update the readme and docs to reflect this feature, as well as how to use it? Thanks!

MicheleCampi:
Thanks for the review! Addressed both points in fa791a7e7:

Cardinality: the path label now uses axum's MatchedPath (the route pattern, e.g. /v1/responses/{id}) instead of the raw URI, so per-request ids and arbitrary 404 paths no longer create unbounded time series. Requests that match no route fall back to a single <unmatched> label.
Docs: documented the /metrics endpoint in the HTTP API reference (under System), and added a feature entry to the README.

EricLBuehler (approved):
Thank you @MicheleCampi!

PR #2231, opened 2026-06-15
feat(core): expose inference-level metrics via Prometheus

MicheleCampi:
Follow-up to #2189, which added the /metrics endpoint and the Prometheus recorder. That PR exposed transport-level metrics (HTTP request count and latency); this one adds the inference-level counters the engine already tracks, addressing the runtime-metrics direction discussed in #831.

Metrics added (all sourced from values the IntervalLogger already computes):
- inference_tokens_processed_total (counter): tokens processed
- inference_requests_running (gauge): sequences currently running
- inference_requests_waiting (gauge): sequences waiting in queue

Implementation notes:
- Emitted from the existing IntervalLogger loop, reusing the metrics facade already installed by the server. mistralrs-core gains only the metrics facade dependency; the recorder stays in the server.
- Emissions sit outside the existing log guard, so the gauges report current state on every interval including when idle, with no gaps in the scrape.
- The counter increments by the window's token count (the atomic is already swapped to 0 for the throughput log), so it stays monotonic and cumulative.
- Because they live in the IntervalLogger, the metrics share its sampling cadence and enable/disable switch: a deliberate choice to reuse the existing sampling point at zero added cost.

Verified locally against Qwen2.5-0.5B (GGUF). At idle all three report 0; after traffic the scrape shows inference_tokens_processed_total 294, inference_requests_running 1, inference_requests_waiting 0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3897,6 +3897,7 @@ dependencies = [
  "libc",
  "llguidance",
  "lrtable",
+ "metrics",
  "mime_guess",
  "minijinja",
  "minijinja-contrib",
--- a/mistralrs-core/Cargo.toml
+++ b/mistralrs-core/Cargo.toml
@@ -34,6 +34,7 @@ half.workspace = true
 accelerate-src = { workspace = true, optional = true }
 intel-mkl-src = { workspace = true, optional = true }
 tracing.workspace = true
+metrics.workspace = true
 rand.workspace = true
 rand_distr.workspace = true
 regex-automata.workspace = true
--- a/mistralrs-core/src/engine/logger.rs
+++ b/mistralrs-core/src/engine/logger.rs
@@ -57,6 +57,15 @@ impl IntervalLogger {
                 let num_running = t_num_running.load(Ordering::Relaxed);
                 let num_waiting = t_num_waiting.load(Ordering::Relaxed);
 
+                // Emit inference-level metrics every interval, including when
+                // idle, so the gauges report current state without gaps. The
+                // token counter increments by this window's count (the atomic
+                // was already swapped to 0 above for the throughput log).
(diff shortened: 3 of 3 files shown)


EricLBuehler (changes requested):
Thanks for the PR @MicheleCampi! This is close.

I found one correctness issue before merge: `IntervalLogger` is per-engine, but the new `inference_requests_running` and `inference_requests_waiting` gauges are unlabelled process-global metric names. In a multi-model server, each engine logger will write the same gauge every interval, so `/metrics` will show whichever model ticked last rather than a process total or a per-model value.

Please either add a stable model label to these inference metrics, or emit the gauges from a process-level aggregate so their semantics are correct. In addition, maybe this is something we need to fix/integrate for other prometheus logging (e.g in the server)?

MicheleCampi:
Thanks for the careful review — you're right, and it's a real correctness issue.
I traced where the model id is available: create_engine_instance builds the IntervalLogger before the engine id is resolved (it comes from pipeline.name() / the override afterwards), so the id isn't in scope at the logger's construction point. Labelling the gauges per-model from there would mean threading the id through create_engine_instance and its call sites — and it would still be the wrong layer.
So I've narrowed this PR to the metric that's already correct process-wide: inference_tokens_processed_total is additive across engines, so a single counter stays well-defined in a multi-model server. This should be ready to merge as-is.
For requests_running / requests_waiting, I'll do them properly in a follow-up PR at the server layer, where the model id is known — per-model gauges from the server-side registry, alongside the existing Prometheus logging there. That also matches the broader point you raised about integrating this in the server. Happy to open it right after this one lands.

EricLBuehler:
Hey @MicheleCampi! We merged alternative Prometheus support that leverages some of your earlier work too :)

PR #2250, opened 2026-06-19
Fix reversed FCFS priority in PagedAttentionScheduler preemption

### LINKED ISSUE #2033: PagedAttentionScheduler priority sorting causes First-Come-Last-Served behavior
`PagedAttentionScheduler` priority sorting results in First-Come-Last-Served scheduling under load**

The `PagedAttentionScheduler` in `mistralrs-core/src/paged_attention/scheduler.rs` appears to contain a priority inversion in its sort logic. The function `sort_running_by_priority_fcfs` currently sorts sequences in reverse order of arrival, which leads to starvation of older requests under memory pressure.

```rust
    fn sort_running_by_priority_fcfs(&mut self) {
        self.running
            .make_contiguous()
            .sort_by_key(|seq| get_mut_arcmutex!(seq).timestamp());
        self.running.make_contiguous().reverse(); // <-- issue here
    }
```

Since `seq.timestamp()` records the Unix epoch time of creation, `sort_by_key` inherently orders the queue ascending: `[Oldest, ..., Newest]`. 
The `.reverse()` call then flips this to `[Newest, ..., Oldest]`.

In the decode loop, `self.running.pop_front()` is used to select the next sequence for token generation. Because the array is reversed, `pop_front()` selects the **Newest** sequence first.
Conversely, when the KV cache exhausts available blocks during allocation, the scheduler triggers preemption by calling `self.running.pop_back()`, which preempts the **Oldest** sequence.

**Impact:**
Instead of a First-Come-First-Serve (FCFS) queue, this creates a First-Come-Last-Served behavior. Under high concurrency, the engine will prioritize token generation for new requests while preempting the oldest requests. This can lead to starvation of long-context queries and increased recomputation overhead.

For comparison, the `DefaultScheduler` handles FCFS by using an ascending sort (`slice.sort_by_key(|seq| *seq.id());`) without a subsequent reverse. Removing the `.reverse()` call in `PagedAttentionScheduler` or swapping the `pop_front`/`pop_back` semantics should resolve the issue.

pjdurden:
Fixes #2033.

sort_running_by_priority_fcfs sorts self.running ascending by timestamp (oldest
first) and then reverses it to newest first. The decode loop schedules with
pop_front and, under KV-cache pressure, preempts with pop_back. With the reverse
in place the oldest (highest priority) sequences get preempted first and the
newest get scheduled first, the opposite of FCFS and of the function's own name.

Dropping the reverse keeps running oldest-first: pop_front schedules the oldest
sequence first and pop_back preempts the newest (lowest priority) first. This
matches DefaultScheduler, which sorts ascending without reversing.

### DIFF
--- a/mistralrs-core/src/paged_attention/scheduler.rs
+++ b/mistralrs-core/src/paged_attention/scheduler.rs
@@ -589,7 +589,6 @@ impl PagedAttentionScheduler {
         self.running
             .make_contiguous()
             .sort_by_key(|seq| get_mut_arcmutex!(seq).timestamp());
-        self.running.make_contiguous().reverse();
     }
 }
 


pjdurden:
@EricLBuehler the failing Check jobs arent from this PR, its just a crates.io flake fetching tendril (html2text dep) durin cargo check, the actual Test Suites all pass green so the fix is fine, just needs a re-run whenver you get a sec

EricLBuehler (changes requested):
Hey @pjdurden - thanks for the PR. Can you please remove Claude from your commit authorship?

pjdurden:
> Hey @pjdurden - thanks for the PR. Can you please remove Claude from your commit authorship?

Done, removed Claude from the commit authorship. Thanks for the review!

EricLBuehler (approved):
LGTM!

PR #2268, opened 2026-06-21
feat: add Hunyuan v1 dense and MoE support

ASheng1019:
Hi! This PR adds Hunyuan v1 dense and MoE support, including automatic loading, paged attention, tensor parallelism, ISQ/MoQE, and Python SDK registration.

One part I’d particularly appreciate feedback on is the expert-loading helper.

Hunyuan stores routed experts as separate per-expert gate, up, and down projections. Stacking all three before immediate ISQ creates a large temporary memory peak. I added `ExpertCheckpoint::stacked_proj` in `mistralrs-core/src/moe/experts/checkpoint.rs`, allowing `FastExpertsWeights::load_unquantized` to load and quantize each projection sequentially.

The helper produces the existing canonical `[expert, out, in]` layout and reuses the current checkpoint layout detection and tensor-parallel sharding logic. It supports separate per-expert weights as well as canonical and transposed fused checkpoints.

Does this seem like the right level of abstraction for `ExpertCheckpoint`? I kept it generic because sequential loading may also help other large per-expert checkpoints, but I’d be happy to make it backend-local or reshape it around an iterator/callback if that would fit the project better.

The PR also adds a CUDA capacity-mask operation for Hunyuan’s priority-ordered top-k routing, avoiding a GPU-to-CPU synchronization.

### Local CUDA results

Tested on an NVIDIA GB10 with CUDA 13.0:

| Model | Configuration | Prefill 128 | Prefill 512 | Decode d128 | Decode d512 |
|---|---|---:|---:|---:|---:|
| Hunyuan-1.8B | BF16 | 149.6 tok/s | 598.4 tok/s | 67.7 tok/s | 67.3 tok/s |
| Hunyuan-A13B | q4k ISQ + MoQE | 140.0 tok/s | 491.9 tok/s | 24.2 tok/s | 23.6 tok/s |

The 154 GB A13B checkpoint was successfully loaded and quantized within the reported 92 GB device budget. Initial q4k quantization and loading took about 6 minutes 43 seconds.

Thanks! I’d be glad to adjust the helper or integration based on your preferences.


EricLBuehler (changes requested):
Hey @ASheng1019! Awesome work with this PR 🙌.

I took a look at it and it looks great. I'll do some testing, but the only thing I think we need before merging this is to be able to parse the tool call format.

Ideally this can be fully-featured to match other models: making sure parsing works, as well as the constrained grammar tool calling. For the constrained/required-grammar path to select the right format, you'll also need to register the format, for example in `template_tool_call_format` as well as the `PARSERS` registry. To start, please see: https://github.com/EricLBuehler/mistral.rs/tree/master/mistralrs-core/src/tools/parsers

Users will need to provide the tool-call jinja chat template for this model manually, unless we implement the vllm-style built-in chat template for tool calling, and I think the usage is fine here.

ASheng1019:
Thanks! I’ve added Hunyuan tool-call parsing and constrained grammar support.

The implementation registers `ToolCallFormat::Hunyuan` in the parser registry and chat-template format detection, and supports the `<tool_calls>[...]</tool_calls>` JSON-array format, including parallel tool calls and required/strict grammar generation via the shared JSON grammar builder.

I also fixed an A13B tokenizer edge case where `>[` can be emitted as a single token, which could otherwise duplicate the JSON array opener during constrained auto tool calling.

Locally validated with Hunyuan-A13B q4k/MoQE using the vLLM A13B tool-call template:
- `tool_choice: auto` returns `finish_reason: tool_calls`
- `tool_choice: required` with strict schema also returns a valid tool call

Please let me know if this parser/helper shape looks aligned with the project’s preferred tool-calling architecture.

EricLBuehler (approved):
Looks good! Additionally, the improvements to ISQ are awesome.

ASheng1019:
Thank you for your affirmation. It gives me more motivation😄.

EricLBuehler:
Of course, nice work with this! Happy to review any futher PRs!

PR #2279, opened 2026-06-23
Fix agentic loop not starting for registered tool callbacks

### LINKED ISSUE #2278: Tool callbacks do not trigger `agentic_loop` unless tools are added to `RequestBuilder`
When I register a new tool through 'with_tool_callback_and_tool', the model does not receive it (and therefore cannot use it). However, if I specify a tool via `set_tools` in `RequestBuilder`, then everything works correctly and the model recognizes the tools. Event more discouraging, if I specify both the callback and the list in set_tools at the same time, I get an error: `Tool 'my_tool' conflicts with a registered internal tool. Internal tool names cannot be overridden.`

```Rust
#[tool(description = "This is my tool")]
fn my_tool() -> anyhow::Result<String> {
    todo!()
}

let model = GgufModelBuilder::new("/Users/aeiklorvy/ai_bot/", vec!["Qwen3-4B-Q4_K_M.gguf"])
    .with_logging()
    // NOT WORKING WITHOUT set_tools
    .with_tool_callback_and_tool(
        "my_tool",
        my_tool_callback(),
        my_tool_tool(),
    )
    .build()
    .await?;

let request = RequestBuilder::new()
    .set_tool_choice(ToolChoice::Auto)
    // HERE ERROR: conflicts with a registered internal tool
    .set_tools(vec![ my_tool_tool() ]);

let response = model.send_chat_request(request).await?;
println!("{response:?}");
```

I looked at the source codes and found a bug in `mistralrs_core::engine::add_request::Engine::handle_request`:
```Rust
let has_tooling = !self.tool_callbacks.is_empty()
    && request.tools.as_ref().is_some_and(|t| !t.is_empty());
```

This looks like a logical error. To run the `agentic_loop` function, I need to specify both callbacks in model and tools in `RequestBuilder`. But if I do that, I will get an error (Tool 'my_tool' conflicts...).
I assume that instead of AND there should be OR:
```Rust
let has_tooling = !self.tool_callbacks.is_empty()
    || request.tools.as_ref().is_some_and(|t| !t.is_empty());
```

In my case, after these changes, everything worked correctly. However, I am very superficially immersed in this project and may not take something into account (maybe it was done on purpose?).

I used the latest commit from the master branch:
mistralrs = { git = "https://github.com/EricLBuehler/mistral.rs", rev = "3ee69a7" }

aeiklorvy:
Fixes #2278

Tool callbacks registered via `with_tool_callback_and_tool()` did not trigger `agentic_loop` unless tools were also manually added to the request.

### DIFF
--- a/mistralrs-core/src/engine/add_request.rs
+++ b/mistralrs-core/src/engine/add_request.rs
@@ -36,7 +36,7 @@ impl Engine {
                 let in_agentic_loop =
                     request.max_tool_rounds == agentic_loop::AGENTIC_LOOP_REENTRY_SENTINEL;
                 let has_tooling = !self.tool_callbacks.is_empty()
-                    && request.tools.as_ref().is_some_and(|t| !t.is_empty());
+                    || request.tools.as_ref().is_some_and(|t| !t.is_empty());
                 let has_search = request.web_search_options.is_some();
                 let has_code_exec =
                     request.enable_code_execution && !self.tool_callbacks.is_empty();
--- a/mistralrs-core/src/engine/add_request.rs
+++ b/mistralrs-core/src/engine/add_request.rs
@@ -43,9 +43,6 @@ impl Engine {
                     )
                 });
                 let has_search = request.web_search_options.is_some();
-                let has_code_exec =
-                    request.enable_code_execution && !self.tool_callbacks.is_empty();
-                let has_shell = request.enable_shell && !self.tool_callbacks.is_empty();
                 let has_agentic =
                     request.max_tool_rounds.is_some() || request.tool_dispatch_url.is_some();
                 let has_input_files =
@@ -53,12 +50,7 @@ impl Engine {
 
                 if is_chat
                     && !in_agentic_loop
-                    && (has_search
-                        || has_tooling
-                        || has_code_exec
-                        || has_shell
-                        || has_agentic
-                        || has_input_files)
+                    && (has_search || has_tooling || has_agentic || has_input_files)
                 {
                     agentic_loop::agentic_loop(self.clone(), *request).await;
                 } else if request.files.as_ref().is_some_and(|f| !f.is_empty()) {
--- a/mistralrs-core/src/engine/add_request.rs
+++ b/mistralrs-core/src/engine/add_request.rs
@@ -35,8 +35,13 @@ impl Engine {
                 );
                 let in_agentic_loop =
                     request.max_tool_rounds == agentic_loop::AGENTIC_LOOP_REENTRY_SENTINEL;
-                let has_tooling = !self.tool_callbacks.is_empty()
-                    || request.tools.as_ref().is_some_and(|t| !t.is_empty());
+                let has_tooling = self.tool_callbacks.keys().any(|name| {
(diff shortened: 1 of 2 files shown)


EricLBuehler (changes requested):
Hey @aeiklorvy! Thanks for the PR and for tracking this down. The existing `&&` is definitely the reason `with_tool_callback_and_tool()` does not start the loop.

I think the fix needs to be slightly narrower than `|| request.tools...`, though. `request.tools` alone is also the normal client-side tool-calling path, where the model should return tool calls to the caller instead of the engine trying to execute them. Also, `self.tool_callbacks` includes gated built-ins like code execution and shell, so a plain chat request on a code-exec-enabled engine could enter `agentic_loop` even when that request did not enable code execution.

Can you gate on registered tools that are actually active for this request?

In `agentic_loop.rs`, something like:

```rust
pub(super) fn registered_tool_active_for_request(
    name: &str,
    enable_code_execution: bool,
    enable_shell: bool,
) -> bool {
    if is_shell_tool(name) {
        enable_shell
    } else if is_code_exec_tool(name) {
        enable_code_execution
    } else {
        true
    }
}
```

Then in add_request.rs, something like the following to hande all the cases:

```
let has_tooling = self.tool_callbacks.keys().any(|name| {
    agentic_loop::registered_tool_active_for_request(
        name,
        request.enable_code_execution,
        request.enable_shell,
    )
});
let has_search = request.web_search_options.is_some();
let has_agentic = request.max_tool_rounds.is_some() || request.tool_dispatch_url.is_some();
let has_input_files =
    cfg!(feature = "code-execution") && is_chat && !request.input_files.is_empty();
```

And the loop condition can stay:

```rust
if is_chat
    && !in_agentic_loop
    && (has_search || has_tooling || has_agentic || has_input_files)
{
    agentic_loop::agentic_loop(self.clone(), *request).await;
}
```

This should fix registered custom/MCP tools without changing ordinary client-managed tool schemas into server-executed agentic requests.

aeiklorvy:
Thanks for the explanation! I have made changes according to your advice. However, I only checked the correctness of my case described in Issue #2278, so I would appreciate if I looked again to make sure that I didn't miss anything.

mistralrs-core/src/engine/add_request.rs
EricLBuehler: One small issue before merge: `has_code_exec` and `has_shell` still use the broad `!self.tool_callbacks.is_empty()` check, so they can start `agentic_loop` even when no matching
active registered tool exists for this request.

Since `has_tooling` now checks active registered tools, can we remove `has_code_exec` and `has_shell` from this condition and leave the trigger as `has_search || has_tooling ||
has_agentic || has_input_files`?

EricLBuehler (changes requested):
Thanks for the changes! Looks basically good. Just one small comment

aeiklorvy:
Indeed, the separate checks `has_code_exec` and `has_shell` have now become redundant, because `has_tooling` already accounts for these parameters by calling `registered_tool_active_for_request`.

EricLBuehler (approved):
Looks great!

PR #2286, opened 2026-06-25
chore(mistralrs-core): cap tojson indent and remove chat template panic paths

eharris128:
## What

Hardening for the Jinja chat template rendering path in `mistralrs-core`
(`src/pipeline/chat_template.rs`). Three changes, all confined to that one file:

1. Cap the `tojson` filter's `indent` argument at 16. The template-supplied
   `indent` flowed straight into `b" ".repeat(indent)`, so a value near
   `usize::MAX` would allocate gigabytes or panic with "capacity overflow".
   Legitimate pretty-printing never needs more than a few spaces.
2. Propagate the serializer error in `tojson` with `?` instead of `.unwrap()`,
   so a serialization failure returns a template error rather than panicking.
3. Propagate the `get_template("chat_template")` error with `?` instead of
   `.unwrap()`.

## Why

A model's chat template (from `tokenizer_config.json`, `chat_template.jinja`,
or GGUF metadata) is rendered with no panic guards. If an operator loads a
malformed or untrusted model, the first chat request can crash or wedge the
engine task instead of returning a clean error. These are availability
hardening changes only; rendering output for well-formed templates is
unchanged.

This follows up on a privately reported issue that the maintainers triaged as
outside the server's security boundary (a remote client cannot supply the
template), with an explicit invitation to send hardening that caps rendering
cost or removes panic paths. This PR does the panic-path removal half.

It intentionally does not add an execution-fuel cap for unbounded loops or
allocations in template logic; that would mean enabling minijinja's `fuel`
feature and choosing a budget, which is better discussed separately.

## Validation

The exact changed logic (a verbatim copy of `tojson` plus the rebuilt
`Environment`) was compiled and run against minijinja 2.14.0:

- `tojson(indent=18446744073709551615)` now returns `Ok` with a clamped indent
  instead of panicking with "capacity overflow".
- A malformed template surfaces as a graceful `Err` instead of panicking.
- `tojson(indent=2)` still produces normal pretty-printed output.

`cargo fmt -p mistralrs-core -- --check` passes.

### DIFF
--- a/mistralrs-core/src/pipeline/chat_template.rs
+++ b/mistralrs-core/src/pipeline/chat_template.rs
@@ -287,13 +287,14 @@ impl GenerationConfig {
 
 fn tojson(value: Value, kwargs: Kwargs) -> Result<Value, Error> {
     if let Ok(indent) = kwargs.get::<usize>("indent") {
-        // Cap the indent width. A template-supplied `indent` flows straight into
-        // `b" ".repeat(indent)`; an attacker-controlled chat template could pass a
-        // huge value (up to `usize::MAX`), allocating gigabytes or panicking with a
-        // "capacity overflow". Legitimate pretty-printing never needs more than a
-        // few spaces, so clamp it.
-        const MAX_INDENT: usize = 16;
-        let indent = indent.min(MAX_INDENT);
+        // Cap the indent: it feeds `b" ".repeat(indent)`, so an attacker-controlled template could request a huge allocation or capacity-overflow panic.
+        const MAX_INDENT: usize = 256;
+        if indent > MAX_INDENT {
+            return Err(Error::new(
+                ErrorKind::InvalidOperation,
+                format!("tojson `indent` of {indent} exceeds the maximum of {MAX_INDENT}"),
+            ));
+        }
         let mut buf = Vec::new();
         let repeat = b" ".repeat(indent);
         let formatter = serde_json::ser::PrettyFormatter::with_indent(&repeat);
--- a/mistralrs-core/src/pipeline/chat_template.rs
+++ b/mistralrs-core/src/pipeline/chat_template.rs
@@ -286,12 +286,21 @@ impl GenerationConfig {
 }
 
 fn tojson(value: Value, kwargs: Kwargs) -> Result<Value, Error> {
-    if let Ok(indent) = kwargs.get("indent") {
+    if let Ok(indent) = kwargs.get::<usize>("indent") {
+        // Cap the indent width. A template-supplied `indent` flows straight into
+        // `b" ".repeat(indent)`; an attacker-controlled chat template could pass a
+        // huge value (up to `usize::MAX`), allocating gigabytes or panicking with a
+        // "capacity overflow". Legitimate pretty-printing never needs more than a
+        // few spaces, so clamp it.
+        const MAX_INDENT: usize = 16;
+        let indent = indent.min(MAX_INDENT);
         let mut buf = Vec::new();
         let repeat = b" ".repeat(indent);
         let formatter = serde_json::ser::PrettyFormatter::with_indent(&repeat);
         let mut ser = serde_json::Serializer::with_formatter(&mut buf, formatter);
-        value.serialize(&mut ser).unwrap();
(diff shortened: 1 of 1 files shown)


mistralrs-core/src/pipeline/chat_template.rs
```diff
@@ -286,12 +286,21 @@ impl GenerationConfig {
...
 
 fn tojson(value: Value, kwargs: Kwargs) -> Result<Value, Error> {
-    if let Ok(indent) = kwargs.get("indent") {
+    if let Ok(indent) = kwargs.get::<usize>("indent") {
+        // Cap the indent width. A template-supplied `indent` flows straight into
+        // `b" ".repeat(indent)`; an attacker-controlled chat template could pass a
+        // huge value (up to `usize::MAX`), allocating gigabytes or panicking with a
+        // "capacity overflow". Legitimate pretty-printing never needs more than a
+        // few spaces, so clamp it.
+        const MAX_INDENT: usize = 16;
```
EricLBuehler: Can we make this 256 to give a bit more headroom? Ideally higher is better here I think

mistralrs-core/src/pipeline/chat_template.rs
```diff
@@ -286,12 +286,21 @@ impl GenerationConfig {
 }
 
 fn tojson(value: Value, kwargs: Kwargs) -> Result<Value, Error> {
-    if let Ok(indent) = kwargs.get("indent") {
+    if let Ok(indent) = kwargs.get::<usize>("indent") {
+        // Cap the indent width. A template-supplied `indent` flows straight into
```
EricLBuehler: Can we make this comment follow existing comment styles (eg the wrapping seen here is not ideal).

For example:
> https://github.com/EricLBuehler/mistral.rs/blob/master/mistralrs-core/src/files/mod.rs#L21

mistralrs-core/src/pipeline/chat_template.rs
```diff
@@ -286,12 +286,21 @@ impl GenerationConfig {
...
+        // `b" ".repeat(indent)`; an attacker-controlled chat template could pass a
+        // huge value (up to `usize::MAX`), allocating gigabytes or panicking with a
+        // "capacity overflow". Legitimate pretty-printing never needs more than a
+        // few spaces, so clamp it.
+        const MAX_INDENT: usize = 16;
+        let indent = indent.min(MAX_INDENT);
```
EricLBuehler: Instead of silently capping it, can we throw an error if greater than MAX_INDENT? It does impose a limit but I think 256 is reasonable.

EricLBuehler (changes requested):
Thanks for the PR @eharris128! A few comments

mistralrs-core/src/pipeline/chat_template.rs
eharris128: (yes sorry)

mistralrs-core/src/pipeline/chat_template.rs
eharris128: Yes this is wise.

mistralrs-core/src/pipeline/chat_template.rs
eharris128: Done.

eharris128:
> Thanks for the PR @eharris128! A few comments

Thank you for the comments :)

mistralrs-core/src/pipeline/chat_template.rs
EricLBuehler: No worries!

EricLBuehler (changes requested):
Thanks for the changes, they look great. Can you please amend the commits to exclude credit for claude?

EricLBuehler (changes requested):
Thanks for the changes, they look great. Can you please amend the commits to exclude credit for claude?

eharris128:
Updated:
image

Although the pull request review view still shows the stale commit authorship.

The commits themselves have only my GitHub account as the author.

Let me know if you would like me to adjust anything else.

EricLBuehler (approved):
👍 Thanks for the PR!

PR #2294, opened 2026-06-27
fix(core): stream raw chat deltas without tool state

ASheng1019:
@EricLBuehler hi! This fixes a small OpenAI-compatible chat streaming issue I ran into while testing the server.

## Summary

Ordinary `/v1/chat/completions` streaming responses could emit chunks where `choices[].delta.content` stayed `null`, even though the model generated text and the equivalent non-streaming request returned `choices[].message.content` correctly.

The issue was in chat streaming chunk assembly: when there was no reasoning parser and no tool-call state, the raw decoded token delta from `seq.get_delta()` was not copied into `delta.content`.

This patch falls back to the raw delta only when no parser/tool state owns the stream, while preserving the existing reasoning/tool-call suppression behavior.

## What Changed

- Promoted the existing streaming text/tool-call helper into the production path.
- For ordinary chat streaming with no tool state, use the raw token delta as `delta.content`.
- If an external reasoning parser is active, do not fall back to raw delta.
- If tool-call state exists, keep delegating to `ToolCallState::parse_streaming`.
- Added regression coverage for:
  - no tool state uses raw delta
  - external reasoning does not leak raw delta
  - existing tool-call streaming behavior

## Verification

```bash
cargo test -p mistralrs-core pipeline::sampling::tests
```

I also verified the behavior end-to-end with a local OpenAI-compatible server.

<details>
<summary>End-to-end reproduction script and output</summary>

Model used:

```text
Qwen/Qwen2.5-0.5B-Instruct
```

Start the server:

```bash
target/debug/mistralrs serve \
  --host 127.0.0.1 \
  --port 8123 \
  --no-ui \
  --cpu \
  --max-seq-len 512 \
  --max-batch-size 1 \
  -m Qwen/Qwen2.5-0.5B-Instruct
```

Run this script with the server base URL:

```bash
python3 repro_chat_streaming_content.py http://127.0.0.1:8123
```

Script:

```python
#!/usr/bin/env python3
import json
import sys
import time
import urllib.request


BASE_URL = sys.argv[1] if len(sys.argv) > 1 else "http://127.0.0.1:1234"


def post_json(path, payload, timeout=120):
    data = json.dumps(payload).encode("utf-8")
    req = urllib.request.Request(
        BASE_URL + path,
        data=data,
        headers={"Content-Type": "application/json"},
        method="POST",
    )
    return urllib.request.urlopen(req, timeout=timeout)


def wait_for_server(timeout=180):
    deadline = time.time() + timeout
    while time.time() < deadline:
        try:
            urllib.request.urlopen(BASE_URL + "/v1/models", timeout=2).read()
            return
        except Exception:
            time.sleep(1)
    raise RuntimeError(f"server did not become ready at {BASE_URL}")


def main():
    wait_for_server()

    common = {
        "model": "default",
        "messages": [
            {
                "role": "user",
                "content": "Reply with exactly this word and nothing else: banana",
            }
        ],
        "max_tokens": 16,
        "temperature": 0.0,
    }

    non_stream = json.load(post_json("/v1/chat/completions", {**common, "stream": False}))
    message_content = non_stream["choices"][0]["message"].get("content")
    print("NON_STREAM_CONTENT:", repr(message_content))
    if not message_content:
        raise SystemExit("non-streaming response had empty message.content")

    content_chunks = []
    null_chunks = 0
    total_chunks = 0
    finish_reasons = []

    with post_json("/v1/chat/completions", {**common, "stream": True}) as resp:
        for raw_line in resp:
            line = raw_line.decode("utf-8").strip()
            if not line.startswith("data:"):
                continue
            data = line[len("data:") :].strip()
            if data == "[DONE]":
                break
            event = json.loads(data)
            for choice in event.get("choices", []):
                total_chunks += 1
                finish_reasons.append(choice.get("finish_reason"))
                delta = choice.get("delta", {})
                content = delta.get("content")
                if content is None:
                    null_chunks += 1
                elif content:
                    content_chunks.append(content)

    stream_text = "".join(content_chunks)
    print("STREAM_TOTAL_CHUNKS:", total_chunks)
    print("STREAM_NULL_CONTENT_CHUNKS:", null_chunks)
    print("STREAM_NONEMPTY_CONTENT_CHUNKS:", len(content_chunks))
    print("STREAM_JOINED_CONTENT:", repr(stream_text))
    print("STREAM_FINISH_REASONS:", finish_reasons)

    if not content_chunks:
        raise SystemExit(
            "streaming response produced no non-empty choices[].delta.content chunks"
        )


if __name__ == "__main__":
    main()
```

On current `master`, non-streaming returns content, but streaming produces no non-empty `choices[].delta.content` chunks:

```text
NON_STREAM_CONTENT: 'banana'
STREAM_TOTAL_CHUNKS: 2
STREAM_NULL_CONTENT_CHUNKS: 2
STREAM_NONEMPTY_CONTENT_CHUNKS: 0
STREAM_JOINED_CONTENT: ''
STREAM_FINISH_REASONS: [None, 'stop']
streaming response produced no non-empty choices[].delta.content chunks
```

After this patch, the same request streams content as expected:

```text
NON_STREAM_CONTENT: 'banana'
STREAM_TOTAL_CHUNKS: 2
STREAM_NULL_CONTENT_CHUNKS: 0
STREAM_NONEMPTY_CONTENT_CHUNKS: 1
STREAM_JOINED_CONTENT: 'banana'
STREAM_FINISH_REASONS: [None, 'stop']
```

</details>

### DIFF
--- a/mistralrs-core/src/pipeline/sampling.rs
+++ b/mistralrs-core/src/pipeline/sampling.rs
@@ -36,21 +36,34 @@ fn parse_text_and_tool_calls(
     Ok((parsed.content, parsed.tool_calls))
 }
 
-#[cfg(test)]
 fn parse_streaming_text_and_tool_calls(
     content_delta: Option<String>,
     raw_delta: &str,
-    has_reasoning_parser: bool,
+    has_external_reasoning_parser: bool,
+    is_done: bool,
     state: Option<&mut ToolCallState>,
-) -> Result<(Option<String>, Vec<crate::tools::ToolCallResponse>)> {
+) -> Result<crate::tools::state::ToolCallParse> {
     let Some(state) = state else {
-        return Ok((
-            content_delta.or_else(|| Some(raw_delta.to_string())),
-            Vec::new(),
-        ));
+        return Ok(crate::tools::state::ToolCallParse {
+            content: content_delta.or_else(|| {
+                if has_external_reasoning_parser {
+                    None
+                } else {
+                    Some(raw_delta.to_string())
+                }
+            }),
+            reasoning_content: None,
+            tool_calls: Vec::new(),
+            tool_use_still_possible: false,
+            tool_use_is_done: false,
+        });
     };
-    let parsed = state.parse_streaming(content_delta, raw_delta, has_reasoning_parser, false)?;
-    Ok((parsed.content, parsed.tool_calls))
+    state.parse_streaming(
+        content_delta,
+        raw_delta,
+        has_external_reasoning_parser,
+        is_done,
+    )
 }
 
 fn activate_required_tool_call_grammar(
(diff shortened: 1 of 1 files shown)


ASheng1019:
A bit more context from my side: I first ran into this while porting the Hunyuan VL architecture, where the OCR path was producing valid non-streaming responses but streaming chat chunks had `delta.content = null`.

To make sure it was not specific to Hunyuan VL, I also reproduced it with `Qwen/Qwen2.5-0.5B-Instruct` using a plain chat QA request with no tools or reasoning options. That showed the same pattern: non-streaming returned `message.content`, while streaming emitted chunks without usable `delta.content`.

After this patch, both cases look good locally:

- Hunyuan VL OCR streams user-visible text correctly.
- Qwen2.5 chat QA streams `delta.content` correctly.

So this seems to be a general chat streaming response assembly issue rather than something model-specific.

PR #2338, opened 2026-07-11
fix(webui): send chat images as data URLs (fix HTTP 500 on image send)

### LINKED ISSUE #2336: Web UI: sending an image returns HTTP 500 `invalid image source` (v1 API works)
## Describe the bug

Sending an image from the built-in web UI (`http://localhost:1234/ui`) fails with HTTP 500:

```
invalid image source, expected http https or data url: relative url without a base
```

Sending the same image directly through the `/v1/chat/completions` API works fine.

## Root cause

The UI uploads the picked image to `/api/upload_image`, which saves it to the cache dir and returns a **relative** URL (`mistralrs-cli/src/ui/handlers/api.rs`):

```rust
let url = format!("uploads/{filename}");
(StatusCode::OK, Json(json!({ "path": path, "url": url }))).into_response()
```

The UI stores that relative string in the message and sends it back as `image_url.url` to `/v1/chat/completions` (`mistralrs-cli/webui/src/lib/stores/chat.svelte.ts` -> `image_url: { url }`). That endpoint resolves image sources under the `ServerRequest` media policy, which only accepts `http` / `https` / `data` URLs (`mistralrs-server-core/src/media_source.rs`):

```rust
anyhow::bail!("Invalid {kind} source: expected http, https, or data URL: {err}");
```

`Url::parse("uploads/<uuid>.png")` fails with `RelativeUrlWithoutBase` ("relative url without a base"), producing the 500. The v1 API works because real callers pass a `data:`/`http(s)` URL. (The relative URL only works for display, via the `/uploads` static route.)

The upload flow was added in #1370 / #1364.

## Steps to reproduce

1. `mistralrs serve -m <any vision model>` (UI is on by default at `/ui`)
2. Open `http://localhost:1234/ui`, attach an image, send.
3. -> HTTP 500 `invalid image source ... relative url without a base`.
4. The same image via `POST /v1/chat/completions` with a `data:`/`http` URL succeeds.

## Other information

- OS: Linux (WSL2); reproduces regardless of backend (it's a request-layer/URL bug).

## Latest commit or version

v0.8.23 / current `master`.

---

I have a fix ready (encode UI images as `data:` URLs client-side, matching what the working v1 path sends) and will open a PR referencing this issue. Note videos share the identical relative-URL path and are fixed the same way.

subin9:
Sending an image from the built-in web UI (`/ui`) failed with HTTP 500 `invalid image source, expected http https or data url: relative url without a base`, while the same image through `/v1/chat/completions` worked. Closes #2336.

### Cause

The UI uploaded the picked image to `/api/upload_image`, which saves it under the cache dir and returns a **relative** URL (`uploads/<uuid>.png`). The UI stored that string and sent it back as `image_url.url` to `/v1/chat/completions`, which resolves image sources under the `ServerRequest` media policy - accepting only `http` / `https` / `data` URLs. `Url::parse("uploads/<uuid>.png")` fails with `RelativeUrlWithoutBase`, producing the 500. The v1 API worked because callers pass a `data:`/`http(s)` URL.

### Fix

Encode picked files as `data:` URLs client-side (via `FileReader.readAsDataURL`) and send those - exactly what the working v1 path sends - instead of round-tripping through the upload endpoint. This also fixes the identical relative-URL failure for videos, and works for display-on-reload since a data URL is self-contained.

### DIFF
--- a/mistralrs-cli/webui/src/lib/components/ChatInput.svelte
+++ b/mistralrs-cli/webui/src/lib/components/ChatInput.svelte
@@ -9,6 +9,15 @@
   let pendingVideos = $state<{ file: File; url: string; uploadedUrl?: string }[]>([]);
   let isDragging = $state(false);
 
+  function fileToDataUrl(file: File): Promise<string> {
+    return new Promise((resolve, reject) => {
+      const reader = new FileReader();
+      reader.onload = () => resolve(reader.result as string);
+      reader.onerror = () => reject(reader.error);
+      reader.readAsDataURL(file);
+    });
+  }
+
   function autoResize() {
     if (!textareaEl) return;
     textareaEl.style.height = "auto";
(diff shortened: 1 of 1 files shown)


EricLBuehler (approved):
Thank you @subin9!

PR #2339, opened 2026-07-11
Fix batched multimodal prefill rank mismatch in Gemma 4 bidirectional mask

### LINKED ISSUE #2323: Batched multimodal prefill crashes with rank mismatch; mixed-length image batches silently corrupt output (Gemma 4)
## Environment
- mistral.rs v0.8.23, Gemma-4-26B-A4B, Q6K UQFF, PagedAttention on
- DGX Spark (GB10, CUDA 13.0), Python bindings in-process
- The crash is a shape/logic error, so it should be hardware-independent
- Note: this is with the scheduler livelock fixed locally (see companion issue) — without that fix, batched multimodal never gets far enough to hit this

## Part A: deterministic crash with identical requests
Submitting **two identical vision requests** (same image, same prompt, same length -> same scheduler bucket -> both image prefills land in one forward) fails both requests:

```
ERROR mistralrs_core::engine: step - Model failed with error: unexpected rank, expected: 1, got: 2 ([2, 1080])
ValueError: unexpected rank, expected: 1, got: 2 ([2, 1080])
```

Something in the vision-embedding path assumes a rank-1 (single-sequence) tensor and receives the batched `[batch=2, n]` version.

Minimal repro (Python bindings; any image, we used a rendered A1 drawing page):

```python
import base64, threading
from pathlib import Path
from mistralrs import Runner, Which, MultimodalArchitecture, ChatCompletionRequest

b64 = base64.b64encode(Path("page.png").read_bytes()).decode()
messages = [{"role": "user", "content": [
    {"type": "image_url", "image_url": {"url": f"data:image/png;base64,{b64}"}},
    {"type": "text", "text": "Describe this drawing. Transcribe all text and numbers."},
]}]

runner = Runner(which=Which.MultimodalPlain(model_id="<gemma4-26b-dir>", arch=MultimodalArchitecture.Gemma4))
def go():
    print(runner.send_chat_completion_request(ChatCompletionRequest(
        model="x", messages=messages, max_tokens=512, temperature=0.0)))
ts = [threading.Thread(target=go) for _ in range(2)]
[t.start() for t in ts]; [t.join() for t in ts]
```

## Part B: silent corruption with mixed-length batches
Real-world concurrent requests have *different* prompt lengths, land in different length buckets, and get their image prefills serialized — so they dodge the Part A crash. But the outputs still degrade with batch width (greedy decoding, temp 0, same 5-page document each time):

| width | page-read outputs |
|---|---|
| 1 | natural EOS at 444-807 tokens; downstream review matches our validated baseline |
| 2 | every output runs to the max_tokens cap (1024), no natural stop; downstream review quality collapses |
| 5 | two requests emit EOS after 6-7 tokens (effectively empty), one truncates at 255; the model's own summary of the extractions: "nonsensical text" |

No error is raised in Part B — requests "succeed" with corrupted content, which is worse than the crash.

## Context
RFC #2171 deferred PagedAttention interaction / KV heterogeneity / multimodal batching for Gemma 4, so some of this may be known territory. Filing because the failure modes are now cleanly separable: (companion issues) a scheduler livelock and a hot-path perf bug, plus this correctness bug, which is the remaining blocker for batched multimodal on this model. Happy to run diagnostics/patches on our hardware — we have a fast repro loop.

gmhelmold:
Fixes #2323 (Part A)

## Root cause
`Gemma4TextModel::apply_image_bidirectional_mask` assumed batch size 1: it did `input_ids.squeeze(0)?` followed by `to_vec1()`, which errors with `unexpected rank, expected: 1, got: 2 ([2, 1080])` the moment two same-length requests are batched into a single prefill (`input_ids` is `[bsz, seq]`). This path is always taken on cuda-only / no-flash-attn builds. The bidirectional vision-group override was also built as one shared 2-D mask, so even past the crash it couldn't represent per-row image positions.

## What changed
Compute the vision-group override per row against the shared 2-D causal mask, then stack the rows into `[bsz, 1, seq, total]`. `bsz == 1` returns the original `[seq, total]` tensor unchanged, so the single-request path is byte-identical (same dims, dtype, values). `[bsz, 1, seq, total]` broadcasts to `[bsz, n_heads, seq, total]` at every consumer of this mask — naive SDPA (`broadcast_add`), the cuBLASLt rank-4 branch, Metal SDPA (`broadcast_as`), and the CPU flash-attn mask indexer — for both the full and sliding-window masks.

## Tests
Three CPU tests added in `vision_models/gemma4/text.rs`:
- batched rows equal the independently-computed per-row masks,
- `bsz == 1` keeps the old `[seq, total]` shape and exact values,
- mixed image+video, image-only and text-only rows group correctly (text-only keeps the causal mask untouched).

`cargo test -p mistralrs-core --lib gemma4` → 47 passed; 0 failed.
`cargo fmt -- --check` clean.

## Scope
This is Part A (the rank-mismatch crash) only. Part B — mixed-length image batches silently corrupting output — is a separate chunk-coordinate issue and is intentionally left out of this PR.

Happy to have @achandi validate on the deterministic repro mentioned in the issue.

### DIFF
--- a/mistralrs-core/src/vision_models/gemma4/text.rs
+++ b/mistralrs-core/src/vision_models/gemma4/text.rs
@@ -9,7 +9,7 @@ use std::{
     },
 };
 
-use candle_core::{DType, Device, Module, Result, Tensor};
+use candle_core::{DType, Device, IndexOp, Module, Result, Tensor};
 use candle_nn::Embedding;
 use mistralrs_quant::{
     softcap, ColumnParallelLayer, QuantMethod, QuantMethodConfig, ReplicatedLayer,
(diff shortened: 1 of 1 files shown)


achandi:
We hit this same crash while investigating #2323 and can offer on-hardware validation: we'd been running an equivalent local patch (same shape — per-row override, `[bs, 1, seq, total]` mask) on real workloads for the past two days.

Setup: DGX Spark (GB10, sm_121, CUDA 13.0), Gemma-4-26B-A4B, Q6K UQFF, 0.8.23 base.

**Crash (Part A): gone.**
- The deterministic repro from #2323 (N identical concurrent image requests, greedy): `unexpected rank, expected: 1, got: 2 ([2, 1080])` no longer reproduces at N=2 or N=5.
- Verified on two builds: cuda-only (which always takes this masked path) and cuda + flash-attn.

**End-to-end effect:** with the equivalent patch, a width-2 batched vision workload (5-page document review) went from near-total output degeneration (every page hitting the token cap with degenerate text) to substantively correct output on our eval — so the fix restores real quality at batch width 2, not just crash-avoidance.

**Still broken after the fix (consistent with this PR's "Part A only" scope):**
- Mixed-length image batches still corrupt stochastically: individual requests collapse to a handful of output tokens (observed out=8-67 where several hundred are expected), with probability rising with batch width; at width 5 the output is unusable. That's Part B of #2323 (chunked-prefill / mixed-length, similar shape to what #2301 fixed for Qwen).
- Same-length batched requests complete post-fix, but identical greedy inputs still diverge token-wise (outputs stay coherent) — looks like cross-sequence reduction-order nondeterminism rather than anything in this mask.

One note for review: downstream mask chunking dispatches on rank, so it's worth confirming the always-4D mask preserves single-request behavior — our local variant conservatively kept the 2-D `[seq, total]` shape for bs == 1; the unit tests added here may already cover this.

Happy to re-run the repro or a width ladder on this exact branch if useful.

gmhelmold:
Thank you for the on-hardware validation — that's exactly the coverage this needed! Great to hear the deterministic repro is gone at both N=2 and N=5 on GB10, and especially that you verified it on the cuda-only build (which always takes the masked path). With your two-build validation on real Gemma-4-26B workloads plus the byte-parity proof for bs=1 in the PR body, I believe Part A is fully covered here; Part B (batched-prefill rank mismatch) remains scoped out as discussed in #2323.

PR #2342, opened 2026-07-12
fix(server): persist stored streaming responses

### LINKED ISSUE #1946: Responses API: streaming OpenResponses with store=true are not persisted for retrieval
`store: true` appears to be honored for non-streaming and background Responses API requests, but not for streaming OpenResponses.

## Summary

In `create_response`:

- Non-streaming path explicitly calls `cache.store_response(...)` when `store` is true.
- Background mode also calls `cache.store_response(...)` when `store` is true.
- Streaming path constructs `OpenResponsesStreamer` and returns `Sse::new(streamer)` without storing a completed `ResponseResource` in cache.

Inside `OpenResponsesStreamer`, completion logic can store conversation history, but it does not store a `ResponseResource` object in response cache.

## Why this is a bug

`GET /v1/responses/{id}` checks:

1. background task manager
2. `cache.get_response(id)`

If streamed responses are not stored in cache, retrieval can return 404 despite `store` defaulting to true.

This means persistence behavior differs by mode:

- non-streaming/background: stored
- streaming: not stored

## Expected behavior

When `store` is true (or default true), completed streaming OpenResponses should be retrievable via `GET /v1/responses/{id}` in the same way as non-streaming/background responses.

glaziermag:
Fixes #1946.

## Summary

Persist terminal streaming OpenResponses when `store=true` (the default), so `GET /v1/responses/{id}` can retrieve them just like non-streaming responses.

This is a one-file, one-commit rebuild of the useful part of closed draft #1985 on current `master`. Current head: `1a6bed7cb5c2361a2b123f143354d87a4a5ef539`.

## Implementation

- Store the final `ResponseResource` immediately before yielding `response.completed`.
- Cover both streamed chunk completion and the fallback `Response::Done` completion path.
- Persist the emitted failed resource for streamed model errors too, matching the existing non-streaming behavior.
- Keep `store=false` responses out of the cache.
- Persist before the terminal event is returned, so correctness does not depend on the client polling once more for the trailing `[DONE]`.
- Reuse the response object's own ID as the cache key.

The earlier draft retained a final response and stored it only after `[DONE]`. This rebuild avoids that extra state and the disconnect window.

## Regression coverage

Focused tests verify that:

- a terminal streaming response with `store=true` is retrievable from the response cache;
- `store=false` leaves no cached response;
- the existing response item lifecycle test remains green.

## Validation

Validated on Linux x86_64 with current stable Rust 1.97.0. The published commit points to the exact tested tree `7205276daf97cadfbcc48ae085b9ac434343c3b8`.

```bash
cargo +stable fmt --all -- --check
cargo +stable check -p mistralrs-server-core
cargo +stable check -p mistralrs-server-core --no-default-features
cargo +stable test -p mistralrs-server-core --no-default-features responses::tests
cargo +stable clippy -p mistralrs-server-core --tests --no-deps --no-default-features -- -D warnings
```

Focused test result: 5 passed, 0 failed (the filter includes the three Responses tests and two response-cache tests).

## Scope

This changes only Responses API persistence. It does not alter streaming event payloads, background response handling, or cache retention policy.


glaziermag:
Runtime validation on a live server (macOS arm64 CPU, gemma-3-1b-it Q4_K_M): `POST /v1/responses` with `stream:true, store:true` (foreground SSE) streamed `response.created`/`response.in_progress` events, and a subsequent `GET /v1/responses/{id}` returned the persisted response — `status: completed` with the message output. This is the exact #1946 scenario: pre-fix, streamed stored responses were not retrievable.

Issue #2343, opened 2026-07-12
Concurrent text-serving throughput does not scale above serial despite PagedAttention engaged (v0.9.0, H100, Llama-3.2-1B bf16)


dharmater:
# Concurrent text-serving throughput does not scale above serial despite PagedAttention engaged (v0.9.0, H100, Llama-3.2-1B bf16)

## Summary

On a plain text model over the OpenAI HTTP surface, concurrent throughput fails
to scale: 32 concurrent requests deliver **341 tok/s total** — below the same
server's **439 tok/s serial** — while the engine log confirms PagedAttention and
the paged scheduler are active. vLLM 0.25.0 on the identical GPU, model, prompts,
and client reaches 10,098 tok/s at the same concurrency (~30×).

## Environment

- mistral.rs **v0.9.0** (tag), built `--features cuda,flash-attn`,
  `CUDA_COMPUTE_CAP=90`, `RUSTFLAGS=-C target-cpu=x86-64-v3`
- NVIDIA **H100** (Modal), CUDA 12.8.1 devel base, Ubuntu 22.04
- Model: `meta-llama/Llama-3.2-1B-Instruct`, bf16 un-quantized (safetensors)
- Server: `mistralrs serve -m meta-llama/Llama-3.2-1B-Instruct --port 1234 --paged-attn on`
- Client: plain `POST /v1/completions`, `temperature 0`, `max_tokens 128`,
  32 distinct prompts, `ThreadPoolExecutor` at concurrency 8 and 32; throughput
  counts actual generated tokens (responses re-tokenized)

## Measurements (identical client/prompts/GPU for all rows)

| server | serial tok/s (p50 lat) | conc=8 | conc=32 |
|---|---|---|---|
| mistral.rs v0.9.0, `--features cuda` (paged-attn auto) | 468–475 | 163 | 227 |
| mistral.rs v0.9.0, `--features cuda,flash-attn`, `--paged-attn on` | 439 (0.29 s) | 257 | **341** |
| mistral.rs v0.9.0, same + `--max-batch-size 32` | 426 (0.30 s) | 244 | 330 |
| vLLM 0.25.0 (defaults, CUDA graphs) | 503 (0.25 s) | 3,309 | 10,098 |
| vLLM 0.25.0 `--enforce-eager` | 131 | 979 | 3,454 |

Concurrency 8 is *slower* than serial in every mistral.rs configuration tested.

## Engine log (the batching stack reports itself active)

```
[...]

Issue #2322, opened 2026-07-08
MemoryUsage::query runs sysinfo System::new_all() in the attention hot path (CPU + integrated CUDA)


achandi:
## Environment
- mistral.rs v0.8.23 (call sites verified unchanged on master `b795620`)
- DGX Spark (GB10, sm_121, aarch64, CUDA 13.0, unified memory -> `is_integrated_gpu` = true)
- Affects: any integrated-memory CUDA box (GB10, Jetson) and CPU inference. Discrete CUDA is unaffected (that branch uses `result::mem_get_info()`).

## Problem
`naive_sdpa` calls `maybe_synchronize`, which calls `MemoryUsage.query(device)` on **every invocation** to check for <4 GiB headroom (attention/backends/naive.rs). On the `Device::Cpu` and integrated-CUDA branches, `query` constructs `sysinfo::System::new_all()`, which enumerates and reads **every process in /proc** (plus CPUs/disks) — even though only `total_memory()` / `available_memory()` are consumed. That is tens of milliseconds of syscall work per attention call, per layer, per step, on the engine thread.

## Captured stack (gdb, engine thread mid-forward)
```
#14 sysinfo::common::system::System::new_all
#15 mistralrs_core::utils::memory_usage::MemoryUsage::query
#16 mistralrs_core::attention::backends::naive::naive_sdpa
#17 mistralrs_core::attention::Sdpa::run_attention_noflash
#18 mistralrs_core::paged_attention::layers::paged_attention::PagedAttention::forward_impl
#19 mistralrs_core::vision_models::gemma4::text::DecoderLayer::forward
```

## Measured impact (DGX Spark, Gemma-4-26B-A4B Q6K UQFF, 2 concurrent vision requests)
| | per vision request | GPU |
|---|---|---|
| before | 312-721 s | ~16 W, mostly idle (engine thread at 100% CPU reading /proc) |
| after (memory-only refresh) | 152-160 s | 42 W / 87% util |

~4x wall-clock on this path from a one-line change.

## Fix
Read only what is used:

```rust
let mut sys = System::new();
sys.refresh_memory();
```

at the three `System::new_all()` call sites in utils/memory_usage.rs (Cpu branch, integrated-CUDA branch, `metal_sysctl_floor_bytes`). PR attached. An alternative/addition is caching the result with a short TTL, but the memory-only refresh (a single /proc/meminfo read) already takes this out of the profile.

Issue #2321, opened 2026-07-08
PagedAttentionScheduler::schedule livelocks with concurrent multimodal requests


achandi:
## Environment
- mistral.rs v0.8.23 (also verified present on master `b795620`)
- Gemma-4-26B-A4B (multimodal), pre-baked Q6K UQFF and BF16 both affected
- DGX Spark (GB10, sm_121, aarch64, CUDA 13.0), Python bindings (in-process), PagedAttention auto-on
- Repro is hardware-independent (pure scheduler logic)

## Symptom
Submitting 2+ concurrent vision chat requests hangs the engine permanently: one engine thread pins a CPU core at 100%, GPU drops to idle, no tokens are ever produced, and nothing is logged (no warning, no panic). Single requests work perfectly. Reproduced identically across BF16/CUTLASS and Q6K/GGUF, with FlashInfer decode on/off, CUDA graphs on/off, and prefix caching on/off.

## Captured stack (gdb, the spinning thread)
```
#0 mistralrs_core::paged_attention::kv_cache_manager::KVCacheManager::allocate_slots
#1 mistralrs_core::paged_attention::scheduler::PagedAttentionScheduler::schedule
#2 mistralrs_core::engine::Engine::run::{{closure}}
```

## Root cause
In the completion-scheduling section of `schedule()` (scheduler.rs), the loop drains `self.running`, and a sequence whose `has_images()` differs from the batch head is pushed **back onto `self.running`** — the same queue the `while !self.running.is_empty()` loop is draining:

```rust
let new_seq_has_images = get_mut_arcmutex!(seq).has_images();
if running.is_empty()
    || get_mut_arcmutex!(running[0]).has_images() == new_seq_has_images
{
    running.push_back(seq);
} else {
    self.running.push_back(seq);
}
```

`Sequence::has_images()` is state-dependent during chunked prefill (sequence.rs): when `prefill_prompt_toks` is set it returns whether the *active multimodal window* overlaps an image, so two vision sequences at different chunk offsets can disagree. Once they do, the loop cannot terminate: the mismatched sequence is popped, `allocate_slots` succeeds (it is a running request needing zero new blocks, so no timeout/preemption path ever fires), the image check fails, and it is pushed back — forever. `schedule()` never returns, the engine never steps, so the chunk windows never advance and the mismatch never resolves.

Perverse detail: allocation *failure* would actually escape (the inner alloc loop preempts), so a healthy block pool spins forever while a starved one survives.

[...]

PR #2325, opened 2026-07-08
perf(core): read only memory stats in MemoryUsage::query

achandi:
## Description

`MemoryUsage::query` is called from `maybe_synchronize` on every `naive_sdpa` invocation. The `Device::Cpu` and integrated-CUDA branches (and `metal_sysctl_floor_bytes`) construct `sysinfo::System::new_all()`, which enumerates and reads every process in /proc plus CPUs and disks, even though only `total_memory()` and `available_memory()` are consumed. On integrated-memory CUDA devices (DGX Spark GB10, Jetson) and CPU inference this lands a full process-table scan inside the per-layer forward.

Replace the three call sites with:

```rust
let mut sys = System::new();
sys.refresh_memory();
```

Measured on a DGX Spark (GB10, CUDA 13.0, Gemma-4-26B-A4B Q6K UQFF, two concurrent vision requests): per-request time 312-721s -> 152-160s (~4x), GPU from ~16W/idle (engine thread at 100% CPU in /proc reads) to 42W/87% util. Discrete CUDA is unaffected (that branch already uses `mem_get_info`).

gdb sample that caught it, engine thread mid-forward:

```
#14 sysinfo::System::new_all
#15 MemoryUsage::query
#16 naive_sdpa
#17 Sdpa::run_attention_noflash
#18 PagedAttention::forward_impl
#19 gemma4::text::DecoderLayer::forward
```

Fixes #2322.

PR #2346, opened 2026-07-14
fix(core): stop decode-scheduler livelock and drop memory query from attention hot path

blasrodri:
Two independent fixes in the paged-attention path.

## Scheduler decode livelock (#2321)

The decode/completion scheduling loop (`while !self.running.is_empty()`) pushed image-incompatible sequences back onto `self.running` -- the very queue it was draining -- with no state change. `has_images()` is state-dependent during chunked prefill, so two vision sequences at different chunk offsets legitimately disagree; the sequence was re-popped, re-evaluated, and re-pushed forever (100% CPU, no progress) on concurrent multimodal requests.

Fix: route image-incompatible sequences to a separate `deferred` queue instead of back into `self.running`. The loop now only pops from `self.running`, so it always terminates.

A batch can only be all-image or all-non-image, so when both types are running only one runs per pass. Two follow-ups keep the deferred type healthy:

- **Fairness (no starvation):** re-appending deferred and relying on FCFS would let the oldest incompatible type win the head every pass and starve the other, since `sort_running_by_priority_fcfs` orders by arrival `timestamp()`. The scheduler now tracks the last batch's image-presence (`last_decode_batch_had_images`) and rotates the FCFS-ordered queue to lead with the opposite type when present, so image and non-image passes alternate.
- **Terminate-all correctness:** the scheduled batch is snapshotted, then deferred sequences are merged back into `self.running` *before* the `RunningCompletion` state update and the `TERMINATE_ALL_NEXT_STEP` block, so cancellation and state updates reach deferred sequences (previously they were merged back only after the flag was cleared and could keep generating).

Deferred sequences keep their allocated KV slots throughout. The `bucket_and_preempt_sequences` behavior (#2343) is untouched.

Adds unit tests for the decode-batch partition (termination when all sequences are image-incompatible, the image-homogeneous batching invariant, edge cases), the fairness rotation, and multi-pass alternation progress.

## Memory query in the attention hot path (#2322)

`MemoryUsage::query` is called per attention op through `maybe_synchronize` on integrated CUDA and Metal. It called `sysinfo::System::new_all()`, which refreshes every process/CPU/disk on the machine, only to read total/available memory.

Fix: replace `System::new_all()` with `System::new()` + `sys.refresh_memory()` at all three sites (CPU, integrated CUDA, Metal `sysctl` helper). The `< 4 GiB` synchronize heuristic and the reported memory values are unchanged; only the incidental full-machine refresh is removed. Discrete CUDA (cheap `mem_get_info`) is unaffected.

### DIFF
--- a/mistralrs-core/src/paged_attention/scheduler.rs
+++ b/mistralrs-core/src/paged_attention/scheduler.rs
@@ -387,6 +387,7 @@ impl PagedAttentionScheduler {
         self.sort_running_by_priority_fcfs();
 
         let mut running: VecDeque<Arc<Mutex<Sequence>>> = VecDeque::new();
+        let mut deferred: VecDeque<Arc<Mutex<Sequence>>> = VecDeque::new();
         while !self.running.is_empty() {
             let seq = self.running.pop_front().unwrap();
             let mut finished_with_break = false;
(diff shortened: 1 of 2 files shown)


blasrodri:
@codex review

blasrodri:
@codex review

blasrodri:
Closing as duplicate. After a closer look at open PRs, everything here is already covered by earlier work:

- The scheduler livelock (#2321) is fixed in #2324, and the paged-attention bucketing/deferral is being reworked in #2034.
- The `MemoryUsage::query` hot-path fix (#2322) is already in #2325.

No sense adding a competing change on top of those. Deferring to the existing PRs. Apologies for the overlap.

PR #2359, opened 2026-07-29
Fix Qwen3 and other GGUF device mismatch when layers are split across CPU and GPU

pknowles:
The final norm runs on the primary device, but device mapping may place trailing layers on CPU, so the hidden state reaching it can be a CPU tensor: rms-norm then fails with `device mismatch, lhs: Cpu, rhs: Cuda`. Move the tensor back before the norm, as quantized_llama.rs already does.

-----

This issue was debugged and fixed by AI. I'm not familiar enough with the repo to know whether it's appropriate. However the change is small and the "as quantized_llama.rs already does" lends an air of correctness. I gather qwen3 support is fairly recent. Take it if it's obviously useful, otherwise I can file an issue with repro instructions.

### DIFF
--- a/mistralrs-core/src/models/quantized_phi3.rs
+++ b/mistralrs-core/src/models/quantized_phi3.rs
@@ -409,7 +409,9 @@ impl ModelWeights {
             let ys = layer.mlp.forward(&ys)?;
             xs = (ys + residual)?
         }
+        // Trailing layers may be device mapped to CPU; the norm weight is not.
         let xs = xs
+            .to_device(&self.device)?
             .apply(&self.output_norm)?
             .i((.., seq_len - 1, ..))?
             .contiguous()?;
--- a/mistralrs-core/src/models/quantized_qwen.rs
+++ b/mistralrs-core/src/models/quantized_qwen.rs
@@ -486,6 +486,8 @@ impl ModelWeights {
             let x = (x + residual)?;
             layer_in = x;
         }
+        // Trailing layers may be device mapped to CPU; the norm weight is not.
+        let layer_in = layer_in.to_device(&self.device)?;
         let x = self.norm.forward(&layer_in)?;
         let x = extract_logits(&x, context_lens)?;
         self.output.forward(&x.contiguous()?)
--- a/mistralrs-core/src/models/quantized_qwen3.rs
+++ b/mistralrs-core/src/models/quantized_qwen3.rs
@@ -455,6 +455,8 @@ impl ModelWeights {
             let x = (x + residual)?;
             layer_in = x;
         }
+        // Trailing layers may be device mapped to CPU; the norm weight is not.
+        let layer_in = layer_in.to_device(&self.device)?;
         let x = self.norm.forward(&layer_in)?;
         let x = extract_logits(&x, context_lens)?;
         self.output.forward(&x.contiguous()?)
--- a/mistralrs-core/src/models/quantized_qwen3_moe.rs
+++ b/mistralrs-core/src/models/quantized_qwen3_moe.rs
@@ -575,6 +575,8 @@ impl ModelWeights {
             let x = (x + residual)?;
             layer_in = x;
         }
+        // Trailing layers may be device mapped to CPU; the norm weight is not.
+        let layer_in = layer_in.to_device(&self.device)?;
         let x = self.norm.forward(&layer_in)?;
         let x = extract_logits(&x, context_lens)?;
         self.output.forward(&x.contiguous()?)
--- a/mistralrs-core/src/models/quantized_starcoder2.rs
+++ b/mistralrs-core/src/models/quantized_starcoder2.rs
@@ -402,7 +402,9 @@ impl ModelWeights {
             let ys = layer.mlp.forward(&ys)?;
             xs = (ys + residual)?
         }
+        // Trailing layers may be device mapped to CPU; the norm weight is not.
         let xs = xs
+            .to_device(&self.device)?
             .apply(&self.output_norm)?
             .i((.., seq_len - 1, ..))?
             .contiguous()?;


pknowles:
*Updated after a local review found the same pattern elsewhere.

> Verdict: correct and general
> The fix is [quantized_qwen3.rs:458-459](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen3.rs#L458-L459):
> 
> 
> // Trailing layers may be device mapped to CPU; the norm weight is not.
> let layer_in = layer_in.to_device(&self.device)?;
> Nothing local is hard-coded. No device index, no CUDA-vs-CPU branch, no layer count, no model-size condition. self.device is the primary device captured from the loader's device argument at [line 391](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen3.rs#L391) — whatever the caller chose, CUDA, Metal, or CPU.
> 
> It targets the right device. norm and output are both built with that same plain device ([lines 262-267](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen3.rs#L262-L267)), deliberately not routed through mapper.device_for(...) the way the per-layer weights are ([line 279](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen3.rs#L279)). So the destination is exactly where the norm weight lives, by construction.
> 
> The bug is real in general, not a quirk of your setup. LayerDeviceMapper::map dispatches on self.mappings[layer] — an arbitrary per-layer device list. Any mapping that puts the last block somewhere other than the primary device hits this, on any backend. CPU-offloading a tail is the ordinary case, not a pathological one.
> 
> It's the established pattern. [quantized_llama.rs:716](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_llama.rs#L716) is the byte-identical line, so this is bringing qwen3 in line with the reference implementation rather than inventing a workaround.
> 
> Zero cost when unmapped. Tensor::to_device short-circuits on same_device to an Ok(self.clone()) (candle tensor.rs:2368-2371) — a refcount bump, no copy. Under DummyDeviceMapper (whose map is the identity) the tensor is already on self.device, so single-device runs are unaffected.
> 
> One thing worth flagging
> Three sibling GGUF models have the same latent bug and were not fixed:
> 
> [quantized_qwen3_moe.rs:578](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen3_moe.rs#L578)
> [quantized_qwen.rs:489](https://github.com/pknowles/mistral.rs/blob/6ee2e05e49172f0ff9126b2b589a31abc109714d/mistralrs-core/src/models/quantized_qwen.rs#L489)
> quantized_phi2.rs / quantized_phi3.rs / quantized_starcoder2.rs — these don't match the self.norm.forward(&layer_in) shape; they'd need a separate look to confirm whether they're structured differently or just spelled differently.
> Not a defect in your commit — the scope you chose is coherent and matches the crash you hit. But if you plan to upstream this, adding the same one-liner to the qwen3_moe and qwen cases would make it a much easier PR to accept, since the reviewer will notice the asymmetry immediately.

...


> 
> What changed
> Extended the guard to four more GGUF models. Two took slightly different forms:
> 
> Same shape as qwen3 — [quantized_qwen3_moe.rs:578](https://github.com/pknowles/mistral.rs/blob/3acb0d93acf28d20915733c90fc4684786865594/mistralrs-core/src/models/quantized_qwen3_moe.rs#L578), [quantized_qwen.rs:489](https://github.com/pknowles/mistral.rs/blob/3acb0d93acf28d20915733c90fc4684786865594/mistralrs-core/src/models/quantized_qwen.rs#L489), byte-identical line.
> 
> apply chain form — [quantized_phi3.rs:412](https://github.com/pknowles/mistral.rs/blob/3acb0d93acf28d20915733c90fc4684786865594/mistralrs-core/src/models/quantized_phi3.rs#L412) and [quantized_starcoder2.rs:404](https://github.com/pknowles/mistral.rs/blob/3acb0d93acf28d20915733c90fc4684786865594/mistralrs-core/src/models/quantized_starcoder2.rs#L404) fold it into the existing chain:
> 
> 
> let xs = xs
>     .to_device(&self.device)?
>     .apply(&self.output_norm)?
> Corrections to what I told you earlier
> Two of my claims from the review were wrong, and both matter:
> 
> phi2 was never affected — it already had the guard at [quantized_phi2.rs:383](https://github.com/pknowles/mistral.rs/blob/3acb0d93acf28d20915733c90fc4684786865594/mistralrs-core/src/models/quantized_phi2.rs#L383). My grep for self.norm.forward missed it because phi2 names the field output_norm and applies it via xs.apply(...). So the reference precedent is two files, not one.
> 
> phi3 and starcoder2 were affected — I'd listed them as "needs a separate look." They had the identical bug, just spelled with .apply() instead of .forward(). Both are now fixed.
> 
> Net: five models needed it, not the three I estimated.
> 
> Verification for the new files
> Each of the four has pub device: Device set to device.clone() from the loader, and loads output_norm on the plain primary device while per-layer weights go through mapper.device_for(...) — same asymmetry that caused the original crash, so self.device is the right destination in every case.
> 
> For phi3 and starcoder2 I also checked self.output, since those two feed the norm result straight into MatMul.qmatmul: both load it via QMatMul::from_qtensor(ct.tensor("output.weight", device)?) on the primary device, so the single move before the norm covers the matmul too — no second to_device needed.
> 
> One caveat on the whole set: this is verified by construction and compilation, not by execution. I haven't run a CPU/GPU-split model through any of the four new paths, so the qwen3 case remains the only one confirmed against a real crash.

pknowles:
Looks like this issue was fixed by https://github.com/EricLBuehler/mistral.rs/pull/2377

Issue #2308, opened 2026-07-05
Proposal: additive KV cache connector seam at paged-attention block pool


Travis-Gilbert:
## Proposal

Add an optional, no-op-by-default KV cache connector seam around the existing paged-attention prefix-cache block flow.

The current code already computes content-addressed block hashes in `mistralrs-core/src/paged_attention/block_hash.rs`, resolves prefix hits through `KVCacheManager::get_computed_blocks`, assigns physical blocks through `allocate_slots`, stores newly full blocks through `cache_blocks`, and evicts cached block hashes when `BlockPool::get_new_blocks` reallocates an LRU free block.

A small connector trait at that seam would let integrations observe or implement external KV tiers without changing default behavior:

- look up a `BlockHash` plus KV cache group IDs before or alongside local prefix-cache lookup
- observe the selected physical block IDs after a hit is allocated
- observe newly cacheable full blocks when `cache_blocks` records them
- observe evicted block hashes when a cached block is reallocated

## Motivation

This is narrower than disaggregated prefill/network transfer. The immediate goal is an additive hook for external KV stores such as RAM/disk/blob registries, while keeping mistral.rs behavior unchanged unless a connector is installed.

The default implementation would return misses and only record no-op observations, so existing tests and users should see no behavior change.

Related prior broader issue: #2000.

## Sketch

```rust
pub trait KvCacheConnector: Send + Sync {
    fn lookup_blocks(&self, _block_hashes: &[BlockHash], _group_ids: &[u32], _max_blocks: usize) -> Option<Vec<usize>> { None }
    fn observe_hit(&self, _block_hashes: &[BlockHash], _block_ids: &[usize], _num_computed_tokens: usize) {}
    fn observe_store(&self, _block_hashes: &[BlockHash], _block_ids: &[usize], _num_full_blocks: usize) {}
    fn observe_evict(&self, _block_hashes: &[BlockHashWithGroupId], _block_id: usize) {}
}
```

I am happy to adjust the exact signatures to match maintainer preference; the key property is that the seam is optional and defaults to current local-only behavior.

PR #2364, opened 2026-08-12
feat(core): add optional no-op KV cache connector seam

marcusacosta:
## Summary
- Adds plumbing for an optional extra KV-cache storage layer on top of today's local prefix cache (#2308).
- Introduces a `KvCacheConnector` hook (default `NoopKvCacheConnector` does nothing) so later backends (memory/disk/S3/etc.) can plug in without changing normal users.
- Calls that hook when the engine looks up, stores, or evicts prefix-cache blocks (`get_computed_blocks`, `cache_blocks`, and block-pool eviction).
- Keeps the default path local-only: same constructors as before (`with_connector` is opt-in); no behavior change unless a real connector is installed.
- This PR is only the connection points. It does not add disk, cloud, or network KV storage yet.

## Notes for reviewers
- If `lookup_blocks` returns block IDs, those IDs must already exist in the local block pool (a future connector would download/hydrate first, then return local IDs).
- Happy to adjust the trait shape to maintainer preference, as noted in #2308.

### DIFF
--- a/mistralrs-core/src/lib.rs
+++ b/mistralrs-core/src/lib.rs
@@ -156,8 +156,9 @@ pub use mistralrs_mcp::{
 pub use mistralrs_quant::{IsqBits, IsqType};
 pub use mistralrs_sandbox::{NetworkMode, SandboxPolicy};
 pub use paged_attention::{
-    BlockHash, BlockHashWithGroupId, KvCacheConnector, MemoryGpuConfig, NoopKvCacheConnector,
-    PagedAttentionConfig, PagedCacheType,
+    compute_block_hashes, BlockHash, BlockHashWithGroupId, InMemoryKvCacheConnector,
+    KVCacheManager, KvCacheConnector, MemoryGpuConfig, NoopKvCacheConnector, PagedAttentionConfig,
+    PagedCacheType,
 };
 pub use pipeline::hf::{
     get_model_file, hf_home_dir, hf_hub_cache_dir, hf_token_path, is_hf_hub_offline,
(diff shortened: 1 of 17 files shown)


EricLBuehler:
Hey @marcusacosta! Thanks for the PR. Perhaps we can add an example of an external KV cache manager if this is the purpose?

marcusacosta:
@EricLBuehler 

Yes, the seam is meant for external KV tiers. I’ll add a small example external connector (in-process / reference manager) that shows store/lookup/evict through KvCacheConnector, without a full cloud backend. Sound good?

marcusacosta:
Added a reference external manager:

- `InMemoryKvCacheConnector` in `mistralrs-core` (process-local hash -> block-id index outside the block pool map)
- runnable example: `cargo run -p mistralrs --example kv_cache_connector`

This shows store/lookup/hit through the seam. It does not offload tensor bytes to disk/S3 yet (that still needs a hydrate path). Happy to adjust the example shape if you want something different.

EricLBuehler:
Hmm, I was thinking something more along the lines of using the manager as one would in a realistic setting, in contrast to demonstrating the manager in isolation. Is this what you had in mind as well?

marcusacosta:
Updated to a realistic path:
- `PagedAttentionMetaBuilder::with_kv_cache_connector(...)` installs the connector into the live paged-attention scheduler
- example now uses `ModelBuilder` + paged attention and runs two chat turns with a shared prefix, printing connector store/lookup/hit stats

`cargo run --release -p mistralrs --example kv_cache_connector`

PR #2365, opened 2026-08-13
feat(core): first-class support for loading GGUFs for all supported models

EricLBuehler:
You can finally use any GGUF* in mistral.rs and immediately take advantage of all the features mistral.rs provides!

- Load GGUF weights through the existing native text and multimodal model pipelines.
- Support automatic config, tokenizer, chat-template, and multimodal projector discovery.
- Preserve all standard mistral.rs features, including tool calling, agents, compatible language-model LoRA, device mapping, and PagedAttention.
- Support GGUF requantization and conversion to UQFF.

* = i-quant support coming very soon.


copumpkin:
Hey @EricLBuehler, I just tried switching my Qwen 3.6 loading code from the old ISQ->AFQ4 loading path to use your new GGUF loader from this PR and while it seems to be working, on my Mac (using Metal and such), this new codepath seems to be doing a lot more work on the CPU, making the whole thing _way_ slower (my test took a couple of minutes before and now has been running for 30 minutes and isn't done yet). Is that expected? Is there additional work that needs to happen for this GGUF loader to be as efficient as the other path? From profiling my test that I just switched to use this commit, it's stuck in `gguf::cpu::cpu_indexed_moe_forward` -> `BlockQ4K::to_float`. It looks like the GGUF k-quant MoE gather has a CPU and a CUDA path (gguf/cpu.rs, gguf/cuda.rs) but no Metal one, whereas AFQ4 has the fused `affine_gather_qmm_rhs` Metal kernel.

PR #2371, opened 2026-08-15
fix(core): align tool chat template JSON rendering

### LINKED ISSUE #2363: Chat template output differs from transformers on tool-calling prompts
Upfront, this is partly a proposal for a crate I maintain. The bug is real and I checked it against
the transformers source, and at the end I suggest my crate for it. You know what a dependency costs
better than I do, so I will just leave it here.

The renderer in `mistralrs-core/src/pipeline/chat_template.rs` already gets most of this right.
`trim_blocks` and `lstrip_blocks` are set, `raise_exception`, `strftime_now` and the pycompat
callback are registered, and `tojson` is replaced with your own. The part that still differs from
transformers is that `tojson`, and it only shows up once tools are in play.

One tool definition, through your filter and through transformers:

```
transformers: {"name": "get_weather", "description": "Compare if temp > 10 & humidity < 50 (user's choice)", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}}}

mistral.rs  : {"description":"Compare if temp \u003e 10 \u0026 humidity \u003c 50 (user\u0027s choice)","name":"get_weather","parameters":{"properties":{"city":{"type":"string"}},"type":"object"}}
```

Three things differ in that one line.

The keys come out alphabetical instead of in insertion order. The workspace pins minijinja without
`preserve_order`, so the order is already gone inside the `Value` before the filter runs. That one
cannot be fixed inside `tojson`, it needs the feature flag, and `serde_json` needs it too.

The separators are `","` and `":"`. Python's `json.dumps` writes `", "` and `": "` by default, and
transformers calls it with the defaults.

`<`, `>`, `&` and `'` come out escaped as `\u003c`, `\u003e`, `\u0026` and `\u0027`. transformers
used to inherit that from Jinja and removed it on purpose. The comment above their override in
`transformers/utils/chat_template_utils.py` reads: "We override the built-in tojson filter because
Jinja's default filter escapes HTML characters".

One more thing outside `tojson`. `strftime_now` uses `chrono::Utc::now()`, but transformers uses
`datetime.now()`, which is local time. Granite 3.x and SmolLM3 print the current date, so they
disagree with transformers for part of every day.

## The proposal

I maintain [hf-chat-template](https://github.com/GregoryBolshakov/hf-chat-template). It renders HF
chat templates and checks its output against python transformers on a corpus of real hub templates,
so this kind of drift gets caught before a release. The same tool above goes through it byte for
byte identical to the transformers line.

It is 1.0, so the API is frozen and no release of it can break you before a 2.0. It also has no
third-party types in its public API, so it cannot force a version of minijinja or anything else
into your tree.

I would be glad to send a PR wiring it into the chat template path. If you would rather keep your
own renderer, no hard feelings, and the four points above still stand on their own.

atirna:
## Summary

- use local time for `strftime_now` and `date_string`
- align current-date rendering with Transformers' local `datetime.now()` behavior

The JSON formatting and ordering changes originally carried by this PR are now present on `master`, so the current diff contains only the remaining clock alignment.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p mistralrs-core chat_template::tests --no-default-features` (20 passed)
- `cargo check -p mistralrs-core --no-default-features`

Fixes #2363


EricLBuehler (approved):
Thank you @atirna!

PR #2380, opened 2026-08-19
perf(gguf): sparse per-expert MoE forward on CPU, no full dequantize

Atituiset:
## Problem

On x86_64, `QTensor::indexed_gemv` returns `None` (the repacked per-expert GEMV path is aarch64-only), so `qtensor_indexed_moe_forward` falls back to **dequantizing the entire stacked expert tensor to f32** and then doing gather+matmul via `UnquantLinear`.

For MoE GGUF models this is catastrophic on CPU: e.g. Qwen3.6-35B-A3B (256 experts) dequantizes ~380MB of Q3_K weights **per MoE layer per token**, resulting in ~1 token/min decode on a modern x86 CPU.

## Fix

Add `sparse_indexed_moe` in `mistralrs-quant/src/gguf/cpu.rs`: slice the stacked `[n_experts, n, k]` QTensor's raw byte storage per expert (GGUF quantized blocks are uniform size), construct a per-expert `QMatMul`, and run quantized matmul **only for the experts actually routed to** (top-k per token). Unused experts are never touched — no dequantization of irrelevant weights.

Falls back to the existing dequantize-and-gather path for layouts it cannot serve (non-CPU, unexpected shapes/dtypes), so behavior is unchanged where the new path doesn't apply.

## Result

~200x decode speedup on x86_64 CPU for Qwen3.6-35B-A3B Q3_K_M (~1 tok/min → ~3.4 tok/s, measured in a multi-device layer-split setup).

## Tests

Two new unit tests in `gguf/cpu.rs`:
- `sparse_indexed_moe_matches_dequantize_gather`: F32 weights — bit-exact match (<1e-5) vs the dequantize+gather reference.
- `sparse_indexed_moe_quantized_close_to_dequantize_gather`: Q8_0 weights — output within 5% of output scale vs reference (covers ggml quantized-activation dot noise; catches routing/indexing errors).

`cargo check`, `cargo test -p mistralrs-quant`, `cargo fmt --check`, `cargo clippy` all clean.

### DIFF
--- a/mistralrs-quant/src/gguf/cpu.rs
+++ b/mistralrs-quant/src/gguf/cpu.rs
@@ -3,13 +3,86 @@
 //! This dequantizes the weights and delegates to UnquantLinear's gather_forward.
 
 use candle_core::{
-    quantized::{QMatMul, QTensor},
-    Result, Tensor,
+    quantized::{QMatMul, QStorage, QTensor},
+    DType, Module, Result, Tensor,
 };
 use candle_nn::Linear;
+use std::borrow::Cow;
 use std::sync::Arc;
 
 use crate::{QuantMethod, QuantMethodConfig, UnquantLinear};
+
+/// Per-expert quantized matmul: route each (token, slot) pair to its expert's slice of
+/// the stacked [num_experts, n, k] QTensor without dequantizing the unused experts.
+/// Returns None for layouts this path cannot serve.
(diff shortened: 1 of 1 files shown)


Atituiset:
Friendly ping — CI is all green. Happy to clarify anything or adjust the approach if you have concerns (e.g. slicing the raw expert bytes vs. a dequantize-once cache). Thanks!

PR #2400, opened 2026-08-25
feat(server): improve Anthropic Messages and Claude Code support

matt-minev:
## Summary

- align Anthropic request translation with current Claude Code traffic, including native `output_config`, mid-conversation system messages, ordered multimodal content, thinking history, and required tool selection
- return cache-aware usage, object-shaped tool input, matched stop sequences, and structurally correct thinking blocks
- emit ordered Anthropic SSE lifecycles for thinking, text, tools, signatures, usage, and stop metadata
- add model-free protocol regressions and update the generated OpenAPI schema, Claude Code configuration, compatibility documentation, and troubleshooting guidance

## Why

Claude Code's Messages traffic exercises more than basic text completion. It sends cached system blocks, adaptive thinking with `display: "omitted"`, native effort configuration, tool definitions, and multi-turn `tool_use`/`tool_result` conversations.

The adapter accepted much of that traffic, but several observable protocol behaviors differed from Anthropic's contract:

- `tool_choice: {"type":"any"}` behaved like automatic selection
- multimodal text and images could be reordered during translation
- thinking streams omitted the required signature lifecycle
- matched custom stop strings were not available to the response adapter
- cached prompt tokens were folded into ordinary input usage
- native output effort and JSON Schema configuration were ignored

This change fixes those behaviors in the protocol adapter rather than adding Claude Code-specific routing.

## Compatibility boundaries

The documentation now calls out the remaining local-backend differences:

- `cache_control` uses mistral.rs automatic prefix caching rather than Anthropic breakpoint and TTL semantics
- local thinking blocks carry an empty structural signature, not an Anthropic cryptographic signature, and `budget_tokens` is not enforced
- rich multimodal tool results and Anthropic context editing are not translated
- server-side web search and code execution retain mistral.rs agentic event shapes

## Rebase follow-up

Master advanced to `a4ad8a244` while this PR was being prepared. That commit's push CI failed Rustfmt and Clippy under Rust 1.98 and left three generated references stale, so this branch also carries the corresponding mechanical cleanup: format OpenAPI imports, remove a needless borrow and clone, scope an intentional `f64`-to-`f32` cast, avoid a test-only precision-loss cast, and refresh the affected CLI and Python reference pages.

## Validation

- `cargo fmt --all -- --check`
- `cargo check --workspace`
[...]

PR #2401, opened 2026-08-25
docs: refresh generated reference artifacts

matt-minev:
## Summary

- preserve the section spacing emitted by the CLI reference generator
- include `encoder_cache_memory_bytes` in the generated Python `Which` reference

## Context

PR #2400 was merged before its generated-artifacts check completed. That check found three stale reference files from the updated upstream base. This follow-up contains only the generator output.

## Validation

- `cd docs && npm run generate`


EricLBuehler (approved):
Thank you @matt-minev

PR #2409, opened 2026-08-30
Resolve oci:// model ids via llmman serve

ericcurtin:
## What

Adds an `oci://` scheme so a model published as a [CNCF ModelPack](https://github.com/modelpack/model-spec) artifact works anywhere a Hugging Face repo id does:

```bash
mistralrs run -m oci://ghcr.io/org/model:tag
mistralrs serve -p 1234 -m oci://ghcr.io/org/model:tag
```

Registries bring the credentials, mirroring and air-gap tooling a deployment already has for container images, which is usually easier to run air-gapped than reaching the Hub.

## Where the hook goes

The obvious place is `ModelSelected`, but `model_id` is a plain `String` on six-plus variants with no shared accessor, so a normalisation pass there would have to match every variant and would rot as variants are added.

Instead this hooks the `model_id` entry points in `pipeline::hf` (`list_repo_files`, `list_repo_files_async`, `get_file`, `try_get_file`, `try_get_file_async`, `read_model_file_range`), **all of which already branch on `model_id.exists()`**. Resolving an `oci://` reference to an extracted directory there means the artifact takes the local-directory path those functions already implement, and every loader (plain, GGUF, vision, diffusion, LoRA/X-LoRA) gets support without knowing about it. The diff to `hf.rs` is 23 added lines, zero deletions.

`files_cached_locally` is deliberately **not** hooked: it answers "is this already local?", so pulling from it would turn a cheap predicate into a multi-gigabyte download.

`try_get_file{,_async}` return `Result<Option<_>, ApiError>` and cannot surface an `anyhow` error, so a resolution failure there falls through to the Hub path, which reports the real problem rather than a misleading one.

Results are memoized for the process's life, since those entry points are called repeatedly for one model and an artifact only needs pulling once.

The pull runs on a thread of its own: a `reqwest` blocking client panics when its runtime is dropped inside an async context, and several of these entry points are async. This is not theoretical, it panicked on the first end-to-end run (see below).

## How acquisition works

Delegated to a running [`llmman serve`](https://github.com/llmmanorg/llmman) rather than reimplemented: llmman already implements the ModelPack media types, registry auth, resumable blob download and a content-addressed store.

- `GET /api/version` probes reachability *and* identity, so a server answering without a `version` field is reported as "not an llmman daemon" rather than as nothing listening.
- `POST /api/pull` streams NDJSON so a multi-gigabyte fetch is not silent. An error arrives **in-band at HTTP 200**, and a stream that ends without `success` is also a failure.
- `llmman resolve --no-pull` reports where the bytes landed; `--no-pull` keeps the daemon the only thing that touches the network.
- `LLMMAN_HOST` is parsed as llmman's own clients do, including rewriting a wildcard bind (`0.0.0.0`, `[::]`) to loopback **by value rather than spelling**, so `[0:0:0:0:0:0:0:0]` is caught too.

A pull needs the daemon reachable and the binary on `PATH` (or `MISTRALRS_LLMMAN_BIN`); each missing piece has its own actionable error, and neither is required unless an `oci://` id is used.

**No new crate:** `reqwest` already has the `blocking` feature enabled in the workspace.

**Explicit scheme, no sniffing.** A bare `registry/name:tag` is indistinguishable from a Hugging Face repo id; guessing would silently hijack existing runs. `maybe_resolve` returns `Ok(None)` immediately for every non-`oci://` path, so the hot path for local dirs and HF ids contacts nothing.

### End-to-end, against a real registry

`Qwen/Qwen3-0.6B` transferred into a local `registry:2` as a genuine ModelPack artifact (`artifactType: application/vnd.cncf.model.manifest.v1+json`, `application/vnd.cncf.model.weight.v1.raw` layers), then removed from llmman's local store so the pull really crossed the network.

```
INFO mistralrs_core::pipeline::oci: Pulling OCI model 'localhost:5555/qwen3-raw:mrs-test' via llmman daemon at http://127.0.0.1:17434
INFO mistralrs_core::pipeline::oci: llmman: pulling manifest
INFO mistralrs_core::pipeline::oci: llmman: pulling (200227345/1519207673 bytes)
...
INFO mistralrs_core::pipeline::oci: llmman: pulling (1345661969/1519207673 bytes)
INFO mistralrs_core::pipeline::normal: Loading model weights.
INFO mistralrs_server_core::mistralrs_for_server_builder: Model loaded.

Paris
```

| Scenario | Result |
| --- | --- |
| `run -m oci://...` (1.5 GB, cold pull from registry) | loads and generates |
| `serve -m oci://...` + `/v1/chat/completions` | serves, `"content": "OK"` |
| Memoization across a whole server run | exactly 1 pull |
| Local dir `-m /path/to/model` (hot path) | unaffected, contacts nothing |
| `LLMMAN_HOST` pointed at a dead port | `no llmman daemon reachable at http://127.0.0.1:19999. Start one with llmman serve, ...` |
| Nonexistent tag (in-band error at HTTP 200) | `llmman pull of '...:doesnotexist' failed: resolve: ...: not found` |

The end-to-end run also caught two things unit tests could not:

1. **The async panic** above (`Cannot drop a runtime in a context where blocking is not allowed`), now fixed by the dedicated thread.
2. An **upstream llmman bug**, unrelated to this PR: `llmman build` writes `.tar` layers but `extract_safetensors_dir` only extracts `.raw` ones, so `llmman resolve` reports a directory it never created. The `!path.exists()` guard in `parse_resolve_output` turned that into a clear error instead of a confusing missing-`config.json` failure later. Artifacts produced by `llmman transfer` use `.raw` and work; I will file this upstream.

### Unit

12 tests in `pipeline::oci`: scheme detection incl. case-insensitivity and the empty-scheme edge case; that an HF repo id, a bare registry ref, a local path and `s3://` are not claimed; `strip_scheme` round-trips; `maybe_resolve` passing non-OCI paths through and rejecting an empty reference; `LLMMAN_HOST` parsing across every form incl. wildcard-to-loopback and expanded IPv6; the resolve contract with leaked-diagnostic and unknown-field tolerance plus six malformed cases; and the NDJSON line protocol (in-band error, `success`, byte progress, non-JSON and blank lines).

- [x] `cargo test -p mistralrs-core --lib pipeline::oci`: 12 passed, 0 failed
- [x] `cargo check --workspace --all-targets` clean
- [x] `cargo clippy --workspace --tests -- -D warnings` clean
- [x] `cargo fmt --all --check` and `cargo doc` clean
- [x] End-to-end against a live `llmman serve` backed by a real registry

### Windows fix in this push

The first CI run failed two of these tests on Windows only: they built the resolve JSON with `format!`, and a Windows temp path (`C:\Users\...`) is full of backslashes, which are invalid JSON escapes. They now build it with `serde_json::json!`, which escapes properly. The macOS Test Suite in that same run was not a separate failure: Windows failed at `01:21:27` and macOS was cancelled at `01:21:32`, i.e. matrix `fail-fast` collateral. All 12 oci tests passed on macOS. CI on the fixed push is awaiting maintainer approval (fork PR).

> Disclosure: written with AI assistance.

### DIFF
--- a/mistralrs-core/src/pipeline/hf.rs
+++ b/mistralrs-core/src/pipeline/hf.rs
@@ -19,6 +19,7 @@ use hf_hub::{
 use tokio::io::{AsyncReadExt, AsyncSeekExt};
 use tracing::{trace, warn};
 
+use crate::pipeline::oci;
 use crate::utils::tokens::get_token;
 
 /// Env variable that, when set to a truthy value, disables all network calls
@@ -323,6 +324,9 @@ pub async fn read_model_file_range(
     token_source: &crate::pipeline::TokenSource,
 ) -> Result<Vec<u8>> {
     let model_path = Path::new(model_id);
+    let resolved = oci::maybe_resolve(model_path)?;
+    let model_path = resolved.as_deref().unwrap_or(model_path);
+
     if model_path.exists() {
         return read_local_file_range(&model_path.join(file), range).await;
     }
@@ -582,6 +586,10 @@ pub(crate) fn list_repo_files(
     should_error: bool,
     revision: &str,
 ) -> Result<Vec<String>> {
+    // Once pulled, an `oci://` artifact is just a local directory.
+    let resolved = oci::maybe_resolve(model_id)?;
+    let model_id = resolved.as_deref().unwrap_or(model_id);
+
     if model_id.exists() {
         let listing = fs::read_dir(model_id).map_err(|err| {
             anyhow!(
@@ -646,6 +654,10 @@ pub(crate) async fn list_repo_files_async(
     should_error: bool,
     revision: &str,
 ) -> Result<Vec<String>> {
+    // Once pulled, an `oci://` artifact is just a local directory.
+    let resolved = oci::maybe_resolve(model_id)?;
+    let model_id = resolved.as_deref().unwrap_or(model_id);
+
     if model_id.exists() {
         let listing = fs::read_dir(model_id).map_err(|err| {
             anyhow!(
@@ -710,6 +722,9 @@ pub(crate) fn get_file(
     file: &str,
     revision: &str,
 ) -> Result<PathBuf> {
+    let resolved = oci::maybe_resolve(model_id)?;
+    let model_id = resolved.as_deref().unwrap_or(model_id);
+
     if model_id.exists() {
         let path = model_id.join(file);
         if !path.exists() {
@@ -741,6 +756,10 @@ pub(crate) fn try_get_file(
     file: &str,
     revision: &str,
 ) -> std::result::Result<Option<PathBuf>, ApiError> {
+    // Cannot surface a resolution error here, so fall through to the Hub path, which reports it.
+    let resolved = oci::maybe_resolve(model_id).ok().flatten();
+    let model_id = resolved.as_deref().unwrap_or(model_id);
+
     if model_id.exists() {
         let path = model_id.join(file);
         if path.exists() {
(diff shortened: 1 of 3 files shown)


EricLBuehler:
Hey @ericcurtin! Thanks for the PR. It seems that some CI is failing.

ericcurtin:
@EricLBuehler happy to contribute! Fixing as we speak. Any other feedback, let us know.

ericcurtin:
Done @EricLBuehler

PR #2439, opened 2026-09-20
fix(cli): don't auto-attach an architecture-incompatible GGUF projector

### LINKED ISSUE #2421: serve: projector auto-discovery binds the directory's only mmproj to ANY model — crashes text-only GGUFs (arch not supported by multimodal loader)
## Environment
- mistral.rs v0.9.3 (prebuilt `mistralrs-cuda130-sm89-x86_64-unknown-linux-gnu`)
- Linux, driver 580.173.02 (CUDA 13.0), RTX 4070 Laptop 8GB

## What happened
Running a **text-only** GGUF whose directory contains one (unrelated) projector file:

```
models/
├── qwen2.5-0.5b-instruct-q4_k_m.gguf   <- text model being served
└── mmproj-F16.gguf                     <- projector of a DIFFERENT (vision) model
```

```
mistralrs serve -f models/qwen2.5-0.5b-instruct-q4_k_m.gguf --host 127.0.0.1 --port N
```

```
INFO mistralrs::commands::serve: GGUF: selected F16 projector `mmproj-F16.gguf`
...
INFO mistralrs_core::pipeline::gguf: Using GGUF base-model assets from `Qwen/Qwen3.5-9B`.
Error: multimodal GGUF architecture `qwen2` is not supported by the native multimodal loader
```

The child exits 1 — every text model in that directory is unserveable.

## Why this matters
Model managers that keep many models (and their projectors) in one flat
directory are a common layout. Discovery (`list_local_gguf_companions`)
scans the whole directory of `-f`, and "auto-selected when unambiguous"
means a single stray `mmproj*.gguf` silently reroutes ANY model into the
multimodal pipeline, which then hard-fails for unsupported arches.

Workaround we use: stage a per-model view directory (symlinks of only the
model's own shard set + its own projector) before spawning — works, but
every integrator has to rediscover this.

## Ask
One (or both) of:
1. An explicit opt-out flag (e.g. `--mmproj none` / `--no-mmproj`) so a
   text-only serve can never be hijacked by directory contents.
2. Arch-aware binding: only auto-attach a projector when the GGUF's
   architecture actually supports multimodal loading (qwen2-vl etc.), and
   skip with a warning otherwise instead of crashing.

Also worth noting: with MULTIPLE projectors in the directory the serve
hard-errors as "ambiguous" — same root cause, also un-serveable.

Happy to test any patch against our integration. Thanks for the great
runtime — the paged-KV + prebuilt CUDA asset matrix is genuinely nice.

ermolushka:
## Problem

Directory-scan mmproj auto-discovery in serve/run only matched on filename ("looks like a projector"), never on whether that projector's architecture actually pairs with the model being served. A non-related mmproj*.gguf belonging to a different (vision) model in the same flat directory would silently get attached to any unrelated text-only GGUF, and the load would then hard-fail deep inside the multimodal loader instead of just serving the text model. With multiple unrelated projectors present, discovery hard-errored as "ambiguous" instead of ignoring them.

Fixes #2421.

- Added `mistralrs_core::gguf_architecture_accepts_projector(architecture, vision_projector_type, audio_projector_type)`, reusing the same acceptance logic (`multimodal_vision_registry::family_from_names`, widened to `pub(crate)`, plus the direct Gemma4/Qwen-VL dispatch architectures) the native multimodal loader itself uses, so the prediction can't drift from what the loader actually accepts.
- Added `filter_projector_candidates` in the CLI's GGUF discovery: reads the local model's own `general.architecture` and each candidate projector's `clip.vision/audio.projector_type` header (metadata-only, no weight data), and drops incompatible candidates before projector resolution runs. Fails open (previous behavior) when the model isn't local or its architecture can't be read.
- Wired the filter into `apply_quant_resolution`, skipped when `ModelType::Multimodal` was explicitly requested (the user already asserted multimodal intent). Logs a `warn!` naming how many candidates were ignored.
- Added `--mmproj none` as an explicit opt-out, via a new mmproj_disabled flag on FormatOptions.

## Testing

- Added `local_gguf_ignores_unrelated_projector_architecture`, `local_gguf_still_attaches_a_genuinely_compatible_projector`, `local_gguf_incompatible_projector_does_not_cause_ambiguity`, and `mmproj_none_disables_auto_selection_even_with_a_compatible_projector` in `mistralrs-cli/src/commands/serve.rs`, using real (metadata-only) GGUF files written via candle's own GGUF writer, not filename stubs.
- Added `projector_compat_tests` (4 tests) for `gguf_architecture_accepts_projector` in `mistralrs-core/src/gguf/mod.rs`.
- Verified the new tests are working by temporarily disabling the filter and confirming both regressed to the issue's exact symptoms (wrong auto-attach; "ambiguous" hard error), then restored.
- Reproduced end-to-end against the real built binary with real (metadata-only) GGUF files on disk, reproducing the issue's exact command and log lines pre-fix, and confirming post-fix it warns and proceeds text-only.
- `cargo test -p mistralrs-core -p mistralrs-cli` (1519 passing)
- `cargo clippy -p mistralrs-core -p mistralrs-cli --tests -- -D warnings` clean
- `cargo fmt --all -- --check` clean
- Regenerated CLI reference docs (`serve.md`, `run.md`, `tune.md`, `bench.md`) for the updated `--mmproj` help text


santanu20:
Tested on the exact repro from #2421: text-only `qwen2.5-0.5b-instruct` GGUF + an unrelated Qwen3.5-9B `mmproj-F16.gguf` (clip) in the same flat directory, built from this branch (Linux, CPU path). Pre-fix this hard-crashes in the multimodal loader; with this PR the projector is skipped with the new warning ("…ignoring them and loading text-only. Pass `--mmproj <file>`…"), the model loads through the native qwen2 text pipeline, the server comes up, and chat completions work. Fix confirmed — thanks!

